PR #10, opened 2023-10-10
feat: mainnet multichain

tynes:
**Description**

Tenderly simultation: https://dashboard.tenderly.co/public/safe/safe-apps/simulator/35dfb460-493c-471a-89fc-1a244c9d8fe2


tynes:
I will follow up with the readme and justfile

PR #11, opened 2023-10-11
path + clarification on derivation

juan-fintech:
corrected a path + added clarification (feedback from the user to add reminder)

**Description**

A clear and concise description of the features you're adding in this pull request.

**Tests**

Please describe any tests you've added. If you've added no tests, or left important behavior untested, please explain why not.

**Additional context**

Add any other context about the problem you're solving.

**Metadata**

- Fixes #[Link to Issue]


juan-fintech:
@zchn , changed the path (was broken, most likely old name) + added a clarification about the derivation path on just.

cc: @maurelian

PR #13, opened 2023-10-11
Merge install-dev-contracts to install

zchn:
since they are needed by security council as well, not just the developers.


maurelian:
Let's see if this merges automatically

PR #34, opened 2024-01-12
Diff checking 

maurelian:
Just a draft for now

### DIFF
--- a/script/SignAndCheck.s.sol
+++ b/script/SignAndCheck.s.sol
@@ -1,0 +1,117 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.15;
+
+import {SignFromJson} from "script/SignFromJson.s.sol";
+import {JsonTxBuilderBase} from "src/JsonTxBuilderBase.sol";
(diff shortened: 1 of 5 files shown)


script/SignAndCheck.s.sol
```diff
@@ -0,0 +1,117 @@
...
+    mapping(address => bool) public loggedAccountAccesses;
+    mapping(address => bool) public loggedAccountDiffs;
+    mapping(address => bool) public loggedAccountHasDiffs;
+
+    function setUp() public {
+        string memory path = string.concat(
+            vm.projectRoot(), "/lib/optimism/packages/contracts-bedrock/deploy-config/", "sepolia", ".json"
+        );
+        cfg = new DeployConfig(path);
+        vm.label(0xd81f43eDBCAcb4c29a9bA38a13Ee5d79278270cC, "StorageSetter");
```
pegahcarter, line 28: Suggestion: these contracts will be used for tests/scripts across repos.  We can move them into a standalone repo with a forge Test / Script contract which inherits these constants.  It will help keep track of active implementations as well.

maurelian: Good suggestion, will take this into future consideration as this repo evolves and becomes more stable.

PR #50, opened 2024-02-01
Update deps

maurelian:
**Description**

A small PR which updates submodules and remapping, to prepare to implement diff checking.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -213,7 +213,7 @@ jobs:
             just simulate-council # simulate again to make sure the json is still valid
   just_simulate_sc_rehearsal_2_0122:
     docker:
-      - image: us-docker.pkg.dev/oplabs-tools-artifacts/images/ci-builder:latest
+      - image: us-docker.pkg.dev/oplabs-tools-artifacts/images/ci-builder:v0.35.0
     steps:
       - checkout
       - run:
--- a/.gitmodules
+++ b/.gitmodules
@@ -23,3 +23,6 @@
 [submodule "lib/eip712sign"]
 	path = lib/eip712sign
 	url = https://github.com/ethereum-optimism/eip712sign.git
+[submodule "lib/solady"]
+	path = lib/solady
+	url = https://github.com/Vectorized/solady
--- a/foundry.toml
+++ b/foundry.toml
@@ -15,5 +15,5 @@ remappings = [
   '@base-contracts/=lib/base-contracts/',
   '@eth-optimism-bedrock/=lib/optimism/packages/contracts-bedrock/',
   '@rari-capital/solmate/=lib/solmate',
-  'solady/utils/=lib/optimism/packages/contracts-bedrock/lib/solady/src/utils/'
+  '@forge-std/=lib/forge-std',
 ]
--- a/script/DeployRehearsalContracts.s.sol
+++ b/script/DeployRehearsalContracts.s.sol
@@ -5,14 +5,14 @@ import {Deployer} from "@eth-optimism-bedrock/scripts/Deployer.sol";
 import {GnosisSafe} from "safe-contracts/GnosisSafe.sol";
 import {ProxyAdmin} from "@eth-optimism-bedrock/src/universal/ProxyAdmin.sol";
 import {Proxy} from "@eth-optimism-bedrock/src/universal/Proxy.sol";
+import {EIP1967Helper} from "@eth-optimism-bedrock/test/mocks/EIP1967Helper.sol";
 import {Script} from "forge-std/Script.sol";
 
-import { console2 as console } from "forge-std/console2.sol";
+import {console2 as console} from "forge-std/console2.sol";
 
 // forge script scripts/Deploy.s.sol:Deploy --private-key $PRIVATE_KEY --broadcast --rpc-url $ETH_RPC_URL
 // forge script scripts/Deploy.s.sol:Deploy --sig 'sync()' --private-key $PRIVATE_KEY --broadcast --rpc-url $ETH_RPC_URL
 contract DeployRehearsalContracts is Deployer {
-
     GnosisSafe owner_safe;
     GnosisSafe council_safe;
     ProxyAdmin proxy_admin;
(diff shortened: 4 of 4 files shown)


mds1:
#51 is merged but I think we'll need to rebase this one now for it to take effect here?

.gitmodules
```diff
@@ -17,3 +17,6 @@
 [submodule "lib/base-contracts"]
 	path = lib/base-contracts
 	url = https://github.com/base-org/contracts
+[submodule "lib/solady"]
+	path = lib/solady
+	url = https://github.com/Vectorized/solady
```
maurelian, line 22: I just checked, we actually don't need this.

I might want to add it in a subsequent PR, but will remove for now.

mds1:
Ah nice, just needed to refresh 😅

.gitmodules
maurelian: did that

lib/base-contracts
mds1, line 1: Would be nice to get all deps on tagged commits if we can:
- solmate has tags but there are not often new ones, so it's ok if we need a non-tagged one
- safe-contracts hasn't had a tag in a while, not sure we should use the most recent or an untagged one
- base-contracts has no upstream tags
```
6213d69c88ab92042e061ed62ac2f39b623b25db lib/base-contracts (remotes/origin/HEAD)
 ae570fec082bfe1c1f45b0acca4a2b4f84d345ce lib/forge-std (v1.7.6)
 fd81a96f01cc42ef1c9a5399364968d0e07e9e90 lib/openzeppelin-contracts (v4.9.3)
 c73850809be1bef888ba7dd1194acdf222e4d819 lib/optimism (v1.2.0)
 186a21a74b327f17fc41217a927dea7064f74604 lib/safe-contracts (v1.1.0-184-g186a21a)
 a6a95729f947bb2a24e05e862ba9522c10453a70 lib/solady (v0.0.166)
 e8f96f25d48fe702117ce76c79228ca4f20206cb lib/solmate (v6-188-ge8f96f2)
```

foundry.toml
mds1, line 1: Getting some compiler warnings from the monorepo when building here. I think we'd have to update the forge-std version in the monorepo to fix those warnings, then bump the optimism version we use here afterwards
```
[⠒] Solc 0.8.15 finished in 27.07sCompiler run successful with warnings:
Warning (2018): Function state mutability can be restricted to pure
   --> lib/optimism/packages/contracts-bedrock/scripts/Deployer.sol:282:5:
    |
282 |     function _getContractNameFromDeployTransaction(string memory _deployTx) internal returns (string memory) {
    |     ^ (Relevant source part starts here and spans across multiple lines).

Warning (2018): Function state mutability can be restricted to view
   --> lib/optimism/packages/contracts-bedrock/scripts/Deployer.sol:453:5:
    |
453 |     function _getDeploymentContext() private returns (string memory) {
    |     ^ (Relevant source part starts here and spans across multiple lines).

[⠢] Solc 0.8.15 finished in 27.07s
```

lib/forge-std
mds1, line 1: Unfortunately it seems CI is failing due to remappings errors
```
[⠊] Compiling...
Error: 
failed to resolve file: "/root/project/lib/base-contracts/lib/solady/src/utils/LibSort.sol": No such file or directory (os error 2); check configured remappings.
    --> "/root/project/lib/base-contracts/script/universal/MultisigBase.sol"
        "solady/utils/LibSort.sol"
[⠊] Compiling...
Error: 
failed to resolve file: "/root/project/lib/base-contracts/lib/solady/src/utils/LibSort.sol": No such file or directory (os error 2); check configured remappings.
    --> "/root/project/lib/base-contracts/script/universal/MultisigBase.sol"
        "solady/utils/LibSort.sol"

Exited with code exit status 1
```

.gitmodules
maurelian: OK, now it seems like we do need it, the updated base dep depends on it.

.circleci/config.yml
```diff
@@ -213,7 +213,7 @@ jobs:
             just simulate-council # simulate again to make sure the json is still valid
   just_simulate_sc_rehearsal_2_0122:
     docker:
-      - image: us-docker.pkg.dev/oplabs-tools-artifacts/images/ci-builder:latest
+      - image: us-docker.pkg.dev/oplabs-tools-artifacts/images/ci-builder:v0.35.0
```
maurelian, line 216: I was getting a command not found error when running `just` on the step below. 
This change just ensures we're using the same docker image tag as the above jobs, and it fixes the ci failure.

lib/forge-std
maurelian: yeah, this was due to me thinking I could remove solady. I put it back in.

foundry.toml
maurelian: bumping the monorepo version fixes this.

lib/base-contracts
maurelian: `git submodule status` is not really printing the ideal labels for each of these commits. 

Here's an update status output, with annotations where the output doesn't have a nice tag:

```
 c2a45d575fbf2a7c41370aaa1e9d47a3f6020576 lib/base-contracts (heads/main) 
# no releases or tags
 74464751854981cc815ac2cfd39a3e39cb4b5654 lib/eip712sign (heads/main) 
# no releases or tags
 ae570fec082bfe1c1f45b0acca4a2b4f84d345ce lib/forge-std (v1.7.6) ✅
 fd81a96f01cc42ef1c9a5399364968d0e07e9e90 lib/openzeppelin-contracts (v4.9.3) ✅
 6de6b5fc81d8ee03bb776219ba25189a04712f99 lib/optimism (op-batcher/v1.5.0) ✅ 
# this is also the commit for `v1.5.0` but that's not shown by the output. 
 04ca2bef190a56a0f79532681323b685fbc0f2f4 lib/presigner (04ca2be)
# no releases or tags
 892448e93f6203b530630f20de45d8a55fde7463 lib/safe-contracts (v1.1.0-41-g892448e)
# This is also the commit for the v1.1.1 release (https://github.com/safe-global/safe-smart-account/releases/tag/v1.1.1)
# Which is what we're using on mainnet. 
 a6a95729f947bb2a24e05e862ba9522c10453a70 lib/solady (v0.0.166) ✅
 e8f96f25d48fe702117ce76c79228ca4f20206cb lib/solmate (v6-188-ge8f96f2)
# This is the newest commit of the v7 branch. https://github.com/transmissions11/solmate/tree/v7
 ```

mds1: > `git submodule status` is not really printing the ideal labels for each of these commits.

Yea I've noticed this in the past. IIRC it has to do with default output of `git describe` (which is used under the hood for `git submodule status`) for lightweight vs. annotated tags. And sometimes repos mix both tags which messes with the behavior, so it's not always what you'd expect. 

This was really helpful though, thank you!

.circleci/config.yml
mds1: Sgtm, but when should we bump to a more recent ci-builder tag, since v0.35.0 is no longer the latest

maurelian: Latest breaks the build, and shouldn't imo block this pr

mds1 (approved):

PR #65, opened 2024-02-08
Pin foundry to the same version as in the monorepo

maurelian:
**Description**

Uses a technique we were using in the monorepo to pin the foundry version to a specific commit.
The commit used here currently [matches the monorepo](https://github.com/ethereum-optimism/optimism/blob/develop/versions.json#L3), though that can drift overtime.

An alternative if we wish to keep them in sync is to use the monorepo install script directly.


maurelian:
> @maurelian do you still want to merge this PR? If so, can you make the CI pass?

No, I misunderstood your issue, and don't think we need this now.

PR #66, opened 2024-02-09
Add state diff tooling

maurelian:
**Description**

Adds `LibStateDiffChecker`. This library establishes a standard format for defining the expected state diff of a transaction in a JSON file. 

The format is a simplified version of the state accesses recorded by foundry. 

The library provides utility functions for:
1. parsing an expected diff spec file.
2. extracting an actual diff in the same format from a list of StateAccesses
3. comparing the expected and actual diffs


test/StateDiffChecker.t.sol
```diff
@@ -0,0 +1,190 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.15;
+
+import {Test} from "forge-std/Test.sol";
+import {VmSafe} from "forge-std/Vm.sol";
+import {stdJson} from "forge-std/StdJson.sol";
+import {console2 as console} from "forge-std/console2.sol";
+import {LibStateDiffChecker as Checker} from "script/LibStateDiffChecker.sol";
+import {LibStateDiff} from "@eth-optimism-bedrock/scripts/libraries/LibStateDiff.sol";
```
mds1, line 9: Two unused importss
```suggestion
import {LibStateDiffChecker as Checker} from "script/LibStateDiffChecker.sol";
```

test/StateDiffChecker.t.sol
```diff
@@ -0,0 +1,190 @@
...
+        assertEq(diffSpec.storageSpecs[0].slot, bytes32(0));
+        assertEq(diffSpec.storageSpecs[0].newValue, bytes32(0));
+        assertEq(diffSpec.storageSpecs[1].slot, bytes32(uint256(1)));
+        assertEq(diffSpec.storageSpecs[1].newValue, bytes32(uint256(uint160(address(0xabba)))));
+        assertEq(diffSpec.storageSpecs[2].slot, keccak256(abi.encodePacked(uint256(1), uint256(2))));
+        assertEq(diffSpec.storageSpecs[2].newValue, hex"acdc");
+    }
+
+    /// @dev Utility function which sets up the subsequent tests by executing Target.set(), parsing the diff spec
+    ///      from testDiff.json, and returning the resutling state diff structs.
```
mds1, line 55:
```suggestion
    ///      from testDiff.json, and returning the resulting state diff structs.
```

script/LibStateDiffChecker.sol
```diff
@@ -0,0 +1,154 @@
...
+    /// @param _jsonData The JSON string containing the account access specifications.
+    /// @return stateDiffSpec_ An array of StorageDiffSpec structs parsed from the JSON data.
+    function parseDiffSpecs(string memory _jsonData) internal pure returns (StateDiffSpec memory stateDiffSpec_) {
+        stateDiffSpec_ = abi.decode(vm.parseJson(_jsonData), (StateDiffSpec));
+    }
+
+    /// @notice Filters account accesses to return only those that modified storage.
+    /// @param accountAccesses An array of account accesses recorded by the EVM.
+    /// @return stateDiffSpec_ An array of account accesses that modified storage.
+    function extractDiffSpecFromAcccountAccesses(VmSafe.AccountAccess[] memory accountAccesses)
```
mds1, line 37:
```suggestion
    function extractDiffSpecFromAccountAccesses(VmSafe.AccountAccess[] memory accountAccesses)
```

test/StateDiffChecker.t.sol
```diff
@@ -0,0 +1,190 @@
...
+pragma solidity ^0.8.15;
+
+import {Test} from "forge-std/Test.sol";
+import {VmSafe} from "forge-std/Vm.sol";
+import {stdJson} from "forge-std/StdJson.sol";
+import {console2 as console} from "forge-std/console2.sol";
+import {LibStateDiffChecker as Checker} from "script/LibStateDiffChecker.sol";
+import {LibStateDiff} from "@eth-optimism-bedrock/scripts/libraries/LibStateDiff.sol";
+
+/// @dev A simple contract to which makes some state modifications for the purpose of testing state diff checking.
```
mds1, line 11: Do you think this is too simple? For example it doesn't verify we handle calls or proxies/delegatecalls correctly. Probably would be good to test a double delegatecall also

test/StateDiffChecker.t.sol
```diff
@@ -0,0 +1,190 @@
...
+        Checker.StateDiffSpec memory copy;
+        copy.chainId = diff.chainId;
+        copy.storageSpecs = new Checker.StorageDiffSpec[](diff.storageSpecs.length);
+        for (uint256 i = 0; i < diff.storageSpecs.length; i++) {
+            copy.storageSpecs[i].account = diff.storageSpecs[i].account;
+            copy.storageSpecs[i].slot = diff.storageSpecs[i].slot;
+            copy.storageSpecs[i].newValue = diff.storageSpecs[i].newValue;
+            copy.storageSpecs[i].previousValue = diff.storageSpecs[i].previousValue;
+        }
+        return copy;
```
mds1, line 116: A downside here is that if we add fields to the state diff we have to manually remember to update this function. Any ideas on a good alternative here so this is easier to maintain? I think copying with assembly like below would copy by reference
```solidity
Checker.StateDiffSpec memory copy;
assembly {
  copy := diff
}
return copy;
```

The full array should be sequential in memory so maybe we can extract the length out and loop over those memory chunks (or use the identity precompile) to copy that to a new memory location?

script/LibStateDiffChecker.sol
```diff
@@ -0,0 +1,154 @@
...
+
+import {VmSafe} from "forge-std/Vm.sol";
+import {stdJson} from "forge-std/StdJson.sol";
+import {console2 as console} from "forge-std/console2.sol";
+
+library LibStateDiffChecker {
+    VmSafe private constant vm = VmSafe(address(uint160(uint256(keccak256("hevm cheat code")))));
+
+    // The order of the fields in the struct is important, and must be alphabetical for some reason related to
+    // how foundry encodes the struct to and from JSON.
```
mds1, line 12:
```suggestion
    // The order of the fields in the struct is important, and must be alphabetical
    // for foundry to properly decode data: https://book.getfoundry.sh/cheatcodes/parse-json#decoding-json-objects-into-solidity-structs
```

script/LibStateDiffChecker.sol
```diff
@@ -0,0 +1,154 @@
...
+        address account;
+        bytes32 newValue;
+        bytes32 previousValue;
+        bytes32 slot;
+    }
+
+    struct StateDiffSpec {
+        uint256 chainId;
+        StorageDiffSpec[] storageSpecs;
+    }
```
mds1, line 23: Some things we neglect in our state diff: nonce increments and accounts with new code (both can happen if a contract deploys a contract)

mds1 (commented):
I haven't finished reviewing everything, but here are my initial comments

test/StateDiffChecker.t.sol
mds1: Idea from @maurelian that we'll go with: just verify that the hash of `copy` matches the hash of `diff` and revert if not

test/StateDiffChecker.t.sol
maurelian: Sure, I will add some more complexity to the diff.

script/LibStateDiffChecker.sol
maurelian: We agreed that this is OK for now.

test/StateDiffChecker.t.sol
maurelian: OK, I added new `Callee` and `DelegateCallee` contracts which result in two new diffs.

test/StateDiffChecker.t.sol
maurelian: I implemented this.

script/LibStateDiffChecker.sol
```diff
@@ -0,0 +1,154 @@
...
+                uint256 storageIndex = 0;
+                storageIndex < accountAccesses[acctIndex].storageAccesses.length;
+                storageIndex++
+            ) {
+                if (
+                    accountAccesses[acctIndex].storageAccesses[storageIndex].previousValue
+                        != accountAccesses[acctIndex].storageAccesses[storageIndex].newValue
+                ) {
+                    modifiedCount++;
+                }
```
mds1, line 59: This will over-allocate length, since we might write different values to a slot multiple times? But that should be ok since later we only save the latest value

test/StateDiffChecker.t.sol
```diff
@@ -0,0 +1,224 @@
...
+            copy.storageSpecs[i].account = diff.storageSpecs[i].account;
+            copy.storageSpecs[i].slot = diff.storageSpecs[i].slot;
+            copy.storageSpecs[i].newValue = diff.storageSpecs[i].newValue;
+            copy.storageSpecs[i].previousValue = diff.storageSpecs[i].previousValue;
+
+            // Ensure that all values have been copied over
+            require(
+                keccak256(abi.encode(copy.storageSpecs[i])) == keccak256(abi.encode(diff.storageSpecs[i])),
+                "StateDiffCheckerTest: StorageDiffSpec copying failed"
+            );
```
mds1, line 148: Probably better to use `keccak256(abi.encode(copy)) == keccak256(abi.encode(diff))` outside of the loop to handle if any top-level fields were added

script/LibStateDiffChecker.sol
```diff
@@ -0,0 +1,154 @@
...
+                    });
+                    specIndex++;
+                }
+            }
+        }
+
+        require(specIndex == modifiedCount, "LibStateDiffChecker: found fewer storage modifications than expected.");
+    }
+
+    /// @notice Checks if a single expected account access matches the actual account access.
```
mds1, line 92: Outdated natspec?
```suggestion
    /// @notice Checks if two state diffs are equivalent
```

script/LibStateDiffChecker.sol
```diff
@@ -0,0 +1,154 @@
...
+    /// @param expectedDiff The expected account access details.
+    /// @param actualDiff   The actual account access details recorded by the EVM.
+    function checkStateDiff(StateDiffSpec memory expectedDiff, StateDiffSpec memory actualDiff) internal pure {
```
mds1, line 95: Should we add a catch-call `keccak256(abi.encode(expectedDiff)) == keccak256(abi.encode(actualDiff))` check at the end for any fields we forgot to add here?

test/StateDiffChecker.t.sol
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    /// @dev Test that the sample testDiff.json file is read and parsed correctly
+    function test_parseDiffSpecs() public {
+        string memory _path = "test/testDiff.json";
+        string memory json = vm.readFile(_path);
+        Checker.StateDiffSpec memory diffSpec = Checker.parseDiffSpecs(json);
+
+        assertEq(diffSpec.chainId, 31337);
+        assertEq(diffSpec.storageSpecs.length, 3);
```
mds1, line 73: This should be 5 now it seems—we should run these tests in CI

script/LibStateDiffChecker.sol
maurelian: They are not filtered out, but your test didn't change anything because `x` was declared as non-zero: 

https://github.com/ethereum-optimism/superchain-ops/blob/60f6c611d7d499f506ab46ceac4e320bced102bc/test/StateDiffChecker.t.sol#L11

Filtering in order to show only final state changes will require more work, and likely a change to how we do this tracking (ie. possibly using a mapping with `vm.dumpState`). 

But for now it is not supported.

test/StateDiffChecker.t.sol
maurelian: Good catch. I added these to CI, and refactored the parsing test to check that the hash of the full diff matches.

script/LibStateDiffChecker.sol
mds1: > your test didn't change anything because `x` was declared as non-zero:

Sorry what do you mean by this? That link gives me a 404, but L11 there on the latest commit is the `uint256 public x;` declaration

maurelian: I had done this in the state var declaration: 

```
contract Target {
   uint256 public x = 1;
...
```

script/LibStateDiffChecker.sol
maurelian: Added this.

test/StateDiffChecker.t.sol
mds1: LGTM. Any thoughts on adding the `deny_warnings = true` to a CI profile also? We now have a solc warning in this method

maurelian: Fixed and added the ci profile: [`0d07195` (#66)](https://github.com/ethereum-optimism/superchain-ops/pull/66/commits/0d071951e0445a14083d5e48a222720ad32fc628)

PR #78, opened 2024-02-16
Add mainnet protocol versions task

trianglesphere:
**Description**

A clear and concise description of the features you're adding in this pull request.

**Tests**

Please describe any tests you've added. If you've added no tests, or left important behavior untested, please explain why not.

**Additional context**

Add any other context about the problem you're solving.

**Metadata**

- Fixes #[Link to Issue]


tasks/eth/4-protocol-versions/README.md
```diff
@@ -0,0 +1,204 @@
...
+
+Transactions were created using
+```
+export PV_ENC=$(cd ../../../lib/optimism && go run ./op-chain-ops/cmd/protocol-version encode --major 5)
+# 0x0000000000000000000000000000000000000005000000000000000000000000 
+export PV_ADDR=0x8062AbC286f5e7D9428a0Ccb9AbD71e50d93b935 
+just add-transaction $PV_ADDR 'setRecommended(uint256)' $PV_ENC
+just add-transaction $PV_ADDR 'setRequired(uint256)' $PV_ENC
+```
+This batches setting both, the recommended and required versions in one multicall.
```
maurelian, line 22: Signers don't need to see this, we can remove it.

tasks/eth/4-protocol-versions/README.md
```diff
@@ -0,0 +1,204 @@
...
+
+## Approving the transaction
+
+### 1. Update repo and move to the appropriate folder for this rehearsal task:
+
+```
+cd superchain-ops
+git pull
+just install
+cd tasks/eth/4-protocol-version
```
maurelian, line 32:
```suggestion
cd tasks/eth/3-protocol-version
```

tasks/eth/4-protocol-versions/README.md
```diff
@@ -0,0 +1,204 @@
...
+    * before: `0x0000000000000000000000000000000000000003000000010000000000000000`
+    * after : `0x0000000000000000000000000000000000000005000000000000000000000000`
+* All other state changes (2) are a nonce change of the sender account and the multisig.
+
+On the "Events" tab, you can verify that two `ConfigUpdate` events were emitted from the `ProtocolVersions` proxy,
+as well as an `ExecutionSuccess` event by the multisig.
+
+![](./images/tenderly-state.png)
+
+You can verify the correctness of the storage slots with `chisel`.
```
maurelian, line 98: I cannot reliably run `chisel`. It fails with this error when I run it in this repo:
```
solc 0.8.15 is not supported by the set evm version: paris. Please install and use a version of solc higher or equal to 0.8.18.
You can also set the solc version in your foundry.toml.
```

It's also not a tool that we've asked signers to use before, so I anticipate confusion. 

Could we instead suggest using:

```
cast keccak "protocolversion.required"
```

And indicate that the slot changed should be one less then what is returned by that?

maurelian (changes requested):
See inline comments

tasks/eth/4-protocol-versions/README.md
sebastianst: Does your `chisel` installation run normally in other repos? I just use the latest installation from `foundryup`, which is also recommended in the README of this repo. 

I'm wondering if asking signers to subtract `1` from a `cast` call might be more confusing than asking them to copy `bytes32(uint256(keccak256("protocolversion.required")) - 1)` into `chisel`.

But if you still think it's better with `cast keccak`, we will adapt this section.

sebastianst: I can also completely remove this section on storage slot verification, if that's not required anyways. I found it a nice additional sanity check.

tasks/eth/4-protocol-versions/README.md
sebastianst: Moved it to `03-protocol-versions-delta`. In anticipation of there being more than 10 tasks, it probably makes sense to move all tasks into two-digit prefixed folders, which gives a better lexicographical ordering.

tasks/eth/4-protocol-versions/README.md
sebastianst: Sure, just added it for the purpose of documentation.

tasks/eth/4-protocol-versions/.env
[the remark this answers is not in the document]
sebastianst: Moved it to `03-protocol-versions-delta`. In anticipation of there being more than 10 tasks, it probably makes sense to move all tasks into two-digit prefixed folders, which gives a better lexicographical ordering.

tasks/eth/4-protocol-versions/README.md
maurelian: I see that you've included both which LGTM

tasks/eth/03-protocol-versions-delta/README.md
```diff
@@ -0,0 +1,202 @@
+# Mainnet Protocol Versions Update - 5.0.0 (Delta)
+
```
maurelian, line 2: This is required until it receives foundation approval.

```suggestion

Status: DRAFT, NOT READY TO SIGN

```

maurelian (approved):

PR #79, opened 2024-02-16
Update Ceremony 4 for nested multisig

maurelian:
**Description**

Updates the Upgrade #4 Ceremony to support signing via the 2 of 2 Safe shared by the Security Council and Optimism Foundation.

The primary changes are: 

1. The `justfile` now has duplicate commands with either the `-council` or `-foundation` suffix.
3. The README instructions document the new commands
2. The Validations now have additional sections to reflect different state changes which occur depending which group is signing.


tasks/eth/3-add-superchainConfig/Validation.md
```diff
@@ -60,6 +77,53 @@ There should also be a single 'State Override' in the Foundation Safe contract
   **Meaning:** Sets the `SuperchainConfigProxy` address at slot `0xfb` (251). The correctness of
    this slot is attested to in the Optimism repo at [storageLayout/L1CrossDomainMessenger.json](https://github.com/ethereum-optimism/optimism/blob/op-contracts/v1.2.0-rc.1/packages/contracts-bedrock/snapshots/storageLayout/L1CrossDomainMessenger.json#L122-L127).
 
+### `0x5a0aae59d09fccbddb6c6cceb07b7279367c3d2a` (The 2 of 2 `ProxyAdmin` owner Safe)
```
maurelian, line 80: I do not love the `cast index` gymnastics required below, but it seemed like the most straightforward way to validate this state change without introducing tools which haven't been used in the ceremony previously.

tasks/eth/3-add-superchainConfig/Validation.md
```diff
@@ -82,6 +146,12 @@ There should also be a single 'State Override' in the Foundation Safe contract
   **After:** `0x00000000000000000000000074e273220fa1cb62fd756fe6cbda8bbb89404ded` <br/>
   **Meaning:** Implementation address is set to the new [`OptimismMintableERC20Factory`](https://github.com/ethereum-optimism/optimism/blob/op-contracts/v1.2.0-rc.1/op-chain-ops/cmd/op-upgrade-extended-pause/main.go#L51-L53).
 
+### Foundation Only: `0x847b5c174615b1b7fdf770882256e2d3e95b9d92` (Foundation Safe)
+
+- **Key:** `0x0000000000000000000000000000000000000000000000000000000000000005` <br/>
+  **After:** `0x0000000000000000000000000000000000000000000000000000000000000001` <br/>
+  **Meaning:** The nonce is increased by one.
+
```
maurelian, line 154: Tenderly shows state changes in ascending order by address. This new block, and the 'Council only' block below are listed in the order that they would appear in the simulation.

script/NestedSignFromJson.s.sol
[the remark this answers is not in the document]
mds1: Based on the comments:

1. `signJson` just generates a signature and broadcasts no transactions
2. `approveJson` sends a transaction to call `approveHash`
3. `runJson` executes the transaction that was approved in the prior step

Is that all correct, e.g. 2 and 3 send transactions but 1 does not?

tasks/eth/3-add-superchainConfig/justfile
```diff
@@ -4,47 +4,115 @@ export rpcUrl := env_var('ETH_RPC_URL')
...
+  fi
+  echo ""
+  forge script NestedSignFromJson \
+    --rpc-url ${rpcUrl} \
+    --sender ${signer} \
+    --sig "signJson(string,address)" \
+    "${location}/${bundle}.json" \
+    "${councilSafe}"
+
+simulate-foundation hdPath='0':
```
mds1, line 35: Do we intentionally duplicate the logic instead of adding a new `safeKind = foundation | council` parameter to avoid any breaking changes to UX? https://just.systems/man/en/chapter_38.html

maurelian: I had a bit of difficulty getting the `just` syntax to work for that, so I just did this, but I agree it's worth investigating further.

tasks/eth/3-add-superchainConfig/Validation.md
```diff
@@ -20,13 +20,30 @@ will begin querying it for the paused status.
...
+  **Meaning:** The number of owners is set to 1.
+- **Key:** 0x0000000000000000000000000000000000000000000000000000000000000004 <br/>
+  **Value:** 0x0000000000000000000000000000000000000000000000000000000000000001
+  **Meaning:** The threshold is set to 1.
+
+- **Key:** 0x316a0aac0d94f5824f0b66f5bbe94a8c360a17699a1d3a233aafcf7146e9f11c <br/>
+  **Value:** 0x0000000000000000000000000000000000000000000000000000000000000001
+- **Key:** 0xe90b7bceb6e7df5418fb78d8ee546e97c83a08bbccc01a0644d599ccd2a7c2e0 <br/>
+  **Value:** 0x000000000000000000000000ca11bde05977b3631167028862be2a173976ca11
+  **Meaning:** The two previous overrides set the owner of the signing safe to the [Multicall](https://etherscan.io/address/0xca11bde05977b3631167028862be2a173976ca11#code) contract. This allows simulating both the approve hash and the final tx in a single Tenderly tx.
```
mds1, line 45: These keys look like mapping slots? Would be nice to explain the mapping names and key values

tasks/eth/3-add-superchainConfig/Validation.md
```diff
@@ -20,13 +20,30 @@ will begin querying it for the paused status.
...
+
```
mds1, line 46: Before value for `0x229047fed2591dbec1ef1118d64f7af3db9eb290` (which is below) is nonzero but is not shown. README says:

> In order to minimize the validation effort, **"Before"** values are only included below when they are non-zero prior to the upgrade.

tasks/eth/3-add-superchainConfig/Validation.md
```diff
@@ -60,6 +77,53 @@ There should also be a single 'State Override' in the Foundation Safe contract
...
+      - The address of the Council Safe: `0xc2819DC788505Aac350142A7A707BF9D03E3Bd03`
+      - The safe hash to approve: `0x782cc13743013d9b1b0854b914bba5ebde49971466bf982c98d4ba911eb0d42d`
+    - The using `cast index`, we can verify that:
+      ```shell
+        $ cast index address 0xc2819DC788505Aac350142A7A707BF9D03E3Bd03 8
+        0xaaf2b641eaf0bae063c4f2e5670f905e1fb7334436b902d1d880b05bd6228fbd
+        ```
+        and
+      ```shell
+        $ cast index bytes32 0x782cc13743013d9b1b0854b914bba5ebde49971466bf982c98d4ba911eb0d42d 0xaaf2b641eaf0bae063c4f2e5670f905e1fb7334436b902d1d88
```
mds1, line 102:
```suggestion
        $ cast index bytes32 0x782cc13743013d9b1b0854b914bba5ebde49971466bf982c98d4ba911eb0d42d 0xaaf2b641eaf0bae063c4f2e5670f905e1fb7334436b902d1d880b05bd6228fbd
```

tasks/eth/3-add-superchainConfig/Validation.md
```diff
@@ -60,6 +77,53 @@ There should also be a single 'State Override' in the Foundation Safe contract
...
+        0x13908ba1c0e379ab58c6445554ab471f3d4efb06e3c4cf966c4f5e918eca67bd
+      ```
+      and
+      ```shell
+        $ cast index bytes32 0x782cc13743013d9b1b0854b914bba5ebde49971466bf982c98d4ba911eb0d42d 0x13908ba1c0e379ab58c6445554ab471f3d4efb06e3c4cf966c4f5e918eca67bd
+        0x66dfee7d20e8ae2c45828a6a3c2c79c377eccb8b4cea869195a802469fe70584
+      ```
+      And so the output of the second command matches the key above.
 
 ### `0x5a7749f83b81b301cab5f48eb8516b986daef23d` (`L1ERC721BridgeProxy`)
```
mds1, line 128: _(arbitrary comment location to start a thread)_ 

I did not validate existing state changes, as presumably thr prior reviewer validated those

tasks/eth/3-add-superchainConfig/justfile
mds1: Tracking in https://github.com/ethereum-optimism/superchain-ops/issues/80, will resolve this

tasks/eth/3-add-superchainConfig/Validation.md
mds1: It was easy for me to follow so I'm good with keeping it 👍

tasks/eth/3-add-superchainConfig/Validation.md
maurelian: I think that's valid.

tasks/eth/3-add-superchainConfig/Validation.md
maurelian: OK, I've added an explanation for these values, it's hard to know how much detail to go into here, but I think this is good.

tasks/eth/3-add-superchainConfig/Validation.md
maurelian: That is just an implementation address change, so I don't think it's helpful to include that. It now says: 

> In order to minimize the validation effort, **"Before"** values are only included below when they are not implementation addresses being swapped out, and when they are non-zero prior to the upgrade.

tasks/eth/3-add-superchainConfig/Validation.md
maurelian: fixed that, thank you.

script/NestedSignFromJson.s.sol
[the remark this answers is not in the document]
maurelian: Yes, that is correct.

tasks/eth/3-add-superchainConfig/Validation.md
```diff
@@ -20,13 +20,45 @@ will begin querying it for the paused status.
...
+- **Key:** 0x316a0aac0d94f5824f0b66f5bbe94a8c360a17699a1d3a233aafcf7146e9f11c <br/>
+  **Value:** 0x0000000000000000000000000000000000000000000000000000000000000001
+  **Meaning:** This is `owners[0xca11bde05977b3631167028862be2a173976ca11] -> 1`, so the key can be
+    derived from `cast index address 0xca11bde05977b3631167028862be2a173976ca11 2`.
+
+
+- **Key:** 0xe90b7bceb6e7df5418fb78d8ee546e97c83a08bbccc01a0644d599ccd2a7c2e0 <br/>
+  **Value:** 0x000000000000000000000000ca11bde05977b3631167028862be2a173976ca11
+  **Meaning:** This is `owners[1] -> 0xca11bde05977b3631167028862be2a173976ca11`, so the key can be
+    derived from `cast index address 0x0000000000000000000000000000000000000001`.
```
mds1, line 60:
```suggestion
    derived from `cast index address 0x0000000000000000000000000000000000000001 2`.
```

tasks/eth/3-add-superchainConfig/Validation.md
mds1: Maybe a good rule of thumb is just "no unexplained 32 byte strings"? For example hashes, mapping slots, etc should all have an explanation of "why that hash/slot". Either way, new explanation lgtm, but leaving this unresolved so you see that proposal—feel free to resolve yourself

tasks/eth/3-add-superchainConfig/justfile
```diff
@@ -4,47 +4,89 @@ export rpcUrl := env_var('ETH_RPC_URL')
...
-  if [ t"$SIMULATE_WITHOUT_LEDGER" == t1 ]
-  then
+  if [ t"{{whichSafe}}" == t"foundation" ]; then
+    safe="${foundationSafe}"
     signer="0x42d27eEA1AD6e22Af6284F609847CB3Cd56B9c64"
-  else
+  fi
+  if [ t"{{whichSafe}}" == t"council" ]; then
+    safe="${councilSafe}"
+    signer="0x07dC0893cAfbF810e3E72505041f2865726Fd073"
```
mds1, line 24: What are the hardcoded signer addresses here? A comment or pulling them into a var would be more clear

tasks/eth/3-add-superchainConfig/Validation.md
maurelian: good catch TYVM!

tasks/eth/3-add-superchainConfig/justfile
maurelian: An env var in the `justfile` makes sense to me. WDYT about deleting `.env` file entirely and having all these values set in the header of this `justfile`?

maurelian: Actually I think we need to keep the `.env` file for the sake of the nonce which is read elsewhere (the forge scripts).

mds1: Looks like you could remove the `.env` file by having the nonce be another input, and export it from the justfile https://just.systems/man/en/chapter_37.html

> Parameters prefixed with a $ will be exported as environment variables

But I don't feel too strongly here

maurelian: I've discovered that we also have the issue with the `OWNER_SAFE` so I'll keep the .env for now. 

Once we're down to a single justfile in the root I think it's likely we'll want to keep the .env.

tasks/eth/3-add-superchainConfig/Validation.md
maurelian: Resolving. I will seek feedback from the signers about the level of information during the ceremony.

tasks/eth/3-add-superchainConfig/Validation.md
maurelian: Fixed locally.

tasks/eth/3-add-superchainConfig/justfile
maurelian: Fixed: ae4503e843244a6050e4c7092bf3faa14c7f5417

mds1 (approved):

PR #82, opened 2024-02-22
Use a 00 prefix on eth and sep tasks to ensure correct ordering for up to 999

maurelian:
This prefixes the `eth` and `sep` task dirs with `00`, and fixes the ordering of the ceremonies so that the delta upgrade is finalized first.

Note: this PR builds on #83 and should be merged second.


maurelian:
@zchn's review was dismissed when the base branch was merged to main in #83. However nothing else has changed, so I will bypass the review requirement.

PR #86, opened 2024-02-24
build: privatize .env

pegahcarter:
**Description**

PR to follow the convention of having the `.env` within `.gitignore` and using an `.env-example`. This provides future operational security if a superchain op requires sensitive data as the `.env` used will no longer be public.

**Tests**

N/A

**Additional context**

N/A

**Metadata**

N/A


maurelian:
Thanks for this PR, I appreciate the suggestion. 

However for now I think we will keep it as is because it makes things easier for signers, and we don't have plans to add secrets to these .env files.

PR #92, opened 2024-03-04
mainnet: Ecotone blobs gas config task

sebastianst:
**Description**

Task for reconfiguring the gas config on the `SystemConfig` for blob transactions.

This PR also contains a change to the `single.just` file which was needed to get simulation working locally:
```
export OWNER_SAFE := env_var_or_default('FOUNDATION_SAFE', '0x847B5c174615B1B7fDF770882256e2D3E95b9D92')
```
I refer to reviewers on how to fix this.

**Tests**

Added CI test `just_simulate_006-ecotone-set-gas-config`

**Additional context**

Needed to use blob batcher transactions after Ecotone activation on `Mar 14 00:00:01 UTC`.

**Metadata**

- Fixes https://github.com/ethereum-optimism/protocol-quest/issues/155


.circleci/config.yml
```diff
@@ -201,6 +201,26 @@ jobs:
             export SIMULATE_WITHOUT_LEDGER=1
             just simulate
 
+  just_simulate_006-ecotone-set-gas-config:
```
mds1, line 204: What do you think of calling these tasks `005-01-protocol-versions-ecotone` and `005-02-ecotone-set-gas-config`? The rationale being that task `006-...` will keep it's numbering corresponding to [upgrade 6](https://gov.optimism.io/t/upgrade-proposal-6-multi-chain-prep-mcp-l1/7677)

tasks/eth/006-ecotone-set-gas-config/.env
```diff
@@ -0,0 +1,3 @@
+ETH_RPC_URL="https://ethereum.publicnode.com"
+FOUNDATION_SAFE="0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A"
+SAFE_NONCE=""
```
maurelian, line 3: Let's hard code both this value and the one in the 005 ecotone ceremony.

tasks/eth/006-ecotone-set-gas-config/README.md
```diff
@@ -0,0 +1,76 @@
...
+blob batch transactions.
+
+The proposal was:
+- [ ] approved by Token House voting here: TODO 
+- [ ] not vetoed by the Citizens' house here: TODO
+
+Following [internal calculations](https://github.com/ethereum-optimism/protocol-quest/issues/84#issuecomment-1977330679),
+the basefee and blob basefee scalars are updated to
+* `L1BaseFeeScalar: 1353`
+* `BlobBaseFeeScalar: 810945`
```
maurelian, line 17: Is there something public we can share here? What justification did we show to governance for these values?

tasks/eth/006-ecotone-set-gas-config/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+This encoding follows the [spec change for Ecotone fee scalars](https://github.com/ethereum-optimism/specs/blob/11099e9908bb7bfa640d73b2a3a2349bef9ab7a1/specs/protocol/system_config.md#scalars).
+Notably, in version 1 of the scalar encoding format, the `overhead` is set to 0 and the old `scalar` field
+now encodes the _base fee scalar_ as well as the _blob base fee scalar_ in a packed format.
+The first byte of the `scalar` field denotes the version `0x01`.
+
+The transaction was created in the root directory with
+
+```
+just add-transaction tasks/eth/006-ecotone-set-gas-config/inputs/op.json 0x229047fed2591dbec1eF1118d64F7aF3dB9EB290 'setGasConfig(uint256,uint256)' 0 0x010000000000000000000000000000000000000000000000000c5fc100000549
```
maurelian, line 48:
```suggestion
just add-transaction tasks/eth/006-ecotone-set-gas-config/input.json 0x229047fed2591dbec1eF1118d64F7aF3dB9EB290 'setGasConfig(uint256,uint256)' 0 0x010000000000000000000000000000000000000000000000000c5fc100000549
```

tasks/eth/006-ecotone-set-gas-config/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+### State 
+
+On the "State" tab, you can verify that the following two state changes occured on the `SystemConfigProxy` at `0x229047fed2591dbec1ef1118d64f7af3db9eb290`:
+
+* `overhead` (key `0x65`) changed from `0xbc = 188` to `0`
+* `scalar` (key `0x66`) changed
+  * from `0x00000000000000000000000000000000000000000000000000000000000a6fe0 = 684000`
+  * to `0x010000000000000000000000000000000000000000000000000c5fc100000549`
+    `= 452312848583266388373324160190187140051835877600158453279134670513164518729`
```
maurelian, line 65: formatting feels off, should it be this?
```suggestion
  * to `0x010000000000000000000000000000000000000000000000000c5fc100000549 = 452312848583266388373324160190187140051835877600158453279134670513164518729`
```

maurelian: Also, what is the reason for showing the decimal representation of that number? It would be more useful to show how to decode the hex value into the actual config being used.

tasks/eth/006-ecotone-set-gas-config/README.md
sebastianst: This is based on calculations by @K-Ho and @MSilb7 - will ask them what we can provide here. Maybe we share the rationale behind this internally first for the purpose of this singing ceremony, and explain where these values come from in a later public post?

tasks/eth/006-ecotone-set-gas-config/.env
sebastianst: To what?

tasks/eth/006-ecotone-set-gas-config/README.md
sebastianst: I just showed the decimal for completion, because both the script and state diff show both. Happy to remove the decimal representation altogether.
I can add the decoding command as well.

maurelian: Yes to replacing decimals with decoding command. TY

tasks/eth/006-ecotone-set-gas-config/.env
maurelian:
```
`cast call 0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A 'nonce()(uint)'`
90
```

So, this one should be 91.

For simplicity we can leave 005 asis.

tasks/eth/006-ecotone-set-gas-config/README.md
MSilb7: Hi yeah actually 2 things:
1. [This sheet (which should be public)](https://docs.google.com/spreadsheets/d/12VIiXHaVECG2RUunDSVJpn67IQp9NHFJqUsma2PndpE/edit#gid=186414307) - Configured to:
- Transactions per day: 360,000
- Transaction Type : OP Mainnet
- op-batcher Max Channel Duration (hours): 5
- Target Data Margin: 5%

2. I can make a comment elsewhere too, but now handling for "batch overhead" and "blob overhead" slightly reducing the available bytes per blob of transaction data, scalars by this calculator should be:
* `L1BaseFeeScalar: 1367`
* `BlobBaseFeeScalar: 810949`

Super super tiny mods, but likely worth updating. I'll update in https://github.com/ethereum-optimism/protocol-quest/issues/84#issuecomment-1977330679 as well

MSilb7: ^ updated the L1BaseFeeScalar to match the sheet

tasks/eth/006-ecotone-set-gas-config/.env
sebastianst: How do we simulate this when we set it to 91? IIUC CI would fail with this.

.circleci/config.yml
sebastianst: renamed this one to `005-2`, can I leave the other as is?

tasks/eth/006-ecotone-set-gas-config/README.md
MSilb7: I could do a quick dev blog on how we calculated if needed as well. Could be good content as part of a series on "where fees come from," and "how we solve/handle for various forms of congestion" which I've been thinking about.

tasks/eth/006-ecotone-set-gas-config/.env
maurelian: CI should not fail. 
1. if a custom SAFE_NONCE is specified, it will be added to the State Overrides used in the simulation
2. CI just tests that the `just simulate` invocation passes, and unfortunately does not have the ability to simulate the actual execution result. #34 or something similar will fix this some day.

tasks/eth/006-ecotone-set-gas-config/README.md
sebastianst: Note that the calculator now shows `810949` and `1368` (1 higher). These values are also in the internal issue.

sebastianst: updated readme to mention public calculator and new scalar values.

tasks/eth/006-ecotone-set-gas-config/.env
sebastianst: got it, set it to 91

PR #95, opened 2024-03-06
mainnet: Ecotone contigency calldata gas config

sebastianst:
**Description**

Ecotone contingency calldata gas config task. Based on #92 

**Tests**

Added CI test `just_simulate_005-3-ecotone-calldata-gas-config`

**Additional context**

Highly unlikely to be executed, but needs to be prepared to run on top of 005-2 (so one nonce higher)


.circleci/config.yml
```diff
@@ -153,6 +153,46 @@ jobs:
...
+      - checkout
+      - run:
+          name: just simulate 005-2-ecotone-set-gas-config
+          command: |
+            just install
+            forge clean
+            forge build
+            cd tasks/eth/005-2-ecotone-set-gas-config
+            export SIMULATE_WITHOUT_LEDGER=1
+            export BUNDLE_NAME=input
```
maurelian, line 170: Don't need this, it's the default value

single.just
```diff
@@ -7,7 +7,7 @@ export taskPath := invocation_directory()
 export bundlePath := "${taskPath}/${bundleName}.json"
 
 # Accounts
-export OWNER_SAFE := "0x847B5c174615B1B7fDF770882256e2D3E95b9D92"
+export OWNER_SAFE := env_var_or_default('FOUNDATION_SAFE', '0x847B5c174615B1B7fDF770882256e2D3E95b9D92')
```
maurelian, line 10: This is an error on my part, but let's fix it here: 
```suggestion
export OWNER_SAFE := env_var_or_default('FOUNDATION_SAFE', '0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A')
```

tasks/eth/005-2-ecotone-set-gas-config/.env
```diff
@@ -0,0 +1,3 @@
+ETH_RPC_URL="https://ethereum.publicnode.com"
+FOUNDATION_SAFE="0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A"
```
maurelian, line 2: We could remove this with the suggested fix above, but I say we keep it for clarity.

single.just
sebastianst: done

.circleci/config.yml
sebastianst: ok, just copied it from other task

sebastianst:
Looks like the vote will pass, changed in README

tasks/eth/005-3-ecotone-calldata-gas-config/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+Please see the signing and execution instructions in [SINGLE.md](../../../SINGLE.md).
+
+## Validations
+
+### State 
+
+On the "State" tab, you can verify that the following two state changes occured on the `SystemConfigProxy` at `0x229047fed2591dbec1ef1118d64f7af3db9eb290`:
+
+* `overhead` (key `0x65`) changed from `0xbc = 188` to `0` (only before execution of task `005-2`)
```
maurelian, line 74: Is this correct? It's the exact same change [listed in 005-2](https://github.com/ethereum-optimism/superchain-ops/blob/seb/ecotone-calldata-contigency-gas-config/tasks/eth/005-2-ecotone-set-gas-config/README.md#L69).

Also, why do we go into extra detail for `scalar` but not overhead?

sebastianst: `overhead` just won't change after `005-2` got executed, because it will already be `0`.

maurelian (approved):

PR #104, opened 2024-03-11
Fix Bundle Path Handling

maurelian:
**Description**

Removes an unused argument from the single sign and execute commands and fixes bundlePath setting in the nested justfile

### DIFF
--- a/nested.just
+++ b/nested.just
@@ -17,7 +17,7 @@ export randomPersonEoa := "0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045"
 
 simulate whichSafe hdPath='0':
   #!/usr/bin/env bash
-  bundlePath=${taskPath}/${bundle}.json
+  bundlePath=${taskPath}/${bundleName}.json
   echo "getting signer address..."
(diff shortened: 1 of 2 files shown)


nested.just
```diff
@@ -18,6 +17,7 @@ export randomPersonEoa := "0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045"
 
 simulate whichSafe hdPath='0':
   #!/usr/bin/env bash
+  bundlePath=${taskPath}/${bundle}.json
```
mds1, line 20: looks like this applies in a few places
```suggestion
  bundlePath=${taskPath}/${bundleName}.json
```

single.just
```diff
@@ -42,9 +41,9 @@ sign taskPath hdPath='0':
     --sig "signJson(string)" \
     ${bundlePath}
 
-execute taskPath hdPath='0':
+execute  hdPath='0':
```
mds1, line 44:
```suggestion
execute hdPath='0':
```

mds1 (approved):

PR #109, opened 2024-03-14
Add validate commands for local testing of signatures

maurelian:
**Description**

Adds a new method for validating signatures on a persistent local network fork.

Benefits of this approach vs. simply removing `--broadcast`: 
1. Less of a potential for error leading to actually submitting the tx
2. allows for manual inspection of the network after execution (ie. the safe nonce can be observed to have incremented.

### DIFF
--- a/NESTED.md
+++ b/NESTED.md
@@ -178,8 +178,11 @@ congrats, you are done!
    environment variable, i.e. `export
    SIGNATURES="0x[SIGNATURE1][SIGNATURE2]..."`.
 3. Run the `just approve` command as described below to approve the transaction in each multisig.
+### Concatenate the signatures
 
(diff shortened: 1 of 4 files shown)


maurelian:
> Should we de-prioritize this until MCP L1 is successfully upgrades to give @mds1 more time to focus on MCP L1 (unless this is useful for MCP L1 as well)

That SGTM. 

This is a pretty simple change which I think is valuable but low priority. It was especially useful for ecotone where the timing of the transaction was crucial so we wanted to know ahead of time that it would execute successfully.

It's less useful for a non-timing sensitive upgrade like MCP-L1. It should also be easy to rebase if necessary when the time comes.

PR #115, opened 2024-03-18
feat: configurable no ledger usage

tynes:
**Description**

Enables the ability to not use a ledger when
the env var `SIMULATE_WITHOUT_LEDGER` is set.
It just needs to be set to anything to work

### DIFF
--- a/single.just
+++ b/single.just
@@ -13,7 +13,7 @@ simulate hdPath='0':
   #!/usr/bin/env bash
   bundlePath=${taskPath}/${bundleName}.json
   echo "getting signer address..."
-  if [ t"$SIMULATE_WITHOUT_LEDGER" == t1 ]
+  if [ ! -z "$SIMULATE_WITHOUT_LEDGER" ]
   then
     signer=${foundationOwner0}
   else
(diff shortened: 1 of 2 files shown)


single.just
```diff
@@ -13,7 +13,7 @@ simulate hdPath='0':
   #!/usr/bin/env bash
   bundlePath=${taskPath}/${bundleName}.json
   echo "getting signer address..."
-  if [ t"$SIMULATE_WITHOUT_LEDGER" == t1 ]
+  if [ ! -z "$SIMULATE_WITHOUT_LEDGER" ]
```
maurelian, line 16: The previous command doesn't work, or this is an improvement because it'll accept any non-empty value? 

Can you also apply this in nested.just?

tynes: I am unaware of the syntax that was used previously. What did you intent for it to do? Now it simply says "if the variable is not zero" which seems like better UX in this case

maurelian: Added a commit doing the same on `nested.just`

maurelian (approved):

PR #117, opened 2024-03-22
readme: note some info on adding a task

tynes:
**Description**

Updates the readme to document the requirement of implementing
the `_postCheck` hook so that we can rely more on automation
of checking the correctness of upgrade scripts

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ Each task should contain the following:
 
 - `README.md`: A brief markdown file describing the task to be executed.
 - `Validation.md`: A markdown file describing and justifying the expected state changes for manual validation by multisig signers.
+- `Script.s.sol`: A foundry script that implements post upgrade assertions.
 - `input.json`: A JSON file which defines the specific transaction for the task to be executed. This file may either be generated automatically or manually created.
 - `.env`: a place to store environment variables specific to this task
 
@@ -117,3 +118,18 @@ Move into the repo and install the contract dependencies
 Tenderly is used to simulate transactions.
 If you don’t already have a Tenderly account, go to https://dashboard.tenderly.co/login and sign up.
 The free account is sufficient.
+
+## Creating a Task
+
+Each task in the `tasks` directory should contain all of the information required to both validate
+and perform the network interaction. To validate that the intention of the network interaction is
+correct, specific contextual information must be included in `Validation.md` that enables signers
+to double check the correctness. This includes linking to authoritative sources such as the
+[superchain-registry](https://github.com/ethereum-optimism/superchain-registry) or
+[Etherscan](https://etherscan.io) to prove the correctness of particular configuration values.
+
+The base scripts for performing network interactions found in the `script` directory should
+be inherited from with the `Script.s.sol` file where the `_postCheck` hook is implemented.
+This function should make assertions on the post state of the network interaction.
+It will run each time that a signer generates a signature, giving additional automation
+on validating the correctness of the network interaction.
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ Each task in the `tasks` directory should contain all of the information require
 and perform the network interaction. To validate that the intention of the network interaction is
 correct, specific contextual information must be included in `Validation.md` that enables signers
 to double check the correctness. This includes linking to authoritative sources such as the
-[superchain-registry](https://github.com/ethereum-optimism/superchain-registry) or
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -108,3 +109,18 @@ Move into the repo and install the contract dependencies
...
 The free account is sufficient.
+
+## Creating a Task
+
+Each task in the `tasks` directory should contain all of the information required to both validate
+and perform the network interaction. To validate that the intention of the network interaction is
+correct, specific contextual information must be included in `Validation.md` that enables signers
+to double check the correctness. This includes linking to authoritative sources such as the
+[superchain-registry](https://github.com/ethereum-optimism/superchain-registry) or
+[Etherscan](https://etherscan.io) to prove the correctness of particular configuration values.
```
maurelian, line 120: Adding the monorepo, as it is referenced by tag in governance proposals:

```suggestion
to double check the correctness. This includes linking to authoritative sources such as the
[superchain-registry](https://github.com/ethereum-optimism/superchain-registry), the [Optimism monorepo](https://github.com/ethereum-optimism/optimism), and
[Etherscan](https://etherscan.io) to prove the correctness of particular configuration values.
```

README.md
```diff
@@ -20,6 +20,7 @@ Each task should contain the following:
 
 - `README.md`: A brief markdown file describing the task to be executed.
 - `Validation.md`: A markdown file describing and justifying the expected state changes for manual validation by multisig signers.
+- `Script.s.sol`: A foundry script that implements post upgrade assertions.
```
maurelian, line 23: We'd like to move away from code towards json spec files, so I'd like to use a more specific name to avoid the scope of this file from creeping overtime. 


```suggestion
- `Checks.s.sol`: A foundry script that implements post upgrade assertions.
```

README.md
```diff
@@ -108,3 +109,18 @@ Move into the repo and install the contract dependencies
...
+
+Each task in the `tasks` directory should contain all of the information required to both validate
+and perform the network interaction. To validate that the intention of the network interaction is
+correct, specific contextual information must be included in `Validation.md` that enables signers
+to double check the correctness. This includes linking to authoritative sources such as the
+[superchain-registry](https://github.com/ethereum-optimism/superchain-registry) or
+[Etherscan](https://etherscan.io) to prove the correctness of particular configuration values.
+
+The base scripts for performing network interactions found in the `script` directory should
+be inherited from with the `Script.s.sol` file where the `_postCheck` hook is implemented.
```
maurelian, line 123:
```suggestion
be inherited from with the `Checks.s.sol` file where the `_postCheck` hook is implemented.
```

README.md
mds1: What about even more specific with `PostCheck.s.sol` to match the function name being implemented?

README.md
mds1: We should be migrating away from using the monorepo as a source of truth for anything except bytecode validation, so I'd suggest linking to the governance forum here instead. And even for bytecode validation, the starting point would be a gov forum post, which then links to a release in the monorepo

README.md
mds1: https://github.com/ethereum-optimism/superchain-ops/pull/117/files#r1538081114

README.md
```diff
@@ -108,3 +109,18 @@ Move into the repo and install the contract dependencies
...
+This function should make assertions on the post state of the network interaction.
```
mds1, line 124: nit
```suggestion
This function must make assertions on the post-transaction state of the network interaction.
```

README.md
tynes: So the script will both sign and do the post check assertions because it inherits `SignFromJson`

maurelian:
@mark Do you want to keep this PR open?

mds1:
We don't have this info documented, even though it's our new standard, so I do think we should get this merged

README.md
mds1: We've ended up naming these based on the file it inherits from, either `SignFromJson.s.sol` or `NestedSignFromJson.s.sol`, so let's use that here also

README.md
maurelian: The place I think we'll have the most difficulty avoiding the monorepo is for storage layout references which we do in various locations for some runbooks.

mds1: Yea I think storage layouts/snapshots are the only time we should need to reference it

maurelian: OK. Please let me know how you think I should resolve this thread, and anything else for the PR itself.

mds1 (approved):

PR #119, opened 2024-03-30
Feat/mcp: implement post check script

0xfuturistic:


tasks/eth/006-MCP-L1/postCheck.s.sol
```diff
@@ -0,0 +1,231 @@
...
+        OptimismPortal2: 0xbEb5Fc579115071764c7423A4f12eDde41f106Ed,
+        SystemConfig: 0x229047fed2591dbec1eF1118d64F7aF3dB9EB290,
+        L1ERC721Bridge: 0x5a7749f83b81B301cAb5f48EB8516B986DAef23D,
+        ProtocolVersions: 0x8062AbC286f5e7D9428a0Ccb9AbD71e50d93b935,
+        SuperchainConfig: 0x95703e0982140D16f8ebA6d158FccEde42f04a4C
+    });
+    DeployConfig cfg = new DeployConfig();
+    uint256 l2OutputOracleStartingTimestamp = 1686068903;
+
+    constructor() {
```
tynes, line 38: Can we inherit from a contract in the monorepo that would let us not need to copy/paste a hardcoded config instead but instead use an env var to configure the path at which the deploy config json is read from?

tynes: This file can handle it: https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/scripts/Deployer.sol

tasks/eth/006-MCP-L1/postCheck.s.sol
```diff
@@ -0,0 +1,231 @@
...
+        rcfg.maxResourceLimit = SystemConfig(prox.SystemConfig).resourceConfig().maxResourceLimit;
+        rcfg.elasticityMultiplier = SystemConfig(prox.SystemConfig).resourceConfig().elasticityMultiplier;
+        rcfg.baseFeeMaxChangeDenominator = SystemConfig(prox.SystemConfig).resourceConfig().baseFeeMaxChangeDenominator;
+        rcfg.minimumBaseFee = SystemConfig(prox.SystemConfig).resourceConfig().minimumBaseFee;
+        rcfg.systemTxMaxGas = SystemConfig(prox.SystemConfig).resourceConfig().systemTxMaxGas;
+        rcfg.maximumBaseFee = SystemConfig(prox.SystemConfig).resourceConfig().maximumBaseFee;
+        ResourceMetering.ResourceConfig memory dflt;
+        dflt.maxResourceLimit = Constants.DEFAULT_RESOURCE_CONFIG().maxResourceLimit;
+        dflt.elasticityMultiplier = Constants.DEFAULT_RESOURCE_CONFIG().elasticityMultiplier;
+        dflt.baseFeeMaxChangeDenominator = Constants.DEFAULT_RESOURCE_CONFIG().baseFeeMaxChangeDenominator;
```
tynes, line 61: this could be modularized so not so many loc are needed? can just do 2 total calls? why split up

tasks/eth/006-MCP-L1/postCheck.s.sol
```diff
@@ -0,0 +1,231 @@
...
+        require(address(bridge.otherBridge()) == Predeploys.L2_STANDARD_BRIDGE);
+        require(address(bridge.superchainConfig()) == prox.SuperchainConfig);
+    }
+
+    /// @notice Asserts that the L2OutputOracle is setup correctly
+    function checkL2OutputOracle() internal view {
+        console.log("Running chain assertions on the L2OutputOracle");
+        L2OutputOracle oracle = L2OutputOracle(prox.L2OutputOracle);
+
+        require(oracle.SUBMISSION_INTERVAL() == cfg.l2OutputOracleSubmissionInterval());
```
tynes, line 157: Can we hardcode these values instead of reading them from the config? Then it'll remove misconfig issues

tasks/eth/006-MCP-L1/postCheck.s.sol
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    /// @notice Asserts the OptimismPortal is setup correctly
+    function checkOptimismPortal() internal view {
+        console.log("Running chain assertions on the OptimismPortal");
+
+        OptimismPortal portal = OptimismPortal(payable(prox.OptimismPortal));
+
+        address guardian = cfg.superchainConfigGuardian();
+        if (guardian.code.length == 0) {
```
tynes, line 200: On mainnet, guardian should always have code

tasks/eth/006-MCP-L1/postCheck.s.sol
```diff
@@ -0,0 +1,231 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.15;
+
```
tynes, line 3: We will need to add a justfile in this directory that runs this contract instead of running the base sign from json contract in the root. Perhaps we could rename this to Execute ? Thoughts on naming @maurelian @mds1

tynes:
If we added a pre hook to go with the post hook then it would work really well with the state diff recording cheatcodes. Thoughts @mds1 @maurelian ?

tynes:
We should add assertions that any address referenced has code

tasks/eth/006-MCP-L1/postCheck.s.sol
mds1: Agreed, we should minimize the amount of hardcoding in this PR—the `DeployConfig.json` file can be read via env var like @tynes suggested, and the hardcoded addresses can be read from the superchain registry which can be a dependency

tasks/eth/006-MCP-L1/postCheck.s.sol
mds1: And once https://www.notion.so/oplabs/Proposal-Align-Sepolia-roles-config-with-Mainnet-a1fb54672f514a5aaec65a564671c621?pvs=4 is done, Sepolia will also

tasks/eth/006-MCP-L1/postCheck.s.sol
mds1: What if we instead modify the root justfiles to look for a `PostCheck.s.sol`, and run that if present, and otherwise fallback to the original file?

Also what about renaming to something very explicit like `SignFromJsonWithAssertions.s.sol`?

script/SignFromJson.s.sol
mds1, line 1: From @tynes: 

> We should add assertions that any address referenced has code

_(arbitrary comment location to pull that comment into a conversation to ensure it gets addressed before merge)_

script/SignFromJson.s.sol
```diff
@@ -27,5 +27,5 @@ contract SignFromJson is MultisigBuilder, JsonTxBuilderBase {
         return vm.envAddress("OWNER_SAFE");
     }
 
-    function _postCheck() internal view override {}
+    function _postCheck() internal view virtual override {}
```
mds1, line 30: From @tynes: 

> If we added a pre hook to go with the post hook then it would work really well with the state diff recording cheatcodes. Thoughts @mds1 @maurelian ?

What exactly would you put in the pre-hook? In general I'm likely to be in favor, but perhaps as a follow up PR to keep a tight scope here?

_(arbitrary comment location to pull that comment into a conversation to ensure it gets addressed before merge)_

tasks/eth/006-MCP-L1/postCheck.s.sol
```diff
@@ -0,0 +1,231 @@
...
+import {OptimismPortal} from "@eth-optimism-bedrock/src/L1/OptimismPortal.sol";
+import {L1CrossDomainMessenger} from "@eth-optimism-bedrock/src/L1/L1CrossDomainMessenger.sol";
+import {OptimismMintableERC20Factory} from "@eth-optimism-bedrock/src/universal/OptimismMintableERC20Factory.sol";
+import {L1ERC721Bridge} from "@eth-optimism-bedrock/src/L1/L1ERC721Bridge.sol";
+import {Predeploys} from "@eth-optimism-bedrock/src/libraries/Predeploys.sol";
+import {Types} from "@eth-optimism-bedrock/scripts/Types.sol";
+import {console2 as console} from "forge-std/console2.sol";
+
+contract PostCheck is SignFromJson {
+    Types.ContractSet prox = Types.ContractSet({
```
mds1, line 22: nit on variable name
```suggestion
    Types.ContractSet proxies = Types.ContractSet({
```

tasks/eth/006-MCP-L1/postCheck.s.sol
```diff
@@ -0,0 +1,231 @@
...
+        L1StandardBridge: 0x99C9fc46f92E8a1c0deC1b1747d010903E884bE1,
+        L2OutputOracle: 0xdfe97868233d1aa22e815a266982f2cf17685a27,
+        DisputeGameFactory: address(0),
+        OptimismMintableERC20Factory: 0x75505a97BD334E7BD3C476893285569C4136Fa0F,
+        OptimismPortal: 0xbEb5Fc579115071764c7423A4f12eDde41f106Ed,
+        OptimismPortal2: 0xbEb5Fc579115071764c7423A4f12eDde41f106Ed,
+        SystemConfig: 0x229047fed2591dbec1eF1118d64F7aF3dB9EB290,
+        L1ERC721Bridge: 0x5a7749f83b81B301cAb5f48EB8516B986DAef23D,
+        ProtocolVersions: 0x8062AbC286f5e7D9428a0Ccb9AbD71e50d93b935,
+        SuperchainConfig: 0x95703e0982140D16f8ebA6d158FccEde42f04a4C
```
mds1, line 33: All of these addresses are in the superchain registry. I'd suggest adding the latest commit as a dependency via `forge install` and then reading in all addresses for mainnet that way

tasks/eth/006-MCP-L1/postCheck.s.sol
```diff
@@ -0,0 +1,231 @@
...
+    function postDeployAssertions() internal view {
+        console.log("Running post-deploy assertions");
+        ResourceMetering.ResourceConfig memory rcfg;
+        rcfg.maxResourceLimit = SystemConfig(prox.SystemConfig).resourceConfig().maxResourceLimit;
+        rcfg.elasticityMultiplier = SystemConfig(prox.SystemConfig).resourceConfig().elasticityMultiplier;
+        rcfg.baseFeeMaxChangeDenominator = SystemConfig(prox.SystemConfig).resourceConfig().baseFeeMaxChangeDenominator;
+        rcfg.minimumBaseFee = SystemConfig(prox.SystemConfig).resourceConfig().minimumBaseFee;
+        rcfg.systemTxMaxGas = SystemConfig(prox.SystemConfig).resourceConfig().systemTxMaxGas;
+        rcfg.maximumBaseFee = SystemConfig(prox.SystemConfig).resourceConfig().maximumBaseFee;
+        ResourceMetering.ResourceConfig memory dflt;
```
mds1, line 58: What is `dflt`, default? Let's use explicit variable names everywhere so this is maximially readable for signers:
- `prox` -> `proxies`
- `dflt` -> `expectedResourceConfig`
- `rcfg` -> `actualResourceConfig`
- etc...

tasks/eth/006-MCP-L1/postCheck.s.sol
```diff
@@ -0,0 +1,231 @@
...
+        ResourceMetering.ResourceConfig memory resourceConfig;
+        resourceConfig.maxResourceLimit = config.resourceConfig().maxResourceLimit;
+        resourceConfig.elasticityMultiplier = config.resourceConfig().elasticityMultiplier;
+        resourceConfig.baseFeeMaxChangeDenominator = config.resourceConfig().baseFeeMaxChangeDenominator;
+        resourceConfig.minimumBaseFee = config.resourceConfig().minimumBaseFee;
+        resourceConfig.systemTxMaxGas = config.resourceConfig().systemTxMaxGas;
+        resourceConfig.maximumBaseFee = config.resourceConfig().maximumBaseFee;
+
+        require(config.owner() == cfg.finalSystemOwner());
+        require(config.overhead() == cfg.gasPriceOracleOverhead());
```
mds1, line 92: Let's add reason strings to all `require` statements in this file, otherwise when one fails it will be hard to track down which failed. For brevity, I sometimes use incrementing numbers of 100 (count by 100 so you can add new ones in between without adjusting all numbers), for example:
```solidity
require(config.owner() == cfg.finalSystemOwner(), "100");
require(config.overhead() == cfg.gasPriceOracleOverhead(), "200");
```

then in the future, minimal diff to add a new check
```solidity
require(config.owner() == cfg.finalSystemOwner(), "100");
require(myNewCheck == cfg.myNewCheck(), "150");
require(config.overhead() == cfg.gasPriceOracleOverhead(), "200");
```

tasks/eth/006-MCP-L1/postCheck.s.sol
```diff
@@ -0,0 +1,231 @@
...
+        console.log("Running chain assertions on the L1CrossDomainMessenger");
+        L1CrossDomainMessenger messenger = L1CrossDomainMessenger(prox.L1CrossDomainMessenger);
+
+        require(address(messenger.OTHER_MESSENGER()) == Predeploys.L2_CROSS_DOMAIN_MESSENGER);
+        require(address(messenger.otherMessenger()) == Predeploys.L2_CROSS_DOMAIN_MESSENGER);
+
+        require(address(messenger.PORTAL()) == prox.OptimismPortal);
+        require(address(messenger.portal()) == prox.OptimismPortal);
+        require(address(messenger.superchainConfig()) == prox.SuperchainConfig);
+        bytes32 xdmSenderSlot = vm.load(address(messenger), bytes32(uint256(204)));
```
mds1, line 136: Let's either:
1. Easier: Add a link to the snapshots so this can be validated, and pull this assumption/declaration to the top of the file so it's more visible
2. Harder: Read from the snapshots, by either having the monorepo at the expected tag as a dependency, or taking the path as an input

tasks/eth/006-MCP-L1/postCheck.s.sol
```diff
@@ -0,0 +1,231 @@
...
+        require(ProtocolVersion.unwrap(versions.required()) == cfg.requiredProtocolVersion());
+        require(ProtocolVersion.unwrap(versions.recommended()) == cfg.recommendedProtocolVersion());
+    }
+
+    /// @notice Asserts that the SuperchainConfig is setup correctly
+    function checkSuperchainConfig(bool _isPaused) internal view {
+        console.log("Running chain assertions on the SuperchainConfig");
+        SuperchainConfig superchainConfig = SuperchainConfig(prox.SuperchainConfig);
+        require(superchainConfig.guardian() == cfg.superchainConfigGuardian());
+        require(superchainConfig.paused() == _isPaused);
```
mds1, line 229: For this task we never need `isPaused == true`, so we can hardcode the expected value of `false` and remove the input

tasks/eth/006-MCP-L1/postCheck.s.sol
```diff
@@ -0,0 +1,231 @@
...
+        require(portal.l2Sender() == Constants.DEFAULT_L2_SENDER);
+    }
+
+    /// @notice Asserts that the ProtocolVersions is setup correctly
+    function checkProtocolVersions() internal view {
+        console.log("Running chain assertions on the ProtocolVersions");
+        ProtocolVersions versions = ProtocolVersions(prox.ProtocolVersions);
+        require(versions.owner() == cfg.finalSystemOwner());
+        require(ProtocolVersion.unwrap(versions.required()) == cfg.requiredProtocolVersion());
+        require(ProtocolVersion.unwrap(versions.recommended()) == cfg.recommendedProtocolVersion());
```
mds1, line 221: Perhaps read these values from `tasks/eth/005-protocol-versions-ecotone/input.json`?

tasks/eth/006-MCP-L1/postCheck.s.sol
0xfuturistic: @mds1 question -- this check is only for the MCP upgrade, does it make sense to modify the root justfiles? if so, it could be as simple as adding a line that checks whether this is the MCP upgrade, and if so to look for `SignFromJsonWithAssertions`.s.sol in the task dir for MCP mainnet. the alternative, which is to just include a new justfile in the task dir, seems more elegant imo

tasks/eth/006-MCP-L1/postCheck.s.sol
0xfuturistic: I'm down to make this change, but this would deviate from the `ChainAssertions.s.sol` implementation

tasks/eth/006-MCP-L1/postCheck.s.sol
0xfuturistic: I'm down to make these change, but this would deviate from the `ChainAssertions.s.sol` implementation

tasks/eth/006-MCP-L1/postCheck.s.sol
0xfuturistic: do you know how to read a `.yaml` file from the solidity contract? `ProtocolVersions` and `SuperchainConfig` would be located in `lib/superchain-registry/superchain/configs/mainnet/superchain.yaml`

tasks/eth/006-MCP-L1/postCheck.s.sol
0xfuturistic: I didn't fully understand -- should we hardcode all values from cfg? or just references to `cfg.l2OutputOracleSubmissionInterval()`?

tasks/eth/006-MCP-L1/postCheck.s.sol
0xfuturistic: implemented approach 1. lmk if this is correct! https://github.com/ethereum-optimism/superchain-ops/blob/8630b8c96e7f6244c33bbf3d8b15176282c61148/tasks/eth/006-MCP-L1/SignFromJsonWithAssertions.s.sol#L22-L24

tasks/eth/006-MCP-L1/postCheck.s.sol
tynes: We will want to follow this pattern for all future upgrades and continue to improve on it over time

script/SignFromJson.s.sol
tynes: The pre hook would be the cheatcode for starting state diff recording then the post hook would stop and return the state diffs. I am fine with doing that in a follow up but do think its a good way to force being exhaustive

tasks/eth/006-MCP-L1/postCheck.s.sol
tynes: There was a cheatcode for this, not sure if its been merged yet, otherwise will need some crazy bash with something like `yq`

tasks/eth/006-MCP-L1/DeployConfig.json
```diff
@@ -0,0 +1,45 @@
+{
+  "finalSystemOwner": "0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A",
+  "superchainConfigGuardian": "0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A",
+  "l1StartingBlockTag": "0x438335a20d98863a4c0c97999eb2481921ccd28553eac6f913af7c12aec04108",
+  "l1ChainID": 1,
+  "l2ChainID": 10,
```
tynes, line 6: We should be able to delete this JSON file completely

tasks/eth/006-MCP-L1/SignFromJsonWithAssertions.s.sol
```diff
@@ -0,0 +1,274 @@
...
+import {SuperchainConfig} from "@eth-optimism-bedrock/src/L1/SuperchainConfig.sol";
+import {OptimismPortal} from "@eth-optimism-bedrock/src/L1/OptimismPortal.sol";
+import {L1CrossDomainMessenger} from "@eth-optimism-bedrock/src/L1/L1CrossDomainMessenger.sol";
+import {OptimismMintableERC20Factory} from "@eth-optimism-bedrock/src/universal/OptimismMintableERC20Factory.sol";
+import {L1ERC721Bridge} from "@eth-optimism-bedrock/src/L1/L1ERC721Bridge.sol";
+import {Predeploys} from "@eth-optimism-bedrock/src/libraries/Predeploys.sol";
+import {Types} from "@eth-optimism-bedrock/scripts/Types.sol";
+import {console2 as console} from "forge-std/console2.sol";
+import {stdJson} from "forge-std/StdJson.sol";
+
```
tynes, line 20: Hmm i bet its not possible to use `Deployer` because it inherits from script, seems like we need more decoupling

https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/scripts/Deployer.sol

That would save all of the logic of needing to set deploy config contract

tasks/eth/006-MCP-L1/postCheck.s.sol
mds1: I think the suggestion here is that this script is for a specific task—the mainnet MCP upgrade—as a result we know precisely what values should be and therefore can hardcode expected values. For example, we do not want to change the submission interval as part of MCP, so let's hardcode the current value so we assert it was not changed

tasks/eth/006-MCP-L1/postCheck.s.sol
mds1: Agreed with @tynes. There is a balance to strike here—some things, such as https://github.com/ethereum-optimism/superchain-ops/pull/119#discussion_r1548234133, we do not want to generalize and can hardcode. Others, such as modifying the root justfile, are worth generalizing now to avoid introducing tech debt

script/SignFromJson.s.sol
mds1: Great, let's track in an issue and resolve this thread once the issue is created

tasks/eth/006-MCP-L1/postCheck.s.sol
mds1: That seems ok to me, unless there is a reason we need to follow the chain assertion variable names? I'd prefer optimizing for readability here

tasks/eth/006-MCP-L1/postCheck.s.sol
mds1: Resolving in favor of this thread: https://github.com/ethereum-optimism/superchain-ops/pull/119#discussion_r1546717485

tasks/eth/006-MCP-L1/postCheck.s.sol
mds1: Toml reading/writing cheats were merged by not yaml. We have a root `just install` command that installs deps, so I'm ok with adding a new dep on yq or yj as long as we unsure the new install command is cross-platform (e.g. no `brew install`). The easiest way is probably do one ffi call that converts the yaml to json in a temp dir and returns the tmp filepath, and then just use the JSON cheats to read that file and delete it afterwards

tasks/eth/006-MCP-L1/SignFromJsonWithAssertions.s.sol
```diff
@@ -0,0 +1,274 @@
...
+contract PostCheck is SignFromJson {
+    /// @dev Verify that the slot number of xDomainMsgSender in lib/optimism/packages/contracts-bedrock/snapshots/storageLayout/L1CrossDomainMessenger.json
+    ///      is 24. If not, update xdmSenderSlotNumber to the correct slot number.
+    uint256 constant xdmSenderSlotNumber = 24;
```
mds1, line 24: I see slot 204, which was what you originally had. When I run `just simulate` it also passes. We should update CI/etc. accordingly to catch this mistake

tasks/eth/006-MCP-L1/postCheck.s.sol
mds1: Thank you!

script/SignFromJson.s.sol
0xfuturistic: issue: https://github.com/ethereum-optimism/superchain-ops/issues/120

tasks/eth/006-MCP-L1/SignFromJsonWithAssertions.s.sol
0xfuturistic: good idea!

tasks/eth/006-MCP-L1/SignFromJsonWithAssertions.s.sol
0xfuturistic: decided to make changes to read `xDomainMsgSender` directly from snapshot!

tasks/eth/006-MCP-L1/postCheck.s.sol
0xfuturistic: I believe that no changes were made to `DeployConfig` as part of the upgrade. does this mean we can hardcode all values from `DeployConfig`? if so, what's the point of still importing the contract? we could skip it altogether it seems?

tasks/eth/006-MCP-L1/SignFromJsonWithAssertions.s.sol
mds1: Awesome. Can we still ensure that CI and a local simulation would fail if the wrong `xDomainMsgSender` slot is used? The purpose of this PR is to catch state errors and we clearly missed one, so it makes me nervous that everything passed when we had that bug

tasks/eth/006-MCP-L1/SignFromJsonWithAssertions.s.sol
```diff
@@ -0,0 +1,296 @@
...
+
+        require(proxies.ProtocolVersions.code.length != 0, "6700");
+
+        ProtocolVersions versions = ProtocolVersions(proxies.ProtocolVersions);
+
+        require(versions.owner() == cfg.finalSystemOwner(), "6800");
+
+        string memory json;
+
+        try vm.readFile(string.concat(vm.projectRoot(), "/tasks/eth/005-protocol-versions-ecotone/input.json"))
```
tynes, line 208: This shows that we should be using the deploy config as the source of truth whenever we do state modifications

tasks/eth/006-MCP-L1/SignFromJsonWithAssertions.s.sol
```diff
@@ -0,0 +1,296 @@
...
+        _proxies.L1ERC721Bridge = stdJson.readAddress(json, "$.L1ERC721BridgeProxy");
+
+        // Read superchain.yaml
+        string[] memory cmds = new string[](6);
+
+        // Build ffi command string
+        cmds[0] = "cat";
+        cmds[1] = "superchain.yaml";
+        cmds[2] = "|";
+        cmds[3] = "yq";
```
tynes, line 287: We will need to add `yq` to the required deps in the `README` if we agree on using it

tynes:
This is looking good, i believe it still needs a justfile update such that it attempts to run the script in the `$PWD` but then falls back to the one in the root

tasks/eth/006-MCP-L1/SignFromJsonWithAssertions.s.sol
```diff
@@ -0,0 +1,296 @@
...
+
+        require(resourceConfig.maxResourceLimit == rconfig.maxResourceLimit, "900");
+        require(resourceConfig.elasticityMultiplier == rconfig.elasticityMultiplier, "1000");
+        require(resourceConfig.baseFeeMaxChangeDenominator == rconfig.baseFeeMaxChangeDenominator, "1100");
+        require(resourceConfig.systemTxMaxGas == rconfig.systemTxMaxGas, "1200");
+        require(resourceConfig.minimumBaseFee == rconfig.minimumBaseFee, "1300");
+        require(resourceConfig.maximumBaseFee == rconfig.maximumBaseFee, "1400");
+        // Depends on start block being set to 0 in `initialize`
+        uint256 cfgStartBlock = cfg.systemConfigStartBlock();
+        require(config.startBlock() == (cfgStartBlock == 0 ? block.number : cfgStartBlock), "1500");
```
tynes, line 84: We should expect it to be set to an exact number here, we can remove the ternary

tasks/eth/006-MCP-L1/SignFromJsonWithAssertions.s.sol
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    /// @notice The name of the script
+    function name() public pure override returns (string memory) {
+        return "SignFromJsonWithAssertions";
+    }
+
+    /// @notice Asserts that the SystemConfig is setup correctly
+    function checkSystemConfig() internal view {
+        console.log("Running chain assertions on the SystemConfig");
```
tynes, line 59: A lot of these checks are copy pasted, can we use the assertions library that the deploy script uses to deduplicate things?

tasks/eth/006-MCP-L1/SignFromJsonWithAssertions.s.sol
```diff
@@ -0,0 +1,296 @@
...
+
+        require(superchainConfig.guardian() == cfg.superchainConfigGuardian(), "7200");
+        require(superchainConfig.paused() == false, "7300");
+    }
+
+    /// @notice Checks the correctness of the deployment
+    function _postCheck() internal view override {
+        console.log("Running post-deploy assertions");
+        ResourceMetering.ResourceConfig memory actualResourceConfig =
+            SystemConfig(proxies.SystemConfig).resourceConfig();
```
tynes, line 243: Why is the resource config checked here instead of inside of `checkSystemConfig`?

tasks/eth/006-MCP-L1/SignFromJsonWithAssertions.s.sol
```diff
@@ -0,0 +1,296 @@
...
+        ResourceMetering.ResourceConfig memory resourceConfig = config.resourceConfig();
+
+        require(config.owner() == cfg.finalSystemOwner(), "300");
+        require(config.overhead() == cfg.gasPriceOracleOverhead(), "400");
+        require(config.scalar() == cfg.gasPriceOracleScalar(), "500");
+        require(config.batcherHash() == bytes32(uint256(uint160(cfg.batchSenderAddress()))), "600");
+        require(config.gasLimit() == uint64(cfg.l2GenesisBlockGasLimit()), "700");
+        require(config.unsafeBlockSigner() == cfg.p2pSequencerAddress(), "800");
+        // Check _config
+        ResourceMetering.ResourceConfig memory rconfig = Constants.DEFAULT_RESOURCE_CONFIG();
```
tynes, line 74: We could hash the config here like the other check instead of checking individual properties, gives guarantee of exhaustiveness

tasks/eth/006-MCP-L1/postCheck.s.sol
0xfuturistic: thoughts @tynes @mds1?

mds1: Hardcoding and removing the dependency on DeployConfig seems reasonable to me. I'd recommend hardcoding them by defining constants, like `uint256 constant expectedSubmissionInterval = 1800`

script/SignFromJson.s.sol
mds1: I see we have these assertions on `proxies` members, but not on e.g. the address resulting from `superchainConfig.guardian()` calls

tasks/eth/006-MCP-L1/postCheck.s.sol
mds1: I still find some of the existing naming confusing, for example if this assertion fails:
```solidity
require(config.owner() == cfg.finalSystemOwner(), "300");
```
which is expected and which is actual? Both variables are the same name, one is just `config` abbreviated as `cfg`  😅

tasks/eth/006-MCP-L1/SignFromJsonWithAssertions.s.sol
mds1: @0xfuturistic let me know what the fix ended up being here and how I can test out that failures are properly detected

tasks/eth/006-MCP-L1/SignFromJsonWithAssertions.s.sol
```diff
@@ -0,0 +1,282 @@
...
+    function checkProtocolVersions() internal view {
+        console.log("Running chain assertions on the ProtocolVersions");
+
+        require(proxies.ProtocolVersions.code.length != 0, "6700");
+
+        ProtocolVersions versions = ProtocolVersions(proxies.ProtocolVersions);
+
+        require(versions.owner() == cfg.finalSystemOwner(), "6800");
+
+        string memory json;
```
mds1, line 199: This shadows a global `json` variable from `JsonTxBuilderBase`, let's rename this + the below references to something more descriptive to avoid possible bugs from shadowing

tasks/eth/006-MCP-L1/SignFromJsonWithAssertions.s.sol
```diff
@@ -0,0 +1,282 @@
...
+        l2OutputOracleStartingTimestamp = cfg.l2OutputOracleStartingTimestamp();
+
+        // Read the slot number for the xDomainMsgSender in the L1CrossDomainMessenger
+        try vm.readFile(
+            string.concat(
+                vm.projectRoot(),
+                "/lib/optimism/packages/contracts-bedrock/snapshots/storageLayout/L1CrossDomainMessenger.json"
+            )
+        ) returns (string memory data) {
+            xdmSenderSlotNumber = stdJson.readUint(data, string(abi.encodePacked("$.[13].slot")));
```
mds1, line 44: nit, should be able to clean this up in all three places
```suggestion
            xdmSenderSlotNumber = stdJson.readUint(data, "$.[13].slot");
```

tasks/eth/006-MCP-L1/SignFromJsonWithAssertions.s.sol
mds1: Sorry @tynes what do you mean here? Want to make sure we're on the same page, ref https://github.com/ethereum-optimism/superchain-ops/pull/119#discussion_r1548622313

0xfuturistic: I think we means that we should use the required and recommended protocol version values from DeployConfig rather than from `/tasks/eth/005-protocol-versions-ecotone/input.json"`

tasks/eth/006-MCP-L1/SignFromJsonWithAssertions.s.sol
0xfuturistic: ok I added an assert to the forge script that reads the value from the snapshot and checks whether the correct slot was used. this means that if the wrong slot had been used, CI should fail bc the simulation would fail. previously, we didn't have this. do you think this would be good enough? seems that if we wanted to make an individual CI check for this, it'd make more sense to do it in the monorepo, since otherwise that would mean testing a dependency here (which should test that on its own)

single.just
```diff
@@ -12,6 +12,11 @@ export foundationOwner0 := "0x42d27eEA1AD6e22Af6284F609847CB3Cd56B9c64"
 simulate hdPath='0':
   #!/usr/bin/env bash
   bundlePath=${taskPath}/${bundleName}.json
+  script=SignFromJson
+  if [ -f tasks/eth/006-MCP-L1/SignFromJsonWithAssertions.s.sol ]; then
```
tynes, line 16: This won't work for all possible tasks

We could do something like:

```bash
if [ -f "$PWD/SignFromJson" ]; then
  script="$PWD/SignFromJson"
fi
```

tynes: This would enforce the requirement that the script is always called `SignFromJson` whether its the root or the task specific script

tynes:
Adding `yq` to `ci-builder` here: https://github.com/ethereum-optimism/optimism/pull/10033

We can remove the install once we get a new release of `ci-builder`

tasks/eth/006-MCP-L1/SignFromJsonWithAssertions.s.sol
```diff
@@ -0,0 +1,353 @@
...
+        _proxies.SystemConfig = stdJson.readAddress(addressesJson, "$.SystemConfigProxy");
+        _proxies.L1ERC721Bridge = stdJson.readAddress(addressesJson, "$.L1ERC721BridgeProxy");
+
+        // Get go path
+        string[] memory inputs = new string[](3);
+        inputs[0] = "go";
+        inputs[1] = "env";
+        inputs[2] = "GOPATH";
+        string memory goPath = string(vm.ffi(inputs));
+
```
tynes, line 340: Hmm we should't couple this to the go path, it seems like you should add the go path to your `$PATH` in your `.bashrc`

tynes:
This is really close, good work so far

single.just
0xfuturistic: good idea!

tasks/eth/006-MCP-L1/SignFromJson.s.sol
```diff
@@ -0,0 +1,346 @@
...
+
+    uint256 xdmSenderSlotNumber;
+
+    /// @notice Sets up the contract
+    function setUp() public {
+        proxies = _getContractSet();
+
+        // Read the slot number for the xDomainMsgSender in the L1CrossDomainMessenger
+        try vm.readFile(
+            string.concat(
```
tynes, line 67: How do we feel about adding the monorepo as a dep here? cc @mds1 @maurelian 

Wondering if we could just include a hardcoded 204 and then a link to the code to validate. While reading it directly is really nice, its a super heavy solution

tasks/eth/006-MCP-L1/SignFromJson.s.sol
```diff
@@ -0,0 +1,346 @@
...
+
+    /// @notice Asserts the OptimismPortal is setup correctly
+    function checkOptimismPortal() internal view {
+        console.log("Running chain assertions on the OptimismPortal");
+
+        require(proxies.OptimismPortal.code.length != 0, "5700");
+
+        OptimismPortal portalToCheck = OptimismPortal(payable(proxies.OptimismPortal));
+
+        address guardian = superchainConfigGuardian;
```
tynes, line 245: smol style nit: doing extra assignment here while not doing this in other functions. Not the end of the world, just calling out how consistent style is important

tasks/eth/006-MCP-L1/SignFromJson.s.sol
```diff
@@ -0,0 +1,346 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.15;
+
+import {SignFromJson as OriginalSignFromJson} from "../../../script/SignFromJson.s.sol";
```
tynes, line 4: small nit: no need to give relative path, can always do a path relative to project root, ie `src/script/SignFromJson.s.sol". This makes it easier to refactor things, no need to use `../...` and change the import path during refactor

tasks/eth/006-MCP-L1/SignFromJson.s.sol
```diff
@@ -0,0 +1,346 @@
...
+import {L1CrossDomainMessenger} from "@eth-optimism-bedrock/src/L1/L1CrossDomainMessenger.sol";
+import {OptimismMintableERC20Factory} from "@eth-optimism-bedrock/src/universal/OptimismMintableERC20Factory.sol";
+import {L1ERC721Bridge} from "@eth-optimism-bedrock/src/L1/L1ERC721Bridge.sol";
+import {Predeploys} from "@eth-optimism-bedrock/src/libraries/Predeploys.sol";
+import {Types} from "@eth-optimism-bedrock/scripts/Types.sol";
+import {console2 as console} from "forge-std/console2.sol";
+import {stdJson} from "forge-std/StdJson.sol";
+
+contract SignFromJson is OriginalSignFromJson {
+    address constant finalSystemOwner = 0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A;
```
tynes, line 21: Could we add comments above each of these with a link to its source of truth?

tynes: This would be a permalink to the mainnet deploy config. For the guardian, final system owner, batch inbox, batch sender, proposer and challenger it could be good to have a link to etherscan

The idea is that we want to reduce the cognitive overhead of a reviewer so the chances of a bad code review happening are small

tasks/eth/006-MCP-L1/SignFromJson.s.sol
tynes: I lean towards removing the dep and just using a hyperlink to allow for easy validation. Would greatly simplify things

tynes:
I see that there are codesize checks on the superchain config proxy address. This would catch the previous issue introduced where a testnet address was used where there is no code at the address

tynes:
Looks like there is a conflict with the ci config file, a rebase can be used to fix it

.circleci/config.yml
```diff
@@ -157,6 +157,25 @@ jobs:
...
+          command: |
+            just install
+            forge clean
+            forge build
+            cd tasks/eth/006-MCP-L1
+            export SIMULATE_WITHOUT_LEDGER=1
+            just \
+              --dotenv-path .env \
+              --justfile ../../../single.just \
+              simulate
```
mds1, line 177: Since this is a contract upgrade, the SC is involved, therefore the nested 2/2 safe is used and we need to specify who we are simulating with
```suggestion
              --justfile ../../../nested.just \
              simulate \
              council
```

tasks/eth/006-MCP-L1/SignFromJsonWithAssertions.s.sol
mds1: Sorry, I was mainly referring to the fact that I'd expect this assertion to fail if we have the wrong `xdmSenderSlot` value, and therefore the `just simulate` command should exit with an error.
https://github.com/ethereum-optimism/superchain-ops/blob/6aeba81d59819b0680e649519b68ecb089e41582/tasks/eth/006-MCP-L1/SignFromJson.s.sol#L154

But if I hardcode the wrong slot number into the=is line:
https://github.com/ethereum-optimism/superchain-ops/blob/6aeba81d59819b0680e649519b68ecb089e41582/tasks/eth/006-MCP-L1/SignFromJson.s.sol#L153

And run:
```
just \
  --dotenv-path .env \
  --justfile ../../../single.just \
  simulate \
  council
```

The simulation still succeeds.

This is because we do not call `_postCheck()` anywhere as far as I can tell 😅

This should be called from both public methods in the `script/SignFromJson.s.sol` file

tasks/eth/006-MCP-L1/SignFromJson.s.sol
mds1: I'm ok with hardcoding and using a link to validate to keep it lighter weight, as if we have the wrong slot the assertion should fail. Here is the thread that led to using it as a dep, for context: https://github.com/ethereum-optimism/superchain-ops/pull/119#discussion_r1548249035

tasks/eth/006-MCP-L1/SignFromJson.s.sol
mds1: +1 for absolute paths

tasks/eth/006-MCP-L1/SignFromJson.s.sol
mds1: Agreed, I'd suggest preferring links to the superchain registry when applicable, and falling back to the deploy config or etherscan for data not in the registry

tasks/eth/006-MCP-L1/SignFromJson.s.sol
```diff
@@ -0,0 +1,346 @@
...
+        require(address(bridgeToCheck.OTHER_BRIDGE()) == Predeploys.L2_ERC721_BRIDGE, "5200");
+        require(address(bridgeToCheck.otherBridge()) == Predeploys.L2_ERC721_BRIDGE, "5300");
+        require(address(bridgeToCheck.otherBridge()).code.length != 0, "5301");
+
+        require(address(bridgeToCheck.MESSENGER()) == proxies.L1CrossDomainMessenger, "5400");
+        require(address(bridgeToCheck.messenger()) == proxies.L1CrossDomainMessenger, "5500");
+        require(address(bridgeToCheck.messenger()).code.length != 0, "5502");
+
+        require(address(bridgeToCheck.superchainConfig()) == proxies.SuperchainConfig, "5600");
+        require(address(bridgeToCheck.superchainConfig()).code.length != 0, "5603");
```
mds1, line 234: From @tynes

> I see that there are codesize checks on the superchain config proxy address. This would catch the previous issue introduced where a testnet address was used where there is no code at the address

We would actually need codesize checks on the implementation address to catch this. Once we have that, we should run this version of the script against the old (wrong) `input.json` to sanity check that the mistake is caught. Currently it is not caught using the attached old `input.json`, but this is because we do not yet have the `_postCheck` actually being used (see https://github.com/ethereum-optimism/superchain-ops/pull/119#discussion_r1551820033)

[input.json](https://github.com/ethereum-optimism/superchain-ops/files/14873135/input.json)

tasks/eth/006-MCP-L1/SignFromJson.s.sol
0xfuturistic: for links in the monorepo, I'm using the MCP commit as the version for the files

tasks/eth/006-MCP-L1/SignFromJsonWithAssertions.s.sol
0xfuturistic: looks like when adding `_postCheck() `to function `signJson`, the call fails. for `approveJson` and `runJson` everything goes smoothly (everything in file `script/NestedSignFromJson.s.sol`)

0xfuturistic:
per @mds1's comments, migrated everything to `nested.just`, which includes looking for a local `NestedSignFromJson.s.sol` in the task. still, I decided to leave the mechanism in `single.just` to use local `SignFromJson.s.sol` if it exists because i think it may be useful in the future (it also makes sense to preserve this symmetry with `nested.just`)

tasks/eth/006-MCP-L1/NestedSignFromJson.s.sol
```diff
@@ -0,0 +1,387 @@
...
+
+        SystemConfig configToCheck = SystemConfig(proxies.SystemConfig);
+
+        ResourceMetering.ResourceConfig memory resourceConfigToCheck = configToCheck.resourceConfig();
+
+        require(configToCheck.owner() == superchainConfigGuardian, "300");
+
+        require(configToCheck.overhead() == gasPriceOracleOverhead, "400");
+
+        require(configToCheck.scalar() == gasPriceOracleScalar, "500");
```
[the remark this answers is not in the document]
tynes: To be honest all of the hardcoded values could be populated this way. Curious how you feel about this @mds1 ?

tynes:
We are going to want to add in installing `yq` as part of CI again until we have another release of `ci-builder` that includes it

tasks/eth/006-MCP-L1/NestedSignFromJson.s.sol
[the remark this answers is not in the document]
tynes: I don't see a diff with the scalar on the tenderly link in CI so this isn't an issue

tynes:
Having CI print the tenderly link is super useful. Great job to security team for building this!

tasks/eth/006-MCP-L1/SignFromJsonWithAssertions.s.sol
mds1: Do you need help debugging this?

README.md
```diff
@@ -70,6 +70,15 @@ We’ll use the [eip712sign](https://github.com/base-org/eip712sign) utility dev
 You should see a message something like this:
   ` One (and only one) of --private-key, --ledger, --mnemonic must be set`
 
+### Installing `mikefarah/yq`
+
+We’ll use `mikefarah/yq` to manipulate `.yaml` files:
```
mds1, line 75: Nit
```suggestion
We’ll use `mikefarah/yq` to parse `.yaml` files:
```

script/DeployRehearsalContracts.s.sol
```diff
@@ -30,9 +30,6 @@ contract DeployRehearsalContracts is Deployer {
         council_safe = GnosisSafe(payable(vm.envAddress("COUNCIL_SAFE")));
 
         require(owner_safe.isOwner(address(council_safe)));
-
-        console.log("Deploying from %s", deployScript);
-        console.log("Deployment context: %s", deploymentContext);
```
mds1, line 35: What were these logs for originally and why were they removed?

mds1: cc @zchn who committed these logs, to confirm these were for debugging and not used by signers?

tasks/eth/006-MCP-L1/NestedSignFromJson.s.sol
```diff
@@ -0,0 +1,387 @@
...
+import {OptimismMintableERC20Factory} from "@eth-optimism-bedrock/src/universal/OptimismMintableERC20Factory.sol";
+import {L1ERC721Bridge} from "@eth-optimism-bedrock/src/L1/L1ERC721Bridge.sol";
+import {Predeploys} from "@eth-optimism-bedrock/src/libraries/Predeploys.sol";
+import {Types} from "@eth-optimism-bedrock/scripts/Types.sol";
+import {EIP1967Helper} from "@eth-optimism-bedrock/test/mocks/EIP1967Helper.sol";
+import {console2 as console} from "forge-std/console2.sol";
+import {stdJson} from "forge-std/StdJson.sol";
+
+contract NestedSignFromJson is OriginalNestedSignFromJson {
+    /// @notice Verify against https://etherscan.io/address/0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A
```
mds1, line 22: Etherscan links should be a last resort IMO, e.g. looking at this address on etherscan just tells me it's a contract. It doesn't tell me that it's the expected value. I'm not sure what `finalSystemOwner` actually corresponds to based on the roles in https://docs.optimism.io/chain/security/privileged-roles, it's not a variable name in the contracts anywhere afaik

tasks/eth/006-MCP-L1/NestedSignFromJson.s.sol
```diff
@@ -0,0 +1,387 @@
...
+    address constant finalSystemOwner = 0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A;
+
+    /// @notice Verify against https://etherscan.io/address/0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A
+    address constant superchainConfigGuardian = 0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A;
```
mds1, line 26:
```suggestion
    /// @notice Verify against https://docs.optimism.io/chain/security/privileged-roles#guardian
    address constant superchainConfigGuardian = 0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A;
```

tasks/eth/006-MCP-L1/NestedSignFromJson.s.sol
```diff
@@ -0,0 +1,387 @@
...
+
+    /// @notice Verify against https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/deploy-config/mainnet.json#L7
```
mds1, line 28: We should use a permalink to a specific commit, not to `develop`, since this link might change or break otherwise

tasks/eth/006-MCP-L1/NestedSignFromJson.s.sol
```diff
@@ -0,0 +1,387 @@
...
+    uint256 constant l2BlockTime = 2;
+
+    /// @notice Verify against https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/deploy-config/mainnet.json#L12
+    address constant p2pSequencerAddress = 0xAAAA45d9549EDA09E70937013520214382Ffc4A2;
+
+    /// @notice Verify against https://etherscan.io/address/0xFF00000000000000000000000000000000000010
+    address constant batchInboxAddress = 0xFF00000000000000000000000000000000000010;
+
+    /// @notice Verify against https://etherscan.io/address/0x6887246668a3b87F54DeB3b94Ba47a6f63F32985
+    address constant batchSenderAddress = 0x6887246668a3b87F54DeB3b94Ba47a6f63F32985;
```
mds1, line 38:
```suggestion
    /// @notice Verify against https://docs.optimism.io/chain/security/privileged-roles#batcher
    address constant batchSenderAddress = 0x6887246668a3b87F54DeB3b94Ba47a6f63F32985;
```

tasks/eth/006-MCP-L1/NestedSignFromJson.s.sol
```diff
@@ -0,0 +1,387 @@
...
+    uint256 constant l2OutputOracleSubmissionInterval = 1800;
+
+    /// @notice Verify against https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/deploy-config/mainnet.json#L16
+    uint256 constant l2OutputOracleStartingTimestamp = 1686068903;
+
+    /// @notice Verify against https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/deploy-config/mainnet.json#L17
+    uint256 constant l2OutputOracleStartingBlockNumber = 105235063;
+
+    /// @notice Verify against https://etherscan.io/address/0x473300df21D047806A082244b417f96b32f13A33
+    address constant l2OutputOracleProposer = 0x473300df21D047806A082244b417f96b32f13A33;
```
mds1, line 50:
```suggestion
    /// @notice https://docs.optimism.io/chain/security/privileged-roles#proposer
    address constant l2OutputOracleProposer = 0x473300df21D047806A082244b417f96b32f13A33;
```

script/DeployRehearsalContracts.s.sol
mds1: Thanks! I'd advocate for keeping it in then unless there was a reason @0xfuturistic chose to remove it

0xfuturistic: exactly. upon updating `lib/optimism`, the vars are missing so the lines give an error

tasks/eth/006-MCP-L1/NestedSignFromJson.s.sol
0xfuturistic: good idea!

tasks/eth/006-MCP-L1/NestedSignFromJson.s.sol
0xfuturistic: does it correspond to [`System Config Owner`](https://docs.optimism.io/chain/security/privileged-roles#system-config-owner)? addresses would line up as well (cc @tynes)

tasks/eth/006-MCP-L1/SignFromJsonWithAssertions.s.sol
0xfuturistic: if you have a minute, we could parallelize working on this. do you think you could take a look?

script/DeployRehearsalContracts.s.sol
0xfuturistic: what does it make sense to use in replacement for `deployScript`? thought about using `Config.sig()`, but it defaults to string `string("run")`, which is not very descriptive @zchn

tasks/eth/006-MCP-L1/SignFromJsonWithAssertions.s.sol
mds1: I saw the error of `revert: 500` which is from this line:
```solidity
require(configToCheck.scalar() == gasPriceOracleScalar, "500");
```

Logging those two values we see:
```
configToCheck.scalar(): 452312848583266388373324160190187140051835877600158453279134670530344387928
 gasPriceOracleScalar:   0
```

I believe this is because the deploy config's value of 0 is not updated to support the new ecotone encoding. I think it should be safe to hardcode the expected value here to `0x10000000000000000000000000000000000000000000000000c5fc500000558`, which is the hex representation of that large decimal number.

I then commented out this one `require` statement and now it looks like SystemConfig calls are reverting, e.g. `configToCheck.startBlock()` is the next call to fail. We'll need to look into that next

tasks/eth/006-MCP-L1/NestedSignFromJson.s.sol
tynes: Yes it does!

tynes:
I don't see the following log printed:
 `console.log("Running post-deploy assertions");`

Are we sure the post check is running in CI?

0xfuturistic:
> I don't see the following log printed: `console.log("Running post-deploy assertions");`
> 
> Are we sure the post check is running in CI?

@mds1 is working on adding `_postCheck()` [here](https://github.com/ethereum-optimism/superchain-ops/pull/119#discussion_r1554205715)

nested.just
```diff
@@ -48,6 +53,11 @@ simulate whichSafe hdPath='0':
 sign whichSafe hdPath='0':
   #!/usr/bin/env bash
   bundlePath=${taskPath}/${bundleName}.json
+  script=NestedSignFromJson
+  if [ -f "${taskPath}/NestedSignFromJson.s.sol" ]; then
```
tynes, line 57: I don't think this line is working correctly.

 Can you add an echo here so we are sure its hitting the script?

```bash
echo "Running script with assertions"
```

tynes: I don't think `taskPath` is correct, i think you want `$PWD`

tasks/eth/006-MCP-L1/SignFromJsonWithAssertions.s.sol
0xfuturistic: could it be that the script is checking against the current version of `SystemConfig`, which doesn't include a method for `startBlock()`?

tynes:
The failure is

```
    ├─ [672] 0x229047fed2591dbec1eF1118d64F7aF3dB9EB290::startBlock() [staticcall]
```

We should double check the implementation of the system config that is upgraded to has `startBlock` in its abi

script/NestedSignFromJson.s.sol
```diff
@@ -8,25 +8,30 @@ import {stdJson} from "forge-std/StdJson.sol";
...
+        globalSignerSafe = _signerSafe;
+        _postCheckWithSim();
     }
 
     /// @dev Executes the transaction from the System Owner Safe.
     function runJson(string memory _path) public {
         _loadJson(_path);
         run();
-        _postCheck();
+        // globalSignerSafe = _signerSafe; // TODO how to handle this one?
```
tynes, line 33: Could `vm.env` work?

tynes (approved):

script/NestedSignFromJson.s.sol
mds1: Good idea. Since this code path isn't run until actual execution of the signed transaction, let's merge this and I'll look into this

mds1 (approved):

PR #130, opened 2024-04-10
feat: clean up simulation code and fix bug on address checks

mds1:
This is in draft until https://github.com/base-org/contracts/pull/72 gets merged. At that point we can update our `base-org/contracts` submodule commit on this PR. Until then, this will not compile, but you can test it locally by checkout out this PR, `cd lib/base-contracts`, then checking out the base commit from that PR

Changes:
- No more hacky simulation approach since it's natively handled by the base contracts repo.
- Fixes bug in address check that ensures all addresses written to storage have code.
- Forces implementer to implement the `_postCheck` method by having a default that reverts. (Note that we use `require(false, "messsage")` instead of `revert("message")` because with the latter solidity is smart enough to detect and warn about unreachable code, so using the former prevents that).


mds1:
Closing in favor of https://github.com/ethereum-optimism/superchain-ops/pull/133

Since MCP L1 was already executed, we should avoid changing the scripts that were used, but we will start using this pattern in future playbooks

PR #141, opened 2024-04-16
why won't it break?

maurelian:
forge build is just not picking up the solidity files in the tasks dir.

This PR modifies `tasks/eth/006-MCP-L1/NestedSignFromJson.s.sol` so that it is:
- not valid solidity (CI should fail)
- renames the file and contract (with an appended `X`) in case collision with `scripts/NestedSignFromJson.s.sol` is the issue

Things I have tried locally:

- `forge clean && forge build`
- `forge build --force`

But I am not seeing any artifacts in `out/NestedSignFromJsonX.s.sol`.


mds1:
Closing this since the issue was resolved. Summary was:
- This task is removed from CI which is why CI passed
- The justfile looks for `NestedSignFromJson` as the name but the file was renamed to `NestedSignFromJsonX`

PR #146, opened 2024-04-17
Move state diff checking to JsonTxBuilderBase

maurelian:
**Description**

Reduces code reuse in the task specific script by moving `checkStateDiff` to `JsonTxBuilderBase` which is shared by all the relevant scripts.

### DIFF
--- a/src/JsonTxBuilderBase.sol
+++ b/src/JsonTxBuilderBase.sol
@@ -130,7 +130,7 @@ abstract contract JsonTxBuilderBase is CommonBase {
     /// @notice Returns a list of addresses which are expected to be in storage, but will not to have code on this
     ///         chain. Examples of such addresses include EOAs, predeploy addresses, and inbox addresses.
     function getCodeExceptions() internal pure virtual returns (address[] memory) {
-        revert("getCodeExceptions not implemented");
+        require(false, "getCodeExceptions not implemented");
     }
 
     /// @notice Checks that values have code on this chain.
--- a/src/JsonTxBuilderBase.sol
+++ b/src/JsonTxBuilderBase.sol
@@ -114,10 +114,10 @@ abstract contract JsonTxBuilderBase is CommonBase {
         }
     }
 
-    /// @notice Returns a list of addresses that are known to not have code on this chain.
+    /// @notice Returns a list of addresses which are expected to be in storage, but will not to have code on this
+    ///         chain. Examples of such addresses include EOAs, predeploy addresses, and inbox addresses.
     function getCodeExceptions() internal view virtual returns (address[] memory) {
-        console.log("\x1b[1;33mWARNING:\x1b[0m getCodeExceptions not implemented");
-        return new address[](0);
+        revert("getCodeExceptions not implemented");
     }
 
     /// @notice Checks that values have code on this chain.
--- a/src/JsonTxBuilderBase.sol
+++ b/src/JsonTxBuilderBase.sol
@@ -56,7 +56,7 @@ abstract contract JsonTxBuilderBase is CommonBase {
     /// @notice Reads all account and storage accesses and makes a series of basic sanity checks on them.
     ///         This function can be overridden to provide more specific checks, but should still call
     ///         super.checkStateDiff().
-    function checkStateDiff(Vm.AccountAccess[] memory accountAccesses) internal view {
+    function checkStateDiff(Vm.AccountAccess[] memory accountAccesses) internal view virtual {
         require(accountAccesses.length > 0, "No account accesses");
 
         for (uint256 i; i < accountAccesses.length; i++) {
@@ -123,7 +123,7 @@ abstract contract JsonTxBuilderBase is CommonBase {
     /// @notice Checks that values have code on this chain.
     ///         This method is not storage-layout-aware and therefore is not perfect. It may return erroneous
     ///         results for cases like packed slots, and silently show that things are okay when they are not.
-    function isLikelyAddressThatShouldHaveCode(uint256 value) internal view returns (bool) {
+    function isLikelyAddressThatShouldHaveCode(uint256 value) internal view virtual returns (bool) {
         // If out of range (fairly arbitrary lower bound), return false.
         if (value > type(uint160).max) return false;
         if (value < uint256(uint160(0x00000000fFFFffffffFfFfFFffFfFffFFFfFffff))) return false;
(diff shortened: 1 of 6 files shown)


maurelian:
CI failure is also occurring on `main`

tasks/sep/metal-001-MCP-L1/SignFromJson.s.sol
maurelian, line 1: I'm pretty sure that I could hav deleted both of these functions, but it felt safer and easier to review to simply have them override the virtuals in JsonTxBuilderBase

mds1: I'm supportive of deleting them and happy to diff/test the changes. We'll be using the metal folder as the template for other upgrades, so keeping it lean is preferable to reduce the amount of code copy/pasted

src/JsonTxBuilderBase.sol
```diff
@@ -51,4 +52,95 @@ abstract contract JsonTxBuilderBase is CommonBase {
...
+                );
+                require(
+                    !storageAccess.reverted,
+                    string.concat("Storage access reverted: ", vm.toString(storageAccess.account))
+                );
+            }
+        }
+    }
+
+    /// @notice Returns a list of addresses that are known to not have code on this chain.
```
mds1, line 117: Nit, can we a bit more explicit on this method's purpose? Mainly to prevent "oh yea this address does have no code on mainnet, ok let me add it here so things pass" without realizing what they're doing

src/JsonTxBuilderBase.sol
```diff
@@ -51,4 +52,95 @@ abstract contract JsonTxBuilderBase is CommonBase {
...
+    function getCodeExceptions() internal view virtual returns (address[] memory) {
+        console.log("\x1b[1;33mWARNING:\x1b[0m getCodeExceptions not implemented");
+        return new address[](0);
```
mds1, line 120: Let's revert here to force implementation. Previously I did warnings in the other `script/` files to avoid breaking changes with the rehersal sims in CI, but I'd prefer to be stricter and force implementations

src/JsonTxBuilderBase.sol
```diff
@@ -51,4 +52,95 @@ abstract contract JsonTxBuilderBase is CommonBase {
 
         return calls;
     }
+
+    /// @notice Reads all account and storage accesses and makes a series of basic sanity checks on them.
+    ///         This function can be overridden to provide more specific checks, but should still call
+    ///         super.checkStateDiff().
+    function checkStateDiff(Vm.AccountAccess[] memory accountAccesses) internal view virtual {
+        require(accountAccesses.length > 0, "No account accesses");
```
mds1, line 60: Any objections to adding this log back? It's nice to be able to read through the logs to make sure all expected assertions ran
```suggestion
        console.log("Running assertions on the state diff");
        require(accountAccesses.length > 0, "No account accesses");
```

src/JsonTxBuilderBase.sol
```diff
@@ -51,4 +52,95 @@ abstract contract JsonTxBuilderBase is CommonBase {
...
+                string.concat("Self-destructed account: ", vm.toString(accountAccess.account))
+            );
+
+            for (uint256 j; j < accountAccess.storageAccesses.length; j++) {
+                Vm.StorageAccess memory storageAccess = accountAccess.storageAccesses[j];
+
+                if (!storageAccess.isWrite) continue; // Skip SLOADs.
+
+                uint256 value = uint256(storageAccess.newValue);
+                if (isLikelyAddressThatShouldHaveCode(value)) {
```
mds1, line 83: Now that I think of it, we can make this check even stricter—we currently have a no-op if `isLikelyAddressThatShouldHaveCode(value)` returns false. Instead we can assert that there is no code, as we expect.

tasks/sep/metal-001-MCP-L1/SignFromJson.s.sol
```diff
@@ -366,87 +366,17 @@ contract SignFromJson is OriginalSignFromJson {
...
-        if (address(uint160(value)) == l2OutputOracleChallenger) return false;
-        if (address(uint160(value)) == systemConfigOwner) return false;
-        if (address(uint160(value)) == batchSenderAddress) return false;
-        if (address(uint160(value)) == p2pSequencerAddress) return false;
-        if (address(uint160(value)) == batchInboxAddress) return false;
-
-        // Otherwise, this value looks like an address that we'd expect to have code.
-        return true;
-    }
+    function getCodeExceptions() internal view override returns (address[] memory) {
```
mds1, line 369: solc warns that this can be pure

tasks/sep/mode-001-MCP-L1/SignFromJson.s.sol
```diff
@@ -366,87 +366,17 @@ contract SignFromJson is OriginalSignFromJson {
...
-        if (address(uint160(value)) == l2OutputOracleChallenger) return false;
-        if (address(uint160(value)) == systemConfigOwner) return false;
-        if (address(uint160(value)) == batchSenderAddress) return false;
-        if (address(uint160(value)) == p2pSequencerAddress) return false;
-        if (address(uint160(value)) == batchInboxAddress) return false;
-
-        // Otherwise, this value looks like an address that we'd expect to have code.
-        return true;
-    }
+    function getCodeExceptions() internal view override returns (address[] memory) {
```
mds1, line 369: also can be pure

src/JsonTxBuilderBase.sol
```diff
@@ -51,4 +52,108 @@ abstract contract JsonTxBuilderBase is CommonBase {
...
+                    string.concat("Storage access reverted: ", vm.toString(storageAccess.account))
+                );
+            }
+        }
+    }
+
+    /// @notice Returns a list of addresses which are expected to be in storage, but will not to have code on this
+    ///         chain. Examples of such addresses include EOAs, predeploy addresses, and inbox addresses.
+    function getCodeExceptions() internal pure virtual returns (address[] memory) {
+        revert("getCodeExceptions not implemented");
```
mds1, line 133: this will "trick" solidity into not emitting an unreachable code warning
```suggestion
        require(false, "getCodeExceptions not implemented");
```

tasks/sep/metal-001-MCP-L1/SignFromJson.s.sol
mds1: Diffed + tested and they lgtm, thank you!

mds1 (approved):
Looks great, thank you!

PR #158, opened 2024-04-18
tasks(eth/metal): add Metal Mainnet MCP L1 playbook

blmalone:
This PR adds the Metal **Mainnet** MCP L1 upgrade task.

Review instructions:
1. Follow this folder's instructions to execute playbook and make sure it does not revert
2. Verify:
    1. That the Validations file is exactly what you see on Tenderly
    1. That the Validations file is consistent with what we'd expect via our [notion analysis](https://www.notion.so/oplabs/Justification-for-Skipping-Upgrade-4-Extended-Pause-and-Going-Straight-from-Bedrock-to-Upgrade-6--d4b2f9bacc6f4e479e5f2ba3561ff586?pvs=4)
    1. That the hardcoded values in `./tasks/eth/metal-001-MCP-L1/SignFromJson.s.sol` are all accurate and sensible
3. Diff the new `tasks/eth/metal-001-MCP-L1/SignFromJson.s.sol` against the existing `tasks/eth/006-MCP-L1/SignFromJson.s.sol` to better see the changes, since most of the file is a copy/paste of that.

**For reference only:** 
Tenderly state diff: https://www.tdly.co/shared/simulation/2e016361-2793-4763-85e5-0c08ffd83717 (you still **must** compute your own)


tasks/eth/metal-001-MCP-L1/.env
```diff
@@ -0,0 +1,3 @@
+ETH_RPC_URL="https://ethereum.publicnode.com"
```
mds1, line 1: Let's use the tenderly URL again here

mds1 (approved):
LGTM, thank you!

PR #164, opened 2024-04-19
fix: skip precompile code check and silence compiler warnings

mds1:
cc @emilianobonassi, give this PR a try


Inphi (approved):
Might be worth changing forge recording behavior to ignore precompiles for Accesses. Since they're more like pure functions.

emilianobonassi:
it works

mds1:
> Might be worth changing forge recording behavior to ignore precompiles for Accesses. Since they're more like pure functions.

This is an interesting point, worth thinking about. I think the current behavior is ok though and I'd rather lean towards the cheat returning a more complete picture of the trace, than a less complete one

PR #179, opened 2024-04-25
feat: FPAC Sepolia Testnet Upgrade Mk. 2

clabby:
## Overview

Lays out the Sepolia testnet upgrade for FPAC, mk. 2

PR #186, opened 2024-04-30
Add Sepolia 007 Presigned Pause Ceremony 

maurelian:
**Description**

Adds a new presigned pause ceremony, based on the one in tasks/sep/005. 

A key change is that `SignFromInputJson` is renamed to `PresignPauseFromJson`, since it was only being used in the context of a presigning runbook. Post checks were also added to the script which make assertions on all storage changes that could occur when the Pause is executed.


maurelian:
[CI Failure](https://app.circleci.com/pipelines/github/ethereum-optimism/superchain-ops/990/workflows/278b1f60-4bb4-492b-8dd7-40dfaa69f6b7/jobs/8663?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary&invite=true#step-102-1229): 

The signature is `0x000000000000000000000000f39fd6e51aad88f6f4ce6ab8827279cfffb92266000000000000000000000000000000000000000000000000000000000000000001`, so it is a prevalidated signature from the standard `test test test...` mnemonic which is passed in [here](https://github.com/ethereum-optimism/superchain-ops/blob/ethi/debug/presigned-pause.just#L58). 

I think that this was passing tests in the previous runbooks because we weren't yet running checks when the `sign()` function is called. 

This is hard to fix, because the tool depends on having access to the private keys of the signer.

script/PresignPauseFromJson.s.sol
maurelian, line 1: Unfortunately the renaming + changes was enough to cause this to appear as a whole new file, but it's just a modification to `SignFromInputJson`. 

You can see the bulk of the changes prior to the rename in this commit: 
d3a5f2d8265d6ba6ea43c99f1e6e8dd38fc2c68d

maurelian: now this commit: a89222be6475b916e7b9c160e79d8ace4f6ceb83

.circleci/config.yml
```diff
@@ -155,6 +155,31 @@ jobs:
             just simulate-council
             just prepare-json
             just simulate-council # simulate again to make sure the json is still valid
+
+  just_simulate_sep_007-presigned-pause:
```
mds1, line 159: FP uses 007 in [this PR](https://github.com/ethereum-optimism/superchain-ops/pull/179), so this should probably be 008

tasks/sep/007-presigned-pause/.env
```diff
@@ -0,0 +1,5 @@
+ETH_RPC_URL="https://ethereum-sepolia.publicnode.com"
+DEPUTY_GUARDIAN_MODULE_ADDR=0xed12261735aD411A40Ea092FF4701a962d25cA21
+SUPERCHAIN_CONFIG_ADDR=0xC2Be75506d5724086DEB7245bd260Cc9753911Be
+OWNER_SAFE=0xDEe57160aAfCF04c34C887B5962D0a69676d3C8B
```
mds1, line 4: Can we rename `OWNER_SAFE` to something more clear? The proxy admin owner is 0x1Eb2fFc903729a0F03966B917003800b145F56E2, and this 0xDEe57160aAfCF04c34C887B5962D0a69676d3C8B address is a signer on that safe

presigned-pause.just
```diff
@@ -1,8 +1,8 @@
 set dotenv-load
 
 export rpcUrl := env_var('ETH_RPC_URL')
-export superchainConfigAddr := env_var('SUPERCHAIN_CONFIG_ADDR')
-export safeAddr := env_var('SAFE_ADDR')
+export superchainConfigAddr := env_var('DEPUTY_GUARDIAN_MODULE_ADDR')
```
mds1, line 4: also need to rename `superchainConfigAddr` when it's used of course
```suggestion
export deputyGardianModuleAddr := env_var('DEPUTY_GUARDIAN_MODULE_ADDR')
```

presigned-pause.just
```diff
@@ -1,8 +1,8 @@
...
+export safeAddr := env_var('OWNER_SAFE')
```
mds1, line 5: same comment on variable name since there are 3 safes, should be more clear which one is this supposed to be

presigned-pause.just
```diff
@@ -42,20 +42,22 @@ sign hdPath='0':
...
-          --script-name SignFromInputJson \
+          --script-name PresignPauseFromJson \
           --rpc-url ${rpcUrl} \
           --json-file $JSON_FILE \
           --ledger \
           --hd-paths "m/44'/60'/{{hdPath}}'/0/0" \
           sign
       else
+      TEST_MNEMONIC="test test test test test test test test test test test junk"
+      export TEST_SENDER=0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266
```
mds1, line 53: Why do we hardcode a sender, and why export it as an env var?

maurelian: We only take this path if `SIMULATE_WITHOUT_LEDGER=true`, otherwise we do use the account the signer is signing with. 

I believe that the reason we do this rather than just getting one of the signers from the Safe like in other runbooks is because we need the private key.

maurelian: Also noting that this code needs an indenting fix.

script/PresignPauseFromJson.s.sol
```diff
@@ -0,0 +1,73 @@
...
+        for (uint256 i; i < accesses.length; i++) {
+            Vm.AccountAccess memory accountAccess = accesses[i];
+
+            for (uint256 j; j < accountAccess.storageAccesses.length; j++) {
+                Vm.StorageAccess memory storageAccess = accountAccess.storageAccesses[j];
+
+                if (storageAccess.isWrite) {
+                    console.log("Checking Owner Safe storage writes");
+                    if (storageAccess.account == vm.envAddress("OWNER_SAFE")) {
+                        require(storageAccess.slot == bytes32(uint256(5)), "The only allowed write is to the nonce");
```
mds1, line 52: Why we not also run `checkStateDiff`?

tasks/sep/007-presigned-pause/.env
maurelian: Using PRESIGNER_SAFE for now.

script/PresignPauseFromJson.s.sol
```diff
@@ -0,0 +1,73 @@
...
+import {console} from "forge-std/console.sol";
+import {Vm} from "forge-std/Vm.sol";
+
+/// @title PresignPauseFromJson
+/// @notice A script that reads a JSON file and builds a series of transactions from it. This script is
+///     intended to be used only in the presigned pause runbooks.
+contract PresignPauseFromJson is MultisigBuilder, JsonTxBuilderBase {
+    function _addGenericOverrides() internal view override returns (SimulationStateOverride memory override_) {
+        // If SIMULATE_WITHOUT_LEDGER is set, we add an override to allow the script to run using the same
+        // test address as defined in presigned-pause.just
```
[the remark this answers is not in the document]
maurelian: Added some more verbiage: 

```
        // If SIMULATE_WITHOUT_LEDGER is set, we add an override to allow the script to run using the same
        // test address as defined in presigned-pause.just. This is necessary because the presigner tool requires
        // access to the private key of the address that will sign the transaction. Therefore we must insert a test
        // address into the owners list.
```

mds1 (approved):

PR #190, opened 2024-05-02
fp: Sepolia bond recovery

Inphi:
Add a task for the Guardian to recover funds from the `DelayedWETH` contract.


tasks/sep/fp-recovery/004-recover-bonds/justfile
```diff
@@ -0,0 +1,18 @@
+set positional-arguments
+
+# default recipe to display help information
+default:
+  @just --list
+
+# Generate the `input.json` with a DelayedWETH address to recover bonds
+generate-input delayed_weth:
+  #!/usr/bin/env bash
+  AMOUNT=$(cast --to-hex $(jq add ./meta.json))
```
clabby, line 10: sanity check -- what size ints does `jq add` use? Could this cease to work if we're using much larger ints, say full fixed-point 1e18 WEI?

tasks/sep/fp-recovery/004-recover-bonds/input-template.json
```diff
@@ -0,0 +1,33 @@
...
+    "name": "Guardian - Recover all funds in the DelayedWETH contract",
+    "description": "This task calls the `recover` function in DelayedWETH to retrieve the funds locked in the DelayedWETH contract. This task should be executed when the Superchain is paused to avoid ongoing games from funding the contract"
+  },
+  "transactions": [
+    {
+      "metadata": {
+        "name": "DelayedWETH.recover",
+        "description": "Recover all funds in DelayedWETH"
+      },
+      "to": "0xca36b00f885d83068f0cffee381f0c974de1d58e",
```
clabby, line 13: Looks to be an out of date `DelayedWETH` deployment: https://github.com/ethereum-optimism/superchain-registry/blob/main/superchain/extra/addresses/sepolia/op.json#L14

tasks/sep/fp-recovery/004-recover-bonds/input-template.json
```diff
@@ -0,0 +1,33 @@
...
+      "value": "0x0",
+      "data": "0x0ca35682ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff",
+      "contractMethod": {
+        "type": "function",
+        "name": "recover",
```
clabby, line 18: qq - If we're recovering funds, wouldn't we theoretically want to recover _all_ funds? Is it worth making this more granular?

clabby (approved):

PR #200, opened 2024-05-15
feat: Sepolia `op-contracts/v1.4.0-rc.3` upgrade

clabby:
## Overview

Upgrades OP Sepolia to `op-contracts/v1.4.0-rc.3`


ajsutton:
Are the changes to the submodules deliberate? (Just checking because its so easy to accidentally change them)

clabby:
> Are the changes to the submodules deliberate? (Just checking because its so easy to accidentally change them)

Yep - deliberate. Pulled in changes to support using the dispute contracts in the `NestedSignFromJson` script so that we can make post-checks.

lib/optimism
maurelian, line 1: is this a tagged commit?

tasks/sep/009-fpac-upgrade/.env
maurelian, line 1: These addresses are correct.

maurelian:
Needs a new `just simulate` job added to CI.

Inphi:
> Needs a new `just simulate` job added to CI.

@maurelian  Note that unlike SC actions, this isn't a task that will always work across upgrades. Every time we upgrade FPs on sepolia, we'll need to update the CI task. 
There aren't any CI tasks for other sepolia operations. Any particular reason why we need one for this op-contracts upgrade?

PR #201, opened 2024-05-15
Adding opsep as a task directory

blmalone:
Upgrades for op-sepolia should go through normal processes outlined in this repo.


tasks/README.md
```diff
@@ -0,0 +1,3 @@
+# Tasks
+
+The name of each directory corresponds to that chains official 'short name' which can be found in [ethereum-lists/chains](https://github.com/ethereum-lists/chains/blob/37756cd15620996b6ed99bce6e95924151ad24da/_data/chains/eip155-11155420.json#L16).
```
mds1, line 3: nit + link fix
```suggestion
The name of each directory corresponds to that chains official 'short name' which can be found in the [ethereum-lists/chains](https://github.com/ethereum-lists/chains) repo.
```

tasks/op-sep/README.md
mds1, line 1: also I think this should be `tasks/opsep` instead of `tasks/op-sep` since the shortname is `opsep`

blmalone: oh good catch!

mds1 (approved):

PR #205, opened 2024-05-22
sep: Upgrading L2 Predeploy From L1 Proxy Admin Owner

blmalone:
**Description**

Performing an L2 predeploy upgrade using the L1 Proxy Admin Owner. The exact upgrade will be a no-op, meaning we will upgrade a predeploy to the same implementation address it is already pointing to. This approach will allow us to confirm that predeploy upgrades on L2 work using a transaction originating from L1.
Ultimately, having the security council have veto power over L2 predeploy upgrades is a requirement for reaching Stage 1. The security council is a signer on the L1 Proxy Admin Owner.

The no-op is upgrading the L2ERC721Bridge [0x4200000000000000000000000000000000000014](https://sepolia-optimism.etherscan.io/address/0x4200000000000000000000000000000000000014) to it's current implementation code [0xc0d3c0d3c0d3c0d3c0d3c0d3c0d3c0d3c0d30014](https://sepolia-optimism.etherscan.io/address/0xc0d3c0d3c0d3c0d3c0d3c0d3c0d3c0d3c0d30014#code).

**Tests**

This is a test transaction that's performing a sanity check on a previous upgrade on L2 where the L2ProxyAdmin Owner was set to the aliased L1ProxyAdmin Owner address: https://github.com/ethereum-optimism/superchain-ops/pull/204

PR #207, opened 2024-05-23
sep: resigning for presigned pause

blmalone:
This [pr](https://github.com/ethereum-optimism/superchain-ops/pull/205) invalidates presigned pause transactions on sepolia.


blmalone:
Closing because we have more than just one presigned pause tx signed.

PR #218, opened 2024-05-31
Move fp upgrade to 009

maurelian:
**Description**

Moves the FP upgrade task to folder `tasks/eth/009-...` to make space at 008 for the presigned pauses.


.circleci/config.yml
```diff
@@ -155,6 +155,27 @@ jobs:
             just simulate-council
             just prepare-json
             just simulate-council # simulate again to make sure the json is still valid
+  just_simulate_009-fp-upgrade:
```
mds1, line 158: Need to add this job to the bottom too so it actually runs, it's currently unused

maurelian: every time.
fixed

mds1 (approved):

PR #232, opened 2024-06-12
mainnet-013: Fault Proofs Fjord Upgrade

Inphi:
Mainnet task for the Fjord Upgrade. See [gov post](https://gov.optimism.io/t/upgrade-proposal-9-fjord-network-upgrade/8236) for the details.


tasks/eth/013-fp-upgrade-fjord/.env
```diff
@@ -0,0 +1,9 @@
+ETH_RPC_URL="https://ethereum.publicnode.com"
+SUPERCHAIN_CONFIG_ADDR=0x95703e0982140d16f8eba6d158fccede42f04a4c
+OWNER_SAFE=0x5a0Aae59D09fccBdDb6C6CcEB07B7279367C3d2A
+COUNCIL_SAFE=0xc2819DC788505Aac350142A7A707BF9D03E3Bd03
+FOUNDATION_SAFE=0x847B5c174615B1B7fDF770882256e2D3E95b9D92
+
+SAFE_NONCE_0X5A0AAE59D09FCCBDDB6C6CCEB07B7279367C3D2A=4
+SAFE_NONCE_0XC2819DC788505AAC350142A7A707BF9D03E3BD03=5
+SAFE_NONCE_0X847B5C174615B1B7FDF770882256E2D3E95B9D92=4
```
mds1, line 9: Can you add a comment to justify each nonce here so it's easier to verify (i.e. based on the current nonce + expected transactions to execute before)?

tasks/eth/013-fp-upgrade-fjord/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+Status: DRAFT, NOT READY TO SIGN
+
+## Objective
+
+This is a playbook for upgrading Fault Proofs on Mainnet for Fjord compatibility.
+
+This sets the `FaultDisputeGame` and `PermissionedDisputeGame` implementations in the `DisputeGameFactory` to newly deployed contracts that have an updated prestate.
+
+Governance post of the upgrade can be found at https://gov.optimism.io/t/upgrade-proposal-9-fjord-network-upgrade/8236.
```
mds1, line 11: Let's use a similar format to what we typically have in e.g. https://github.com/ethereum-optimism/superchain-ops/tree/main/tasks/eth/010-1-guardian-upgrade

---

image

tasks/eth/013-fp-upgrade-fjord/.env
Inphi: Fixed. Though please confirm that the L2 POA change does not require the L1 Foundation Safe. The nonces here assume the safe nonces will remain unchanged at signing.

mds1: My checks showed that the last one (0x847B5c174615B1B7fDF770882256e2D3E95b9D92) should be 5. I went through the [sequencing table](https://www.notion.so/oplabs/4e021daa38444d8ebda3770a43c48a6b?v=7b9ecb4d958044c6b16ccbd9a195519e&pvs=4) and filled in the Hardcoded Nonce table, can you confirm? 

> Though please confirm that the L2 POA change does not require the L1 Foundation Safe. 

What did you mean here? The L2 PAO change was already completed as part of upgrade 8

tasks/eth/013-fp-upgrade-fjord/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+The governance proposal should be treated as the source of truth and used to verify the correctness of the onchain operations.
+
+Governance post of the upgrade can be found at https://gov.optimism.io/t/upgrade-proposal-9-fjord-network-upgrade/8236.
+
+Details on the upgrade procedure can be found in [EXEC.md](./EXEC.md). Signers need not validate the, but they are provided for reference.
+
+## Preparing the Upgrade
+
+The new Fault Dispute Game coontract implementations have been pre-deployed to mainnet.
```
mds1, line 26:
```suggestion
The new Fault Dispute Game contract implementations have been pre-deployed to mainnet.
```

tasks/eth/013-fp-upgrade-fjord/NestedSignFromJson.s.sol
```diff
@@ -0,0 +1,85 @@
...
+        string memory addressesJson;
+        // Read addresses json
+        try vm.readFile(
+            string.concat(vm.projectRoot(), "/lib/superchain-registry/superchain/extra/addresses/mainnet/op.json")
+        ) returns (string memory data) {
+            addressesJson = data;
+        } catch {
+            revert("Failed to read lib/superchain-registry/superchain/extra/addresses/mainnet/op.json");
+        }
+        dgfProxy = DisputeGameFactory(stdJson.readAddress(addressesJson, "$.DisputeGameFactoryProxy"));
```
mds1, line 41: This is reverting for me with `no matching value found at "$.DisputeGameFactoryProxy"`. I suspect we need to upgrade the superchain-registry dependency to the latest version.

We also should make sure to add this task to CI, which would also catch this

Inphi: whoops. I had it updated locally but neglected to add that change.
Yeah, a CI check would be nice. But there are alot of broken scripts in superchain-ops as they relied on older versions of submodules.

Inphi: Unless we configure CI to ignore tasks that have been `EXECUTED`.

mds1: Currently every task gets manually added to CI, check out the diff here: https://github.com/ethereum-optimism/superchain-ops/pull/240/files

mds1: In other words, this PR would only add this task to CI, not any of the previously executed tasks

tasks/eth/013-fp-upgrade-fjord/.env
Inphi: You're right. I was looking at a totally different signing ceremony table.

tasks/eth/013-fp-upgrade-fjord/NestedSignFromJson.s.sol
```diff
@@ -0,0 +1,85 @@
...
+                Vm.StorageAccess memory storageAccess = accountAccess.storageAccesses[j];
+                if (storageAccess.isWrite) {
+                    address account = storageAccess.account;
+                    require(
+                        // DisputeGameFactoryProxy is expected to be updated
+                        account == address(dgfProxy)
+                        // State changes the Safe's are also expected.
+                        || account == proxyAdminOwnerSafe || account == securityCouncilSafe
+                            || account == foundationUpgradesSafe,
+                        "state-100"
```
mds1, line 73: I get a revert with `state-100` when simulating. This is because we also have state changes in the LivenessGuard anytime the SC signs a transaction, which you can see in the tenderly simulation

tasks/eth/013-fp-upgrade-fjord/VALIDATION.md
```diff
@@ -0,0 +1,91 @@
...
+  **Value:** 0x0000000000000000000000000000000000000000000000000000000000000001 <br/>
+  **Meaning:** This is `owners[0xca11bde05977b3631167028862be2a173976ca11] -> 1`, so the key can be
+    derived from `cast index address 0xca11bde05977b3631167028862be2a173976ca11 2`.
+
+- **Key:** 0xe90b7bceb6e7df5418fb78d8ee546e97c83a08bbccc01a0644d599ccd2a7c2e0 <br/>
+  **Value:** 0x000000000000000000000000ca11bde05977b3631167028862be2a173976ca11 <br/>
+  **Meaning:** This is `owners[1] -> 0xca11bde05977b3631167028862be2a173976ca11`, so the key can be
+    derived from `cast index address 0x0000000000000000000000000000000000000001 2`.
+
+## State Changes
```
mds1, line 60: Needs an entry for state changes in the LivenessGuard, for Security Council only. The mapping key used for the slot calculation is the signer's own address, or the multicall3 address when simulating without ledger. The value written will be the timestamp of the simulation (or execution) so it will be different for everyone

tasks/eth/013-fp-upgrade-fjord/VALIDATION.md
```diff
@@ -0,0 +1,91 @@
...
+  **After:** [`0x000000000000000000000000f691f8a6d908b58c534b624cf16495b491e633ba`](https://etherscan.io/address/0xf691f8a6d908b58c534b624cf16495b491e633ba) <br/>
+  **Meaning:** This is `gameImpls[1] -> 0xf691f8a6d908b58c534b624cf16495b491e633ba` (where `1` is the [`PERMISSIONED_CANNON` game type](https://github.com/ethereum-optimism/optimism/blob/op-contracts/v1.4.0/packages/contracts-bedrock/src/dispute/lib/Types.sol#L31)), so the key can be derived from `cast index uint32 0 101`.
+
+### Safe Contract State Changes
+
+The only other state changes should be restricted to one of the following addresses:
+
+- L1 2/2 ProxyAdmin Owner Safe: `0x5a0Aae59D09fccBdDb6C6CcEB07B7279367C3d2A`
+  - The nonce (slot 0x5) should be increased from 4 to 5.
+  - Another key is set from 0 to 1 reflecting an entry in the `approvedHashes` mapping.
```
mds1, line 87: We need to explain how to verify that this key is correct. Let's pull this out into a ful `### 0x5a0Aae59D09fccBdDb6C6CcEB07B7279367C3d2A` state changes entry. Here is a reference implementation, you will just need to change the keys and the safe hashes: https://github.com/ethereum-optimism/superchain-ops/blob/main/tasks/eth/010-1-guardian-upgrade/VALIDATION.md#0x5a0aae59d09fccbddb6c6cceb07b7279367c3d2a-the-22-proxyadmin-owner-1

tasks/eth/013-fp-upgrade-fjord/NestedSignFromJson.s.sol
mds1: As clarified in https://github.com/ethereum-optimism/superchain-ops/pull/232#discussion_r1649226929, this will only show when simulating as the council, not as the foundation

tasks/eth/013-fp-upgrade-fjord/.env
Inphi: Fixed. Thanks for confirming this!

.circleci/config.yml
```diff
@@ -185,6 +185,23 @@ jobs:
...
+          name: just simulate eth/013-fp-upgrade-fjord
+          command: |
+            go install github.com/mikefarah/yq/v4@latest
+            just install
+            cd tasks/eth/013-fp-upgrade-fjord
+            export SIMULATE_WITHOUT_LEDGER=1
+            just \
+              --dotenv-path .env \
+              --justfile ../../../nested.just \
+              simulate
```
mds1, line 203: Two changes needed here:
1. This is a nested task, so we need to simulate as both the council and foundation
2. This currently doesn't run since it's not listed in this section of the CI file:

```yml
check_just_simulate:
    jobs:
      # Please add an invocation to `just simulate` if a ceremony is
      # active (e.g. it is the next ceremony to perform or you
      # expect the ceremony to work continuously), and remove it once
      # the ceremony is for historical archive only (e.g. the
      # ceremony is done).
      - just_simulate_sc_rehearsal_1
      - just_simulate_sc_rehearsal_2
      - just_simulate_sc_rehearsal_4
      - just_simulate_sep-base-001-MCP-L1
```

```suggestion
  just_simulate_eth-013-fp-upgrade-fjord:
    docker:
      - image: <<pipeline.parameters.ci_builder_image>>
    steps:
      - checkout
      - run:
          name: just simulate eth/013-fp-upgrade-fjord
          command: |
            go install github.com/mikefarah/yq/v4@latest
            just install
            cd tasks/eth/013-fp-upgrade-fjord
            export SIMULATE_WITHOUT_LEDGER=1
            just \
              --dotenv-path .env \
              --justfile ../../../nested.just \
              simulate council
            just \
              --dotenv-path .env \
              --justfile ../../../nested.just \
              simulate foundation
```

mds1: Just unresolving to track item 1 in the comment above since that was not part of my suggest 😅

tasks/eth/013-fp-upgrade-fjord/VALIDATION.md
Inphi: That example was very useful. Do we also want to specify the expected nonce changes for the SC and FND safes? Or is that redundant given that the approval hash includes it?

.circleci/config.yml
Inphi: Good to go now. Thanks.

tasks/eth/013-fp-upgrade-fjord/VALIDATION.md
Inphi: Confirming that this task, and neither 011 nor 012, will be the first to update the liveness check. So we expect the initial timestamp at the time of execution to be zeroed.

tasks/eth/013-fp-upgrade-fjord/VALIDATION.md
```diff
@@ -0,0 +1,137 @@
...
+      ```shell
+        $ cast index bytes32 0x530085da61c08016f1625301df4686fcfe40291b98db29a4bf801237ca3098c4 0xaaf2b641eaf0bae063c4f2e5670f905e1fb7334436b902d1d880b05bd6228fbd
+        0x0f7b79f6b38abe5e02d33eec6fcdf7d9447ff17f6803d46f0afd5f628bac8504 
+        ```
+      And so the output of the second command matches the key above.
+
+- **Key:** `0x24424336f04440b1c28685a38303ac33c9d14a25`
+  **Before:** `0x0000000000000000000000000000000000000000000000000000000000000000`
+  **After:** `0x000000000000000000000000000000000000000000000000000000006675b61f`
+  **Meaning:** This updates the [`lastLive`](https://github.com/ethereum-optimism/optimism/blob/op-contracts/v1.5.0/packages/contracts-bedrock/src/Safe/LivenessGuard.sol#L36) indicating liveness of an owner that participated in signing. This will be updated to a block timestamp that's close to when this task was executed.
```
mds1, line 115: I think you are missing the header here since this is the Liveness Guard address, not the 2/2 Proxy Admin Owner. Also, the before value may not be zero for some signers—we can just note that in the Meaning text

tasks/eth/013-fp-upgrade-fjord/VALIDATION.md
mds1: I think it's good to specify, it gives extra confidence to signers that they are signing the right thing, since it's more user-friendly than the approveHash

tasks/eth/013-fp-upgrade-fjord/VALIDATION.md
Inphi: Good catch. Thanks for your patience 😅

tasks/eth/013-fp-upgrade-fjord/VALIDATION.md
```diff
@@ -0,0 +1,167 @@
...
+  **After:** `0x0000000000000000000000000000000000000000000000000000000000000006` <br/>
+  **Meaning:** The nonce is increased from 5 to 6. The key can be validated by the location of the `nonce` variable in the [Safe's Storage Layout](https://github.com/safe-global/safe-smart-account/blob/v1.3.0/contracts/examples/libraries/GnosisSafeStorage.sol#L17). Note that if this task is simulated before 011 has executed, then the nonce should be increased from 4 to 5.
+
+### `0x24424336F04440b1c28685a38303aC33C9D14a25` (`LivenessGuard`)
+
+State Changes:
+
+#### For the Council:
+
+- **Key:** `0xee4378be6a15d4c71cb07a5a47d8ddc4aba235142e05cb828bb7141206657e27` <br/>
```
mds1, line 164:
```suggestion
- **Key:** Compute with `cast index address {yourSignerAddressHere} 0` <br/>
```

tasks/eth/013-fp-upgrade-fjord/VALIDATION.md
```diff
@@ -0,0 +1,167 @@
...
+  **Before:** `0x0000000000000000000000000000000000000000000000000000000000000000` <br/>
+  **After:** `0x000000000000000000000000000000000000000000000000000000006675b61f` <br/>
+  **Meaning:** This updates the [`lastLive`](https://github.com/ethereum-optimism/optimism/blob/op-contracts/v1.5.0/packages/contracts-bedrock/src/Safe/LivenessGuard.sol#L36) indicating liveness of an owner that participated in signing. This will be updated to a block timestamp that's close to when this task was executed. Note that the "before" value may be non-zero for signers that have participated in signing.
```
mds1, line 167:
```suggestion
  **Meaning:** This updates the [`lastLive`](https://github.com/ethereum-optimism/optimism/blob/op-contracts/v1.5.0/packages/contracts-bedrock/src/Safe/LivenessGuard.sol#L36) indicating liveness of an owner that participated in signing. This will be updated to a block timestamp that matches the time when this task was executed. Note that the "before" value may be non-zero for signers that have participated in signing.
  
The only other state change is a nonce increment of the account that sent the transaction.
```

tasks/eth/013-fp-upgrade-fjord/VALIDATION.md
```diff
@@ -0,0 +1,167 @@
...
+
+- The following state changes (and none others) are made to that contract. This validates that no unexpected state changes occur.
+- All addresses (in section headers and storage values) match the provided name, using the Etherscan and Superchain Registry links provided. This validates the bytecode deployed at the addresses contains the correct logic.
+- All key values match the semantic meaning provided, which can be validated using the storage layout links provided.
+
+## State Overrides
+
+The following state overrides should be seen:
+
+### `0x5a0Aae59D09fccBdDb6C6CcEB07B7279367C3d2A` (The 2/2 `ProxyAdmin` Owner)
```
mds1, line 16: Can we sort the address keys here to match the order they are shown in tenderly (alphabetical)? This just makes it easier for singers to cross-check everything

tasks/eth/013-fp-upgrade-fjord/VALIDATION.md
```diff
@@ -0,0 +1,169 @@
...
+
+## State Changes
+
+### `0x24424336F04440b1c28685a38303aC33C9D14a25` (`LivenessGuard`)
+
+State Changes:
+
+#### For the Council:
+
+- **Key:** Compute with `cast index address {yourSignerAddress} 0` <br/>
```
mds1, line 68: nit, just a little more readable 
```suggestion
### `0x24424336F04440b1c28685a38303aC33C9D14a25` (`LivenessGuard`, Council only)

State Changes:

- **Key:** Compute with `cast index address {yourSignerAddress} 0` <br/>
```

tasks/eth/013-fp-upgrade-fjord/VALIDATION.md
```diff
@@ -0,0 +1,169 @@
...
+  **After:** [`0x000000000000000000000000c307e93a7c530a184c98eade4545a412b857b62f`](https://etherscan.io/address/0xc307e93a7c530a184c98eade4545a412b857b62f) <br/>
+  **Meaning:** This is `gameImpls[0] -> 0xc307e93a7c530a184c98eade4545a412b857b62f` (where `0` is the [`CANNON` game type](https://github.com/ethereum-optimism/optimism/blob/op-contracts/v1.4.0/packages/contracts-bedrock/src/dispute/lib/Types.sol#L28), so the key can be derived from `cast index uint32 0 101`.
+
+- **Key:** `0xffdfc1249c027f9191656349feb0761381bb32c9f557e01f419fd08754bf5a1b` <br/>
+  **Before:** `0x0000000000000000000000004146df64d83acb0dcb0c1a4884a16f090165e122` <br/>
+  **After:** [`0x000000000000000000000000f691f8a6d908b58c534b624cf16495b491e633ba`](https://etherscan.io/address/0xf691f8a6d908b58c534b624cf16495b491e633ba) <br/>
+  **Meaning:** This is `gameImpls[1] -> 0xf691f8a6d908b58c534b624cf16495b491e633ba` (where `1` is the [`PERMISSIONED_CANNON` game type](https://github.com/ethereum-optimism/optimism/blob/op-contracts/v1.4.0/packages/contracts-bedrock/src/dispute/lib/Types.sol#L31)), so the key can be derived from `cast index uint32 0 101`.
```
mds1, line 167:
```suggestion
  **Meaning:** This is `gameImpls[1] -> 0xf691f8a6d908b58c534b624cf16495b491e633ba` (where `1` is the [`PERMISSIONED_CANNON` game type](https://github.com/ethereum-optimism/optimism/blob/op-contracts/v1.4.0/packages/contracts-bedrock/src/dispute/lib/Types.sol#L31)), so the key can be derived from `cast index uint32 0 101`.
  
The only other state changes is a nonce increments on the account that sent the transaction.
```

mds1 (approved):
This is looking great, thank you!

mds1 (approved):

PR #237, opened 2024-06-17
test: restrict task statuses

mds1:
Adds `./script/utils/check-task-statuses.sh` to enforce a fixed set of allowed statuses, and requires a link to the executed transaction when the status is EXECUTED. This also removes the goerli task directories—goerli is deprecated, and there wasn't much in those directories anyway.


.circleci/config.yml
```diff
@@ -186,7 +195,6 @@ jobs:
             forge fmt --check
 
 workflows:
-  version: 2
```
mds1, line 189: This key is deprecated: https://github.com/SchemaStore/schemastore/blob/91d3af2df2a62f93faaf1e97c4b642557bf81440/src/schemas/json/circleciconfig.json#L1443

maurelian (approved):

PR #239, opened 2024-06-19
Rehearsal 5

maurelian:
Creates a new template for Rehearsal 5. 

By following the instructions in the template's README

1. the facilitator should be able to: 
	- deploy a new system to use for the rehearsal
	- generate a new rehearsal task
2. signers should be able to complete the signing successfully with the generated input.json file.


security-council-rehearsals/README.md
```diff
@@ -20,6 +24,8 @@ where `rehearsal-name` can be one of the following rehearsal names:
    members from the multisig.
 3. `r4-jointly-upgrade`: a rehearsal to perform a protocol upgrade
    jointly with Optimism Foundation via a nested 2-of-2 multisig.
+3. `r5-guardian-override`: a rehearsal to unpause withdrawals and
```
mds1, line 27: nit since markdown rendering would fix it
```suggestion
4. `r5-guardian-override`: a rehearsal to unpause withdrawals and
```

foundry.toml
```diff
@@ -21,6 +21,9 @@ remappings = [
   '@solady=lib/optimism/packages/contracts-bedrock/lib/solady/src',
   'ds-test/=lib/optimism/packages/contracts-bedrock/lib/forge-std/lib/ds-test/src',
   'forge-std/=lib/forge-std/src/',
+  'safe-contracts/=lib/optimism/packages/contracts-bedrock/lib/safe-contracts/contracts/',
+  'safe-contracts-v1.3.0/=lib/safe-contracts/contracts/',
```
mds1, line 25: Can you remind me why we have this twice? I believe all safe's are on 1.3.0, but the monorepo version is different?

security-council-rehearsals/templates/r5-guardian-override/README.md
mds1, line 1: We should add this task to CI

security-council-rehearsals/templates/r5-guardian-override/.env
```diff
@@ -0,0 +1,9 @@
+# Modify these values according to the Rehearsal instructions
+COUNCIL_SAFE=
+GuardianSafe_ADDRESS=
+SuperchainConfig_ADDRESS=
+
+# Do not edit
+SAFE_NONCE=""
+DUMMY_DEPUTY_GUARDIAN_MODULE=0xAbCD1234ABCD1234AbCD1234ABcD1234ABCD1234
```
mds1, line 8: Why is this address hardcoded, but not the above addresses? A DGM only can be used with 1 safe, but since we aren't actually using the DGM in the rehearsal—just rmeoving it—I'm guessing we reuse this DGM regardless of the above addresses?

mds1: I see, this is not even a real DGM, it's just a made up address

security-council-rehearsals/templates/r5-guardian-override/README.md
```diff
@@ -0,0 +1,297 @@
...
+- **Key:** `0x0000000000000000000000000000000000000000000000000000000000000004` <br/>
+  **Value:** `0x0000000000000000000000000000000000000000000000000000000000000001` <br/>
+   **Meaning:** Enables the simulation by setting the threshold to 1. The key can be validated by the location of the `threshold` variable in the [Safe's Storage Layout](https://github.com/safe-global/safe-smart-account/blob/v1.3.0/contracts/examples/libraries/GnosisSafeStorage.sol#L14).
+
+**In the State Changes:**
+
+The Guardian Safe (`$GuardianSafe_ADDRESS`), will having the following changes:
+
+1. the `nonce` will be increased from `2` to `4`. This is increased by two because the
+   Safe has made two separate calls.
```
mds1, line 107: Interesting, why isn't it just a single multicall with a nonce increment of 1?

script/DeployRehearsal5.s.sol
```diff
@@ -0,0 +1,112 @@
...
+pragma solidity ^0.8.0;
+
+import {GnosisSafe} from "safe-contracts-v1.3.0/GnosisSafe.sol";
+import {GnosisSafeProxyFactory} from "safe-contracts-v1.3.0/proxies/GnosisSafeProxyFactory.sol";
+import {ModuleManager} from "safe-contracts-v1.3.0/base/ModuleManager.sol";
+import {OwnerManager} from "safe-contracts-v1.3.0/base/OwnerManager.sol";
+import {Enum} from "safe-contracts-v1.3.0/common/Enum.sol";
+
+import {SuperchainConfig} from "@eth-optimism-bedrock/src/L1/SuperchainConfig.sol";
+// import {DeputyGuardianModule} from "@eth-optimism-bedrock/src/Safe/DeputyGuardianModule.sol";
```
mds1, line 11:
```suggestion
```

script/DeployRehearsal5.s.sol
```diff
@@ -0,0 +1,112 @@
...
+    ///         are used.
+    function name() public pure returns (string memory name_) {
+        name_ = "DeployRehearsal5";
+    }
+
+    function setUp() public {
+        councilSafe = GnosisSafe(payable(vm.envAddress("COUNCIL_SAFE")));
+
+        console.log("Deploying from %s", name());
+        console.log("Deployment context: %s", vm.envOr("DEPLOYMENT_CONTEXT", string("Deployment context not set")));
```
mds1, line 39: What is this `DEPLOYMENT_CONTEXT` for? Seems accidentally carried over from the monorepo to these rehearsal 4 and 5 scripts

script/DeployRehearsal5.s.sol
```diff
@@ -0,0 +1,112 @@
...
+        console.log("Deploying contracts for rehearsal");
+        vm.startBroadcast();
+        deployGuardianSafe();
+        deploySuperchainConfigProxy();
+        vm.stopBroadcast();
+    }
+
+    function deployGuardianSafe() public {
+        address[] memory owners = new address[](1);
+        owners[0] = msg.sender;
```
mds1, line 52: tangential, non-blocker: I like [solarray](https://github.com/emo-eth/solarray) for nicer array UX in tests/scripts. Might be nice to move towards this moving forward

maurelian: Noted :)

script/DeployRehearsal5.s.sol
maurelian: Yeah, I think it can be removed.

security-council-rehearsals/templates/r5-guardian-override/README.md
maurelian: This ceremony is done as a single ceremony, meaning there is a multicall with two tx from the SC Safe. Since both of those tx go to the Guardian Safe, it has two nonce changes.

If we want to do the multicall from the Guardian Safe, we could use the `nested.just` script instead, but then we would have to do the multi-step `approve` and `execute` and I imagine that other complications would arise which I have not yet considered.

foundry.toml
maurelian: Yes, that's correct. Mainly this is so that we can use the `GnosisSafe.sol` naming here of v1.3.0, rather than `Safe.sol` naming of v1.4.1 as used in the monorepo. 

I don't recall all of the details, but I have spent a fair bit of time in the past trying to avoid having to do this, and quite a few challenges crop up. I believe they are mostly related to the dependency graph, remappings and the import names used across this repo, `base-contracts` and the monorepo.

script/DeployRehearsal4.s.sol
maurelian, line 1: Renamed to distinguish between rehearsal deploy scripts.

security-council-rehearsals/templates/r5-guardian-override/README.md
maurelian: dd9a4c1c8730fe9679cefc4f1981a47beac94f0e

maurelian: dd9a4c1c8730fe9679cefc4f1981a47beac94f0e

security-council-rehearsals/templates/r5-guardian-override/.env
```diff
@@ -0,0 +1,9 @@
+# Modify these values according to the Rehearsal instructions
+COUNCIL_SAFE=0xAB23dE0DbE0aedF356af3F815c8B47D88575D82d
```
maurelian, line 2: This safe is the same as used in previous rehearsals, the SC lead prefers to use it.

security-council-rehearsals/templates/r5-guardian-override/.env
```diff
@@ -0,0 +1,9 @@
...
+GuardianSafe_ADDRESS=0x91dCf0F534B36218FE7926eac40089eF54006421
+SuperchainConfig_ADDRESS=0xbEDBB0A4d75514c84D569977B14D400c2Af5b36D
```
maurelian, line 4: I've been unable to verify either the Proxy or Implementation for this contract.

PR #242, opened 2024-06-21
Add print_versions

maurelian:
A little QoL improvement for debugging.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -180,7 +180,25 @@ jobs:
             forge --version
             forge fmt --check
 
+  print_versions:
+    docker:
+      - image: <<pipeline.parameters.ci_builder_image>>
+
+    steps:
+      - checkout
+      - run:
+          name: print versions
+          command: |
+            just --version
+            yq --version
+            forge --version
+            rustc --version
+            cargo --version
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -180,7 +180,25 @@ jobs:
...
+    steps:
+      - checkout
+      - run:
+          name: print versions
+          command: |
+            just --version
+            yq --version
+            forge --version
+            rustc --version
+            cargo --version
```
mds1, line 196: CI is failing with `/bin/bash: line 4: rustc: command not found`. I don't think we require rustc or cargo as deps for this repo

maurelian: 52c3075652c92909091525e7f4b318032ffe342b

maurelian: oh, cargo too. kk

maurelian: OK also: 105e310f9372081fb818daa96df53c24af4cc79c

I made the `just` installation use either `cargo` or `brew`. I think it's fine not to document how to install either of those tools, if a signer has trouble with that step we can just help them directly.

mds1 (approved):

Issue #152, opened 2024-04-17
feat: make postCheck hook required


mds1:
Currently the default `_postCheck` hook just logs a warning, to encourage people to implement it when creating tasks (example [here](https://github.com/ethereum-optimism/superchain-ops/blob/651056909a606e0f6a2b7bdbaf9e61ff1c67e002/script/SignFromJson.s.sol#L39)). It would be ideal to have a `require(false, "_postCheck not implemented")` instead to force an implementation.

However, this is not currently implemented because it breaks the `sc_rehearsal_1`, `sc_rehearsal_2`, and `sc_rehearsal_4` tasks which we want to keep in CI. Therefore the scope of this issue is:
1. Give those tasks their own local solidity file, and have them implement an empty post-check.
2. Change all `_postCheck` hooks to have a reverting, default implementation (using `require(false)` because this prevents solidity from emitting "unreachable code" warnings)

ShubhSensei:
May I work on this issue?

ShubhSensei:
I'm unable to push my code on GitHub in optimism-superchain-ops repo. What should I do? 

![image0.png](https://github.com/ethereum-optimism/superchain-ops/assets/91625601/cca89665-52f0-4552-bc4a-774dab4fb55e)

mds1:
This was completed by https://github.com/ethereum-optimism/superchain-ops/pull/266, thank you @ShubhSensei!

PR #245, opened 2024-06-22
Made postCheck hook required

ShubhSensei:
Issue #152 

**Description**

- Made _postCheck hook Require
- Made separate files for `sc_rehearsal_1`, `sc_rehearsal_2` & `sc_rehearsal_4`


mds1:
Thank you! We have a few ongoing mainnet tasks at the moment, so will hold off on reviewing this until after those are all executed :)

script/SignFromJson.s.sol
```diff
@@ -35,6 +35,5 @@ contract SignFromJson is MultisigBuilder, JsonTxBuilderBase {
     {
         accesses; // Silences compiler warnings.
         simPayload;
-        console.log("\x1b[1;33mWARNING:\x1b[0m _postCheck not implemented");
-    }
```
maurelian, line 39: Looks like you accidentally deleted the `}` which is causing a compiler failure.

maurelian (changes requested):
Overall the changes look sane, and now is a good time to get this merged as there are no pending playbooks in CI, however there are a few blockers:

2. a build failure I've pointed out inline
2. for each rehearsal _template_ the `justfile` will also need to be updated to replace `SignFromJson` with the new scripts added here.

We also need to figure out why CI is not running on this PR (it should be failing for the reasons above). 

In other repos I'm used to at least seeing a message indicating that a maintainer will need to take action to get the checks running. But here they are just being skipped.

maurelian:
Got the checks running by pushing a different branch (postcheck_hook_required-2) with the same commit.

script/SignFromJson.s.sol
ShubhSensei: Oops, I accidentally forgot it, that's why CI is crashing, I updated it.

maurelian:
FYI @ShubhSensei, I took your commits and am building on them over at #266

maurelian:
Closing in favor of #266, which builds on these commits but incorporates some subtle patterns used in this repo.

PR #264, opened 2024-07-22
ci: Clarify what statuses are accepted

maurelian:
The current error message suggests that a PR cannot be merged if it is labelled as in `DRAFT` status, this is true only because DRAFT doesn't match list of accepted statuses, not because the play book can't be in draft.

image

### DIFF
--- a/script/utils/check-task-statuses.sh
+++ b/script/utils/check-task-statuses.sh
@@ -26,7 +26,11 @@ check_status_and_hyperlinks() {
   done
 
   if [[ "$is_valid_status" == false ]]; then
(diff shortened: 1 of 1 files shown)


blmalone (approved):
lgtm

maurelian:
This is wrong, it only prints the first valid status: 
2024-07-22 at 10 31@2x

maurelian:
Looks like this now: 
2024-07-22 at 11 19@2x

mds1 (approved):
This could get verbose and make the results hard to read when there are a lot of invalid statuses, but since typically only 1 task is added per PR in practice it should be ok

PR #266, opened 2024-07-22
Make `postCheck` hook required

maurelian:
This PR builds on @ShubhSensei's changes in #245. There were few non-obvious things about how this repo is used, so it seemed easier to make those tweaks myself. 

Thanks for the commits @ShubhSensei!


security-council-rehearsals/templates/r1-hello-council/justfile
```diff
@@ -1,9 +1,9 @@
 set dotenv-load
+export location := `pwd`
+export OWNER_SAFE := env_var("COUNCIL_SAFE")
```
maurelian, line 3: This change wasn't strictly necessary, but it makes this file more consistent with the r2 and r4 rehearsal justfiles.

security-council-rehearsals/templates/r1-hello-council/SignFromJson.s.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.15;
+
+// This import path should be relative to where the rehearsal script will be located after a new rehearsal is setup,
+// which is one level above the current location in
+// repo_root/security-council-rehearsals/<rehearsal-dir>/SignFromJson.s.sol
+import {SignFromJson as OriginalSignFromJson} from "../../script/SignFromJson.s.sol";
+import {Vm, VmSafe} from "forge-std/Vm.sol";
```
maurelian, line 9: This import shouldn't be necessary, but it [wasn't compiling](https://app.circleci.com/pipelines/github/ethereum-optimism/superchain-ops/1632/workflows/5d6e7f19-24ea-4987-819c-1abe15a3c4f0/jobs/14890) without it, and who am I to blow against the wind.

script/PresignPauseFromJson.s.sol
```diff
@@ -55,6 +55,7 @@ contract PresignPauseFromJson is MultisigBuilder, JsonTxBuilderBase {
         override
     {
         checkStateDiff(accesses);
+        require(false, "_postCheck not implemented");
```
Inphi, line 58: the code below is dead. any use keeping it around?

security-council-rehearsals/templates/r1-hello-council/SignFromJson.s.sol
mds1: I've noticed this before too and was surprised the import was needed, it's harmless enough at least 🤷

script/PresignPauseFromJson.s.sol
mds1: Hm good point, @maurelian since this is `PresignPauseFromJson`, I imagine the post checks will be the same every time so maybe we can just delete the `require` here?

security-council-rehearsals/templates/r1-hello-council/SignFromJson.s.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.15;
+
+// This import path should be relative to where the rehearsal script will be located after a new rehearsal is setup,
+// which is one level above the current location in
+// repo_root/security-council-rehearsals/<rehearsal-dir>/SignFromJson.s.sol
+import {SignFromJson as OriginalSignFromJson} from "../../script/SignFromJson.s.sol";
```
mds1, line 8: I think these relative paths need an extra `../` to reach the root script. Alternatively could just use an absolute path that starts with `script/`. Currently when running `just simulate` from within `security-council-rehearsals/templates/r1-hello-council` it errors with a file not found:

```
[⠢] Compiling 9 files with Solc 0.8.15
[⠆] Solc 0.8.15 finished in 36.04ms
Error: 
Compiler run failed:
Error (6275): Source "security-council-rehearsals/script/SignFromJson.s.sol" not found: File not found. Searched the following locations: "/Users/mds/Documents/projects/superchain-ops".
ParserError: Source "security-council-rehearsals/script/SignFromJson.s.sol" not found: File not found. Searched the following locations: "/Users/mds/Documents/projects/superchain-ops".
 --> security-council-rehearsals/templates/r1-hello-council/SignFromJson.s.sol:8:1:
  |
8 | import {SignFromJson as OriginalSignFromJson} from "../../script/SignFromJson.s.sol";
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

maurelian: Right. The tricky thing here, which the comment above tries to clarify, is that the correct flow is what's written into the CI config file. So, it needs to compile once the this entire folder is copied into a new folder one level higher. 

It's not pretty, but this follows the convention of other rehearsals and satisfies the exisiting CI checks.

script/PresignPauseFromJson.s.sol
maurelian: Oh, good callout. I agree this `require` statement can be deleted.

security-council-rehearsals/templates/r1-hello-council/SignFromJson.s.sol
mds1: I was confused by that comment because the path it gives is:
```
repo_root/security-council-rehearsals/<rehearsal-dir>/SignFromJson.s.sol
```
But there is no `SignFromJson.s.sol` in either the `security-council-rehearsals/<rehearsal-dir>` folders

I still am confused here though, so my original comment of:

> Currently when running `just simulate` from within `security-council-rehearsals/templates/r1-hello-council` it errors with a file not found:

Is that expected to not work?

maurelian: > But there is no SignFromJson.s.sol in either the security-council-rehearsals/<rehearsal-dir> folders

That's because those rehearsals were created prior to this PR.

maurelian: > > Currently when running `just simulate` from within `security-council-rehearsals/templates/r1-hello-council` it errors with a file not found:
> 
> Is that expected to not work?

Without a deeper refactor, I'm not sure how to make both work with the current flow used to create a new rehearsal with `just setup`. Am open to suggestions though.

mds1: If the templates are expected to not work that is ok, just wasn't sure if it was expected behavior :)

mds1 (approved):

PR #238, opened 2024-06-19
mainnet-011: Fjord setRecommended PV & ownership transfers (WIP)

sebastianst:
**Description**

* mainnet `ProtocolVersions.setRecommended` to 7.0.0 (Fjord)
* ownership transfers for `ProtocolVersions` and `SystemConfig` from current owner _Foundation Operation Safe_ to _Foundation Upgrade Safe_.
* template updates
* adding allowed state change checks to base contract
* couple lil docs changes

**Tests**

Added state diff assertions

PR #270, opened 2024-07-24
Update Fault Proof Sepolia Recovery Tasks

awilliams1-cb:
**Description**

Updated the tasks under `fp-recovery` to address the following:

- Further parameterized existing tasks so that they can be used with the Base Sepolia FP contracts
- Fixed existing tasks so that they are compatible with breaking changes https://github.com/ethereum-optimism/superchain-ops/pull/260 and https://github.com/ethereum-optimism/superchain-ops/pull/238

A more detailed breakdown of changes per each task can be found below.

`001-blacklist-game`, `002-fallback-permissioned-game`, and `003-enable-permissionless-game`:

- Parameterized the `OptimismPortalProxy` and L2 chain that can be targeted.
- Added parsing of toml configs from superchain-registry.
- Added `getAllowedStorageAccess` function.
- Updated README.md to reflect new setup steps and validation.

`004-recover-bonds`:

- Updated `.env` to add `WAD_VALUE` variable and cleared values of existing variables `DELAYED_WETH_PROXY` and `OWNER_SAFE`.
- Added `SignFromJson.s.sol` with minimal `_postCheck` implementation. Note this task is not broken yet but will break when https://github.com/ethereum-optimism/superchain-ops/pull/266 is pushed.
- Updated README.md to reflect new setup steps and validation.

**Tests**

Each task was re-tested with the changes against OP and Base's Sepolia Fault Proof contracts. New screenshots of the state changes for OP have been provided, and the state changes were verified against the existing validation steps for each respective recovery task.

**Additional context**

If this needs to be split up into one PR per task change just let me know. I also intend to add another recovery task for removing anchor roots from the `AnchorStateRegistry` as well in a follow-on PR.


Inphi (approved):
Looks good. Thanks

PR #280, opened 2024-07-26
tasks(eth/mmz) Mode, Metal, and Zora Mainnet key handover

sbvegan:
**Description**

Mode, Metal, and Zora Mainnet Key Handover to transfer the ProxyAdminOwner role to the Optimism Foundation and Security Council's 2-of-2 multisig

PR #281, opened 2024-07-29
Updates to mmz _postCheck()

maurelian:
**Description**

This is a PR building on top of #280, which fixes the `checkStateDiff()` implementation. 

My suggest on the Sepolia task was wrong and I'll open a PR to fix it as well.

### DIFF
--- a/tasks/eth/mmz-002-key-handover/SignFromJson.s.sol
+++ b/tasks/eth/mmz-002-key-handover/SignFromJson.s.sol
@@ -59,14 +59,6 @@ contract SignFromJson is OriginalSignFromJson {
 
         return stdJson.readAddress(addressesJson, string.concat("$.", chainId, ".ProxyAdmin"));
     }
-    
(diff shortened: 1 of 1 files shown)


sbvegan (commented):
If I'm undertsanding this correctly:
- the `checkStateDiff` override was removed and just uses the inherited function from `SignFromJson.s.sol`
- getAllowedStorageAccess is added to ensure only the ProxyAdmins for each chain can have their owner slot modified

maurelian:
> * the `checkStateDiff` override was removed and just uses the inherited function from `SignFromJson.s.sol`
> * getAllowedStorageAccess is added to ensure only the ProxyAdmins for each chain can have their owner slot modified

Yes, that's right. Here's another example of the [correct implementation](https://github.com/ethereum-optimism/superchain-ops/blob/main/tasks/eth/base-001-MCP-L1/NestedSignFromJson.s.sol#L388). 

Basically `checkStateDiff` should not need to be modified, as long as `getCodeExceptions` and `getAllowedStorageAccess` are correctly overriden.

PR #286, opened 2024-07-30
ci: refactor to a single workflow

mds1:
Refactors CI to a single `main` workflow, similar to what we have in the monorepo. This makes it easier to add new jobs to CI and have them automatically required as passing before merging PRs. With the current approach, we have to go into settings and add each new workflow as a required status check. With this approach, we add `main` as the only required status check.

I also added a check to fail on solc warnings. This won't do much due to our nonstandard repo structure so not everything compiles by default, but it we should gradually move to a more standard structure over time

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,7 +59,7 @@ jobs:
       - run:
           name: Check Mainnet RPC Endpoints
           command: |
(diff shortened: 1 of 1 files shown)


maurelian (approved):

mds1:
Now that this is merged I have updated the repo settings
image

PR #306, opened 2024-08-23
PresignPause: add `zip-PSPs` + `upload-PSPs` + fix the `getAllowedStorageAccess()`. 

Ethnical:
**Description**
### 1. `upload-PSPs` & `zip-PSPs`. 
This PR includes to new features for the PSPs tooling. 
- `upload-PSPs`: A simple tooling that will upload the `ready-X` file directly into the 1password (I know we have a similar tooling already built in go but I got issue when using it). This is not using a wrapper but the `op` cli directly from just.  
Small video to see how to run it: 

https://github.com/user-attachments/assets/7caebaed-6b46-4afe-acc7-816e7910c3b6

- `zip-PSPs`: Is a simple tool that will zip the PSPs with a password. This is useful for sharing the PSPs with Base. We have already this notion doc for the process of sharing the PSPs but this make easier and faster. 

https://github.com/user-attachments/assets/f1fd8ecc-2781-4d69-9baf-e5cdde156698

Both of the tools (`upload-PSPs` and  `zip-PSPs`) are not necessary for the current process but helped me for the doing the https://github.com/ethereum-optimism/security-pod/issues/137. 

### 3. `getAllowedStorageAccess()` 
To make sure, we are not overwriting an undesired storage there is some checks made using we have to introduce the contract address that are allowed to be modified. 
However, for the PSPs we EXACTLY knows EACH time what contracts storage has to be changed.
So to avoid to create a file in the [current task](https://github.com/ethereum-optimism/superchain-ops/blob/main/tasks/sep/017-presigned-pause/PresignPauseFromJson.s.sol) each time (we can simply add the FoS + SuperChainConfig directly into the file). 

 `getAllowedStorageAccess()`. This checks has been implemented **recently** and this is why was **not required in the past**.
The idea is to make the changes directly into the codebase `PresignPauseFromJson` to avoid having multiples files or during the generation of the PSPS.
So added the storage like this directly:
```diff
 {
+     allowed = new address[](2);
+        allowed[0] = vm.envAddress("SUPERCHAIN_CONFIG_ADDR"); // The storage for the pause will be set to `1`.
+      allowed[1] = _ownerSafe(); // The nonce is updated in the FoS.
    }
```
### 4. removing the `PresignedPauseFromJson_eth_008.s.sol`

Removing the `superchains-ops/script/PresignedPauseFromJson_eth_008.s.sol` to clean up the repository as this is mentioned there: 
 ```solidity
[...]

PR #308, opened 2024-08-28
Add an new CLI  ARG `VAULTNAME` to select the vault name.

Ethnical:
**Description**

By updating the internal [doc](https://www.notion.so/oplabs/RB-111-Pre-signed-pause-8bce5b27728b447aa51ba12fcfff9e58?pvs=4) I realized that we don't have a way to select the vault we actually want to update. 
So this PR include a way to choose the vault name.  
```bash
just \
   --dotenv-path .env \
   --justfile ../../../presigned-pause.just \
   upload-PSPs "VAULTNAME"

```
**Errors**
In the case,  the vault doesn't exist or you don't have the permission this will return an error like this: 
933ae2d6c7cd14cc14e5935b041f017083d06f4f34891e44e698d09279e818f3

**Tests**

I tested it properly using the sepolia PSPs and this worked as expected with the following command: 
```bash
just \
   --dotenv-path .env \
   --justfile ../../../presigned-pause.just \
   upload-PSPs "Pre-signed Pause OP Sepolia"
```

**Additional context**
This PR intervene after this [PR](https://github.com/ethereum-optimism/superchain-ops/pull/306) where I forgot to introduce this feature. 

**Metadata**

- Fixes #[Link to Issue]


presigned-pause.just
```diff
@@ -157,12 +157,12 @@ simulate-all:
       simulate
   done
 
-upload-PSPs:
+upload-PSPs VAULTNAME:
```
mds1, line 160: Nice catch on the missing vault name. Instead of specifying it as an arg, can we infer the vault name automatically based on the directory? e.g. `tasks/eth` uses the mainnet one, `tasks/sep` uses the sepolia one, and other dirs throw an error

Ethnical: Oh wanted to something similar actually!
But then realized I do prefer that security oncall choose the vault manually by adding this a param (this will also be consistent with previous doc + new one). 
This will add an small checks extra step. 
Maybe in the future version?

mds1: > But then realized I do prefer that security oncall choose the vault manually by adding this a param (this will also be consistent with previous doc and new doc created for this).

What is the reason for this preference? It seems less error prone if we can automate it

Ethnical: Yes, the main reasons was:
- Be consistent with previous doc we got, keep the usage of the name into the args. 
- Also, personal feeling I prefer with most of the tooling we have having something that clearly specify where we want to work (Mainnet or Sepolia). 
Maybe we can add a small message to mention: 
```
You are currently on a ethereum task Path and you want to submit the PSPs on the Sepolia vault?
```

Ethnical: > But then, if there's only 1 allowed value of the arg for a given task, why have the arg in the first place?

Yes you right on this point. I have follow then the first idea and added something that will match automatically from the path if `sep` this will return the sepolia vault and vice versa for mainnet.
Now, no arguments is required.  
I just added a confirmation to make sure this is desired.
 
I took a video for the [doc](https://www.notion.so/oplabs/RB-111-Pre-signed-pause-8bce5b27728b447aa51ba12fcfff9e58?pvs=4#a227c38a09e14518b26095baa91f81c8) and also putting this one here to show you: 

https://github.com/user-attachments/assets/5c05deb7-73bf-49f9-9337-555aa3287a46

presigned-pause.just
```diff
@@ -160,18 +160,33 @@ simulate-all:
...
  if which op >/dev/null 2>&1; then
-    echo "op (one-password cli) is installed."
+    echo "op (one-password cli) is installed. We can proceed..."
  else
     echo "op is not installed, please install op (one-password cli) using homebrew. Using brew install --cask 1password-cli"
     exit
  fi
 
+ # Ask for confirmation before proceeding
+ read -p "Do you want to continue, with the ? (y/n): " confirm
```
mds1, line 184: Typo?
```suggestion
 read -p "Do you want to continue, with the upload? (y/n): " confirm
```

presigned-pause.just
mds1: Nice, thanks a lot for this, the UX looks really great!

mds1 (approved):
Great work, thank you!

PR #311, opened 2024-09-03
mainnet-018: Granite FP upgrade

Inphi:
This is the second batch for the [Granite protocol upgrade](https://gov.optimism.io/t/upgrade-proposal-10-granite-network-upgrade/8733). This task upgrades the Fault Proof contracts.

## Sequence

The following table outlines the upgrade sequence, including the nonce allocation of each safe.

| ordering | task | FUS | SC | FOS |
| ---- | ----- | ----- | ---- | ---- |
| [016](https://github.com/ethereum-optimism/superchain-ops/pull/310) | Set recommended/required ProtocolVersions | 7 | | |
| [017](https://github.com/ethereum-optimism/superchain-ops/pull/313) | Resign Presigned Pauses |  |  | 96 |
| **>> 018** | Upgrade FP contracts | 8 | 6 | |
| [019](https://github.com/ethereum-optimism/superchain-ops/pull/321) | Re-enable permissionless FPs + DGM Upgrade | | | 95 |

PR #319, opened 2024-09-05
Add _nestedPostCheck to NestedSignFromJson base script

maurelian:
Implemented the `_nestedPostCheck` function in the `NestedSignFromJson` base script.

This provides a robust fix to the failed simulation on approval calls.

Keeping in draft until #311 is merged to not break it.

### DIFF
--- a/script/NestedSignFromJson.s.sol
+++ b/script/NestedSignFromJson.s.sol
@@ -52,7 +52,6 @@ contract NestedSignFromJson is NestedMultisigBuilder, JsonTxBuilderBase {
     function _nestedPostCheck(Vm.AccountAccess[] memory accesses, SimulationPayload memory simPayload)
         internal
         virtual
-        override
     {
         accesses; // Silences compiler warnings.
         simPayload;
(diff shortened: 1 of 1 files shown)


script/NestedSignFromJson.s.sol
```diff
@@ -41,9 +41,20 @@ contract NestedSignFromJson is NestedMultisigBuilder, JsonTxBuilderBase {
...
         override
+    {
+        if (msg.sig == this.approveJson.selector) {
+            console.log("Skipping assertions on the approval call");
+            return;
+        }
+        _nestedPostCheck(accesses, simPayload);
+    }
+
+    function _nestedPostCheck(Vm.AccountAccess[] memory accesses, SimulationPayload memory simPayload)
```
mds1, line 52: Nit, what do you think about `_postCheckLogic` or similar, so it's more clear for the implementer?
```suggestion
    function _postCheckLogic(Vm.AccountAccess[] memory accesses, SimulationPayload memory simPayload)
```

maurelian: I like that the current name makes it clearer-ish that it's only a requirement for Nested ceremonies.

mds1: Oh! I read "nested" as meaning "nested within the main postcheck function" lol. Ok, this name does make sense then

mds1 (approved):

PR #220, opened 2024-06-03
Resign PSPs for DeputyGuardianModule

maurelian:
**Description**

Adds A new Presigned Pause ceremony, including custom overrides to enable simulation based on the state following the 010 ceremonies in #219.

PR #322, opened 2024-09-06
WIP:Fix the `SCRIPT` PATH For the PresignedPause justfile 

Ethnical:
**Description**

This PR modify the usage of the script used (`PresignPauseFromJson`) during the usage of the `presigned-pause.just`

Now, this will create custom script into the task directory directly exactly like the other standard tasks that are using the `single.just` (cf screenshot of the single tasks here):

8e0b904ce15536ded555393bfc93270a2fc3cabd9ac47fe2b67cca2d13c4070f

For now, if we need to overrides a PresignedPause task we have to create the file at the into `script/` that is not ideal and  would need to be removed after the task (i.e task [eth_017](https://github.com/ethereum-optimism/superchain-ops/pull/314/files) or task [eth_008](https://github.com/ethereum-optimism/superchain-ops/pull/220/files)) to clean up the repository..

Now, with this PR we can use the script directly into the same folder like this:
image
  

**Tests**

I have test using the `prepare` and `sign` using the `SIMULATE_WITHOUT_LEDGER` with the CLI like this: 
```bash
SIMULATE_WITHOUT_LEDGER=1 \
FOUNDRY_SENDER=0xF0871b2F75ecD07e2D709b7a2cc0AF6848c1cE76 \
TEST_SENDER=0xF0871b2F75ecD07e2D709b7a2cc0AF6848c1cE76 \
just \
   --dotenv-path .env \
   --justfile ../../../presigned-pause.just \
   sign
```

I just slightly modified the import like in the `single` task like below of the `017-presigned-pause/PresignPauseFromJson.s.sol`

```solidity
import {PresignPauseFromJson as OriginalPresignPauseFromJson} from "script/PresignPauseFromJson.s.sol";
```
**Additional context**
Merge this PR after the current ceremonies.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -179,7 +179,6 @@ jobs:
               install
             forge build
             export SIMULATE_WITHOUT_LEDGER=1
-            export SCRIPT=PresignPauseFromJson_eth_017
             export FOUNDRY_SENDER=0x3041BA32f451F5850c147805F5521AC206421623
             export TEST_SENDER=0x3041BA32f451F5850c147805F5521AC206421623
             just \
--- a/presigned-pause.just
+++ b/presigned-pause.just
@@ -77,7 +77,7 @@ prepare:
   set -x
   script=PresignPauseFromJson
   if [ -f "${taskPath}/PresignPauseFromJson.s.sol" ]; then
-    script="${taskPath}/PresignPauseFromJson.s.sol"
+    script="./PresignPauseFromJson.s.sol"
   fi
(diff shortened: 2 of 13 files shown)


maurelian (commented):
This LGTM. Agree we should not merge until after we have new PSPs.

Just to test this in CI, could you update the 017 job to work without setting the `SCRIPT` var?

Ethnical:
Sure do that right now!

maurelian (approved):
No longer blocked by 017

PR #326, opened 2024-09-09
feat(docs): enhance setup instructions with shell code blocks

maurelian:
Signers often run into the following error if a script has been removed since they last built:  

```
failed to read artifact source file for `/Users/maurelian/projects/o/superchain-ops/script/PresignPauseFromJson_eth_008.s.sol`
```

This is meant to address that.

### DIFF
--- a/justfile
+++ b/justfile
@@ -31,3 +31,6 @@ add-transaction bundlePath to sig *params:
   mv {{bundlePath}} ${backupBundlePath}
   mv ${newBundlePath} {{bundlePath}}
   echo "Old bundle backed up to ${backupBundlePath}."
+
+clean:
+  forge clean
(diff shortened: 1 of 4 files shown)


justfile
```diff
@@ -31,3 +31,6 @@ add-transaction bundlePath to sig *params:
   mv {{bundlePath}} ${backupBundlePath}
   mv ${newBundlePath} {{bundlePath}}
   echo "Old bundle backed up to ${backupBundlePath}."
+
+clean:
+  forge clean
```
maurelian, line 36: This thin wrapper command obviously isn't necessary, but it felt more consistent with how we use `just` everywhere else.

Ethnical: Agreed!

PR #347, opened 2024-10-17
Update `PRESIGNED-PAUSE.md` with the GAP disclaimer.

Ethnical:
**Description**
This PR include the informations in case an upgrade cause the changes to break the PSPs.
 
In the future, this will be less the case to happen since we will make the DGM proxify. 
As part of the design review for upgrading the DGM here (https://github.com/ethereum-optimism/design-docs/pull/89)

**Tests**

**Additional context**

**Metadata**

### DIFF
--- a/PRESIGNED-PAUSE.md
+++ b/PRESIGNED-PAUSE.md
@@ -45,15 +45,14 @@ A Facilitator will collect the signatures and execute the contract.
 ### Ensure no gaps with PSPs coverage
 
 We need to ensure there is no gaps in the PSPs coverage during upgrade.
-This means that, if we are upgrading core components of the system. It can break the PSPs coverage, thus we need to ensure before the upgrade that the PSPs coverage will not be broken.
-In the case there are broken changes, we need to blindly simulate the PSPs with the **new changes** and presign the new PSPs with the **new changes**.
-This will allow us to have the PSPs coverage during the upgrade and after the upgrade with the new components.
-We need to keep the PSPs before the upgrade and after the upgrade.
+Upgrades to certain components of the system can invalidate existing PSPs and cause them to no longer work, thus we need to ensure before the upgrade that PSPs coverage continues without a gap.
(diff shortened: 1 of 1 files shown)


PRESIGNED-PAUSE.md
```diff
@@ -40,6 +42,19 @@ doesn’t even need to be funded. These are offchain signatures.
 
 A Facilitator will collect the signatures and execute the contract.
 
+### 1. Ensure no GAPs with PSPs coverage
+
+We need to ensure there is no GAPs in the PSPs coverage during upgrade.
```
mds1, line 47: nit, can we change all stylizing of GAP to Gap/gap, since I think it's not an acronym in this context and you mean a literal gap

PRESIGNED-PAUSE.md
```diff
@@ -40,6 +42,19 @@ doesn’t even need to be funded. These are offchain signatures.
 
 A Facilitator will collect the signatures and execute the contract.
 
+### 1. Ensure no GAPs with PSPs coverage
```
mds1, line 45: nit, let's delete the number since it's the only header, and the other numbered headers are to indicate steps in a sequence (table of contents will also need to be updated)
```suggestion
### Ensure no GAPs with PSPs coverage
```

PRESIGNED-PAUSE.md
```diff
@@ -40,6 +42,19 @@ doesn’t even need to be funded. These are offchain signatures.
 
 A Facilitator will collect the signatures and execute the contract.
 
+### Ensure no gaps with PSPs coverage
+
+We need to ensure there is no gaps in the PSPs coverage during upgrade.
+This means that, if we are upgrading core components of the system. It can break the PSPs coverage, thus we need to ensure before the upgrade that the PSPs coverage will not be broken.
```
mds1, line 48:
```suggestion
This means that, if we are upgrading core components of the system, it can break the PSPs coverage, thus we need to ensure before the upgrade that the PSPs coverage will not be broken.
```

mds1 (approved):

PR #350, opened 2024-10-22
feat: executed gas target increase

blmalone:


bitwiseguy (approved):

mds1 (approved):
It looks like this task was never added to CI? If it was and I am missing it, let's make sure to remove that before merge

Issue #268, opened 2024-07-23
Wrong state diff checks during nested approval calls


sebastianst:
Nested approval calls seem to run the same state diff checks as the actual task execution. This lead to failing checks during the `DGF.setImplementation` task executions for Fjord.

It was locally mitigated by removing the task-specific checks and adding code-less address exceptions to the validation script (letting `getCodeExceptions()` return the `SC.getOwners()` array). The exceptions were necessary because the SC multisig records all member addresses in a timestamp mapping, and members are mostly EOAs, thus failing the generic code exception checks, that values of state changes that are addresses must contain code.

mds1:
My suggestion is to wrap the full contents of the `postCheck` method in an `if (msg.sig != this.approveJson.selector)` block. We don't need to run any checks when doing approve calls, so this ensures they only run for simulations and signing.

Example:
```solidity
/// @notice Checks the correctness of the deployment
function _postCheck(Vm.AccountAccess[] memory accesses, SimulationPayload memory /* simPayload */ )
    internal
    view
    override
{
    if (msg.sig != this.approveJson.selector) {
        console.log("Running post-deploy assertions");

        checkStateDiff(accesses);
        checkSemvers();
        // etc/

        console.log("All assertions passed!");
    } else {
        console.log("Skipping assertions because this is a just approve call");
    }
}
```

mds1:
@maurelian I think we can close this as completed by https://github.com/ethereum-optimism/superchain-ops/pull/319, do you agree?

PR #354, opened 2024-10-28
Add template to set a dispute game implementation

ajsutton:
**Description**

Adds a template to the fp-recovery directory to set the implementation for a particular game type. Most inputs are loaded from the superchain-registry to make it as easy as possible to adapt for other chains.


smartcontracts (commented):
Seems reasonable to me, would also like to get review from someone more familiar with writing superchain-ops tasks

tasks/sep/fp-recovery/005-set-game-implementation/.env
```diff
@@ -0,0 +1,8 @@
+ETH_RPC_URL="https://ethereum-sepolia.publicnode.com"
+SUPERCHAIN_CONFIG_ADDR=0xC2Be75506d5724086DEB7245bd260Cc9753911Be
```
mds1, line 2: Looks like this is unused in the task
```suggestion
```

tasks/sep/fp-recovery/005-set-game-implementation/NestedSignFromJson.s.sol
```diff
@@ -0,0 +1,188 @@
...
+
+    // Chains for this task.
+    string l1ChainName = vm.envString("L1_CHAIN_NAME");
+    string l2ChainName = vm.envString("L2_CHAIN_NAME");
+
+    // Safe contract for this task.
+    GnosisSafe securityCouncilSafe = GnosisSafe(payable(vm.envAddress("COUNCIL_SAFE")));
+    GnosisSafe fndSafe = GnosisSafe(payable(vm.envAddress("FOUNDATION_SAFE")));
+    GnosisSafe ownerSafe = GnosisSafe(payable(vm.envAddress("OWNER_SAFE")));
+    address livenessGuard = 0xc26977310bC89DAee5823C2e2a73195E85382cC7;
```
mds1, line 28: Safe's store the guard address in slot `0x4a204f620c8c5ccdca3fd54d003badd85ba500436a431f0cbda4f558c93c34c8` ([source](https://github.com/safe-global/safe-smart-account/blob/186a21a74b327f17fc41217a927dea7064f74604/contracts/base/GuardManager.sol#L30)), so we can read the guard address from the Safe instead of hardcoding it

tasks/sep/fp-recovery/005-set-game-implementation/NestedSignFromJson.s.sol
```diff
@@ -0,0 +1,188 @@
...
+
+    // Known EOAs to exclude from safety checks.
+    address l2OutputOracleProposer; // cast call $L2OO "PROPOSER()(address)"
+    address l2OutputOracleChallenger; // In registry addresses.
```
mds1, line 32: These are unused, let's delete them

tasks/sep/fp-recovery/005-set-game-implementation/README.md
```diff
@@ -0,0 +1,46 @@
+# Deputy Guardian - Enable Permissioness Dispute Game
```
mds1, line 1: It seems this task does not actually enable the game, just sets the implementation
```suggestion
# Deputy Guardian - Set Permissionless Dispute Game Implementation
```

tasks/sep/fp-recovery/005-set-game-implementation/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+**\_impl:** `<user-input>`
+
+## Preparing the Operation
+
+1. Copy this directory to the appropriate final task location.
+
+2. Update the path in `NestedSignFromJson.s.sol` `setUp` function to the new location
+
+3. Update the relative path to lib/superchain-registry in `justfile` if needed.
```
mds1, line 33: When is this needed?

tasks/sep/fp-recovery/005-set-game-implementation/NestedSignFromJson.s.sol
```diff
@@ -0,0 +1,188 @@
...
+
+        // And finally, we append the Safe signer exceptions.
+        for (uint256 i = 0; i < safeSignersWithNoCode.length; i++) {
+            shouldHaveCodeExceptions[4 + i] = safeSignersWithNoCode[i];
+        }
+
+        return shouldHaveCodeExceptions;
+    }
+
+    function _precheckDisputeGameImplementation() internal view {
```
mds1, line 111: We should read the DGF address from the superchain registry and require that it equals the `dgfProxy` address taken from the input

tasks/sep/fp-recovery/005-set-game-implementation/justfile
```diff
@@ -0,0 +1,23 @@
...
+# default recipe to display help information
+default:
+  @just --list
+
+# Generate the `input.json` with the game type and implementation to set.
+generate-input gameType='' implAddr='':
+  #!/usr/bin/env bash
+  set -euo pipefail
+  . .env
+  SET_GAME_TYPE_SIG="setImplementation(uint32, address)"
```
mds1, line 12: cast does normalize the signature, so this is just a nit to be a bit more defensive against cast bugs
```suggestion
  SET_GAME_TYPE_SIG="setImplementation(uint32,address)"
```

tasks/sep/fp-recovery/005-set-game-implementation/justfile
```diff
@@ -0,0 +1,23 @@
...
+  ENCODED_CALL=$(cast calldata "${SET_GAME_TYPE_SIG}" "{{gameType}}" "{{implAddr}}")
+
+  REGISTRY_DIR="../../../../lib/superchain-registry"
```
mds1, line 15: Should we just read the latest registry values directly with curl instead of requiring a local checkout? This way there's no risk the user has an outdated local version

tasks/sep/fp-recovery/005-set-game-implementation/NestedSignFromJson.s.sol
```diff
@@ -0,0 +1,188 @@
...
+        console.log("pre-check new game implementations");
+
+        if (address(currentImpl) == address(0)) {
+            return;
+        }
+        require(address(currentImpl.vm()) == address(faultDisputeGame.vm()));
+        require(address(currentImpl.weth()) == address(faultDisputeGame.weth()));
```
mds1, line 118: Can we add reason string to all require statements? Forge won't give a line number on revert, so it's a lot easier to track down with unique strings. I like to just make them incrementing numbers, with a gap large enough to add future assertions without requiring re-numbering. For example
```suggestion
        require(address(currentImpl.vm()) == address(faultDisputeGame.vm()), "10");
        require(address(currentImpl.weth()) == address(faultDisputeGame.weth()), "20");
```

tasks/sep/fp-recovery/005-set-game-implementation/NestedSignFromJson.s.sol
```diff
@@ -0,0 +1,188 @@
...
+        require(currentImpl.maxGameDepth() == faultDisputeGame.maxGameDepth());
+        require(uint64(Duration.unwrap(currentImpl.maxClockDuration())) == uint64(Duration.unwrap(faultDisputeGame.maxClockDuration())));
+        require(uint64(Duration.unwrap(currentImpl.clockExtension())) == uint64(Duration.unwrap(faultDisputeGame.clockExtension())));
+
+        if (targetGameType.raw() == GameTypes.PERMISSIONED_CANNON.raw()) {
+            PermissionedDisputeGame currentPDG = PermissionedDisputeGame(address(currentImpl));
+            PermissionedDisputeGame permissionedDisputeGame = PermissionedDisputeGame(address(faultDisputeGame));
+            require(address(currentPDG.proposer()) == address(permissionedDisputeGame.proposer()));
+            require(address(currentPDG.challenger()) == address(permissionedDisputeGame.challenger()));
+        }
```
mds1, line 131: So we are basically verifying that if a game type was already set, no configuration of that game change?

tasks/sep/fp-recovery/005-set-game-implementation/NestedSignFromJson.s.sol
```diff
@@ -0,0 +1,188 @@
...
+        allowed = new address[](5);
+        allowed[0] = address(dgfProxy);
+        allowed[1] = address(ownerSafe);
+        allowed[2] = address(securityCouncilSafe);
+        allowed[3] = address(fndSafe);
+        allowed[4] = livenessGuard;
+    }
+
+    /// @notice Checks the correctness of the deployment
+    function _postCheck(Vm.AccountAccess[] memory accesses, SimulationPayload memory /* simPayload */ )
```
mds1, line 144: This is as of https://github.com/ethereum-optimism/superchain-ops/pull/319, which was a fix for https://github.com/ethereum-optimism/superchain-ops/issues/268
```suggestion
    function _nestedPostCheck(Vm.AccountAccess[] memory accesses, SimulationPayload memory /* simPayload */ )
```

tasks/sep/fp-recovery/005-set-game-implementation/README.md
ajsutton: Good clarification. Technically this enables permissionless games (you can create and play them), it doesn't make them the respected game type in the portal.

tasks/sep/fp-recovery/005-set-game-implementation/justfile
ajsutton: We'd need to know where to find the SystemConfig or similar starting point to read the values from on-chain though and that would still come from the superchain-registry.  It may still be better though since the superchain-registry version here may not have been updated since they deployed fault proofs so wouldn't have the DisputeGameFactory address but would have SystemConfig that we could get the DGF address from.

tasks/sep/fp-recovery/005-set-game-implementation/NestedSignFromJson.s.sol
ajsutton: Yes, though mostly only because I copied this from the last upgrade. It does allow modifying the absolute prestate.

It would be worth a note at least that these assertions should be reviewed and updated to reflect any settings that are expected to change with the new implementation.

tasks/sep/fp-recovery/005-set-game-implementation/README.md
mds1: Makes sense, feel free to clarify the title however you feel best :)

tasks/sep/fp-recovery/005-set-game-implementation/justfile
mds1: Yea perhaps we just make an HTTP request with curl or similar to fetch the system config address, then we can fetch the DGF and other addresses we need from that one

tasks/sep/fp-recovery/005-set-game-implementation/README.md
ajsutton: Fixed this - it's the ProxyAdminOwner taking the action and it's not restricted to the Permissionless game. So basically nothing about the original title was right...

tasks/sep/fp-recovery/005-set-game-implementation/NestedSignFromJson.s.sol
ajsutton: I have no idea how to do that from within a solidity script. The `getGuard` method would be ideal but it's internal only.

tasks/sep/fp-recovery/005-set-game-implementation/justfile
ajsutton: I was able to load the SystemConfig address via a curl to the superchain-registry and the DisputeGameFactory from there.

I could also load the `OWNER_SAFE` from the superchain-registry `ProxyAdminOwner` field. From there I can call `getOwners()` get the address of the foundation and council safes, but I don't know which is which. The `NestedSignFromJson.sol` file can be updated to not need them, but the parent `nested.just` needs the env vars set for foundation and council safes so it knows which to use.

I've left it with `NestedSignFromJson.sol` just generically handling all the nested safe owners so it doesn't need to know the specific foundation and council addresses, and the `OWNER_SAFE` env var is populated from the superchain-registry.  We still need the `COUNCIL_SAFE` and `FOUNDATION_SAFE` env vars to make `nested.just` happy though. Not sure if we should just keep `NestedSignFromJson.sol` simpler until `nested.just` doesn't need the council and foundation safe addresses - the new code might be overkill (and it allows foundation safe owners to have no code which wasn't previously allowed).

tasks/sep/fp-recovery/005-set-game-implementation/NestedSignFromJson.s.sol
mds1: Yea, I wish the guard method was public. Form solidity this can be done with `vm.load(address target, bytes32 slot)`: https://github.com/foundry-rs/forge-std/blob/1eea5bae12ae557d589f9f0f0edae2faa47cb262/src/Vm.sol#L588-L589

tasks/sep/fp-recovery/005-set-game-implementation/NestedSignFromJson.s.sol
```diff
@@ -0,0 +1,200 @@
...
+
+    // Safe contract for this task.
+    GnosisSafe ownerSafe = GnosisSafe(payable(vm.envAddress("OWNER_SAFE")));
+    address livenessGuard = 0xc26977310bC89DAee5823C2e2a73195E85382cC7;
+
+    // Known EOAs to exclude from safety checks.
+    address systemConfigOwner; // In registry addresses.
+    address batchSenderAddress; // In registry genesis-system-configs
+    address p2pSequencerAddress; // cast call $SystemConfig "unsafeBlockSigner()(address)"
+    address batchInboxAddress; // In registry yaml.
```
mds1, line 32: I suspect the slots holding these are never touched and we can delete all these

tasks/sep/fp-recovery/005-set-game-implementation/justfile
mds1: > but I don't know which is which

The only way I can think of to infer this is with a check based the threshold and/or number of owners: The foundation is a 5 of 7, and the SC is a 10 of 13. 

`nested.just` may always need some way to know which safe is foundation and which is the council, because signers need to specify which safe they are signing for. Though now that I think of it, in theory we could take the signer address as input, determine which safe has that signer (assuming no duplicate signers across safes), and figure out which safe they need to sign with

I think what you have is ok, and perhaps we just add an assertion that `ownerSafe` has 2 (different) signers, and that each signer has code

tasks/sep/fp-recovery/005-set-game-implementation/NestedSignFromJson.s.sol
Ethnical: Agreed!

tasks/sep/fp-recovery/005-set-game-implementation/justfile
Ethnical: Interesting, I think this is bit overkill to try to remove the:  
`COUNCIL_SAFE`
`FOUNDATION_SAFE`
Even if this unlikely that the 1 signer is owner on both safe (for privacy) as the security council member are public. 
This will leak the identity of the safe owner. 
However, I do think keeping them here for the signers is with clear name like this are explicit and good to keep imo.

tasks/sep/fp-recovery/005-set-game-implementation/justfile
```diff
@@ -0,0 +1,28 @@
...
+  SET_GAME_TYPE_SIG="setImplementation(uint32,address)"
+  ENCODED_CALL=$(cast calldata "${SET_GAME_TYPE_SIG}" "{{gameType}}" "{{implAddr}}")
+
+  CONFIG_URL="https://raw.githubusercontent.com/ethereum-optimism/superchain-registry/refs/heads/main/superchain/configs/${L1_CHAIN_NAME}/${L2_CHAIN_NAME}.toml"
+  CONFIG_TOML=$(curl -s --show-error --fail "${CONFIG_URL}")
+  SC_CONFIG_ADDR=$(echo "${CONFIG_TOML}" | yq -p toml .addresses.SystemConfigProxy)
+  PROXY_ADMIN_OWNER_ADDR=$(echo "${CONFIG_TOML}" | yq -p toml .addresses.ProxyAdminOwner)
+
+  DGF_ADDR=$(cast call "${SC_CONFIG_ADDR}" "disputeGameFactory()(address)")
+  echo "OWNER_SAFE=${PROXY_ADMIN_OWNER_ADDR}" >> .env
```
Ethnical, line 21: NIT: When we launch multiples time the script this will append into the `.env` the `OWNER_SAFE`.
This shouldn't have any impact, I was thinking of a case when we run the script for the first time and wants to know how it works. 
By adding multiples times this parameters this shouldn't not have impact, should we still fix this by checking if the `OWNER_SAFE` is already written into the `.env` or this is overkill in your opinion? 
here -> https://github.com/ethereum-optimism/superchain-ops/blob/b73271f4bcb636f5a34a1857b071dce8f8fff4a5/tasks/sep/fp-recovery/005-set-game-implementation/justfile#L21


CleanShot 2024-11-01 at 14 43 33@2x

Ethnical (approved):
Seems good to me!
Great works, thanks!
There is few nits that can be discussed but overall looks great!

tasks/sep/fp-recovery/005-set-game-implementation/justfile
ajsutton: I figured it wasn't worth it given the last value is the one that gets used anyway.  The smarter approach to this in https://github.com/ethereum-optimism/superchain-ops/pull/363 makes it much less likely this will happen as there's a separate `prep` task and then multiple set-implementation calls can be added to the batch.

mds1:
Closing this since https://github.com/ethereum-optimism/superchain-ops/pull/363 says it will replace this PR, but please reopen if we still need this for any reason

PR #356, opened 2024-10-29
Update base-org/contracts dependency

mdehoog:
**Description**

We did a big cleanup of the `*MultisigBuilder` contracts in base-org/contracts#97.

This PR bumps the base-org/contracts dependency to pick up these changes, and makes all the required updates for the scripts.

**Tests**

I tested one of the tasks before and after this change to ensure that the same data-to-sign was produced, as well as the same output / assertions.

### DIFF
--- a/script/NestedSignFromJson.s.sol
+++ b/script/NestedSignFromJson.s.sol
@@ -3,6 +3,8 @@ pragma solidity ^0.8.15;
 
 import {JsonTxBuilderBase} from "src/JsonTxBuilderBase.sol";
 import {NestedMultisigBuilder} from "@base-contracts/script/universal/NestedMultisigBuilder.sol";
+import {Simulation} from "@base-contracts/script/universal/Simulation.sol";
+import {IGnosisSafe} from "@eth-optimism-bedrock/scripts/interfaces/IGnosisSafe.sol";
 import {IMulticall3} from "forge-std/interfaces/IMulticall3.sol";
 import {stdJson} from "forge-std/StdJson.sol";
 import {console} from "forge-std/console.sol";
@@ -12,13 +14,13 @@ contract NestedSignFromJson is NestedMultisigBuilder, JsonTxBuilderBase {
     address globalSignerSafe; // Hack to avoid passing signerSafe as an input to many functions.
 
     /// @dev Signs the approveHash transaction from the Nested Safe to the System Owner Safe.
-    function signJson(string memory _path, address _signerSafe) public {
+    function signJson(string memory _path, IGnosisSafe _signerSafe) public {
         _loadJson(_path);
         sign(_signerSafe);
     }
 
     /// @dev Submits signatures to call approveHash on the System Owner Safe.
-    function approveJson(string memory _path, address _signerSafe, bytes memory _signatures) public {
+    function approveJson(string memory _path, IGnosisSafe _signerSafe, bytes memory _signatures) public {
         _loadJson(_path);
         approve(_signerSafe, _signatures);
     }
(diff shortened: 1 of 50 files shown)


mdehoog:
This can be simplified with base-org/contracts#103... converting to draft until this is merged

script/NestedSignFromJson.s.sol
```diff
@@ -36,25 +36,4 @@ contract NestedSignFromJson is NestedMultisigBuilder, JsonTxBuilderBase {
...
-
-    function _postCheck(Vm.AccountAccess[] memory accesses, SimulationPayload memory simPayload)
-        internal
-        virtual
-        override
-    {
-        if (msg.sig == this.approveJson.selector) {
-            console.log("Skipping assertions on the approval call");
-            return;
-        }
```
mdehoog, line 48: the `NestedMultisigBuilder` no longer calls `_postCheck` from the approval call, so this can be cleaned up

script/NestedSignFromJson.s.sol
```diff
@@ -36,25 +36,4 @@ contract NestedSignFromJson is NestedMultisigBuilder, JsonTxBuilderBase {
...
-    }
-
-    function _nestedPostCheck(Vm.AccountAccess[] memory accesses, SimulationPayload memory simPayload)
-        internal
-        virtual
-    {
-        accesses; // Silences compiler warnings.
-        simPayload;
-        require(false, "_nestedPostCheck not implemented");
-    }
```
mdehoog, line 59: marking this contract as `abstract` achieves the same goal (sub-contracts must implement `_postCheck`)

script/NestedSignFromJson.s.sol
mds1: This makes sense, I like the new `_postCheck`, `_postSign`, `_postApprove`, `_postRun` split in `NestedMultisigBuilder`

script/PresignPauseFromJson.s.sol
```diff
@@ -19,25 +20,38 @@ contract PresignPauseFromJson is MultisigBuilder, JsonTxBuilderBase {
         allowed[1] = _ownerSafe(); // The nonce is updated in the  Foundation Operations Safe (FOS).
     }
 
-    function _addGenericOverrides() internal view virtual override returns (SimulationStateOverride memory override_) {
+    function _simulationOverrides()
```
mds1, line 23: @Ethnical Do you mind sanity checking that the new presigned pause logic in here still works? We'll need to generate new ones soon once we move the DeputyGuardianModule behind a proxy, so just want to validate this to avoid any issues at that time

mds1 (dismissed):
This looks great, and the upstream `base-org/contracts` change are really nice. Thank you for this!

> I tested one of the tasks before and after this change to ensure that the same data-to-sign was produced, as well as the same output / assertions.

Before merging, @Ethnical can you:
- Double check that tasks generate the same data to sign before and after this change
- Also verify the presigned pauses (https://github.com/ethereum-optimism/superchain-ops/pull/356/files#r1825095331)
- Review https://github.com/base-org/contracts/pull/97, just to ensure we're familiar with the changes so we know how they might help us with upcoming superchain-ops improvements

I don't expect any issues with any of the above, but just want to be sure before we merge

script/PresignPauseFromJson.s.sol
mds1: In 5f7cd28 I updated the most recent mainnet presigned pause file accordingly to serve as a template

mds1 (approved):
I've confirmed all the above tasks, so will merge

@mdehoog One thing I noticed is that the state overrides in the nested case seem to be different—we no longer update the Safe owners to be a single owner of the Multicall3 contract. Was this intentional, and if so what is the new approach for nested overrides?

mdehoog:
thanks for the review @mds1!

I didn't intentionally update the simulation overrides, but I believe the `owner=multicall` override is still present?
https://github.com/base-org/contracts/blob/12afb9f603bfe20a59dd772ae9bb9a8e68e14d45/script/universal/NestedMultisigBuilder.sol#L220

mds1:
I noticed differently tenderly URLs when comparing the `018-granite-upgrade` task on main vs. this branch: Here is a diff, the right side is the new URL: https://difff.jp/en/eyse3.html

And corresponding screenshot of the tenderly UI when loading that URL
image

The issue here might be related to the fact that the task is already executed and now effective has a no-op state diff, so I did not investigate this much

Issue #337, opened 2024-10-02
Automatically run tasks in CI if task is in a non-terminal state


mds1:
In `script/utils/check-task-statuses.sh` we define the following allowed task status:

https://github.com/ethereum-optimism/superchain-ops/blob/4a7d0f398555c9d17c83f84586db29d6743c7a96/script/utils/check-task-statuses.sh#L4

Of these, the SIGNED (for presigned pauses only), EXECUTED. and CANCELLED statuses are terminal, meaning once a task's status is changed to this value it should never change again.

When we add new tasks, the task author must manually remember to add them to CI (see `just_simulate_sc_rehearsal_1` as an example), but this is easy to forget. Instead, we should have CI do the following:
- Loop over all tasks in all `tasks/*` directories.
- Extract all tasks that have a status that is not a terminal status
- Determine if they are single safe or nested safe tasks (this can be done based on the name of the Solidity file—if it's `NestedSignFromJson` it's a nested safe task, otherwise it's a standard single safe)
- Simulate the task in CI (simulation commands differ for nested vs. standard). For nested safes, we should also simulate the approve calls, which provides an additional mitigation for https://github.com/ethereum-optimism/superchain-ops/issues/268
- The CI job should fail if any approval or simuation fails

The reason we filter tasks by "tasks in a non-terminal status" instead of "statues that precede execution" is to be more robust: If a new task status is added, by default CI will try executing tasks with those status, and we can add that task status to the terminal status list if needed

bordolot:
Hey, I'm wondering if I could do that, but I'm not sure, if I'm understanding this task correclty.

Am I right in thinking that the job is to modify ./circleci/config.yml file by adding a new job to the workflow?

The job involves listing all non-terminal tasks, and running a simulation for each, as defined in the README.md file of the task? Is that correct?

mds1:
Hey! Yes that's close. So you'll see the `script/utils/check-task-statuses.sh` already loops through all tasks to check their status. We can modify that file (and rename it accordingly, to also simulate). That file is already executed in CI in the `check_task_statuses`, so we'd modify that existing workflow based on the file rename.

Something like this diff is how I'd approach it:

```diff
diff --git a/script/utils/check-task-statuses.sh b/script/utils/check-task-statuses.sh
index 89c77c5..4494383 100644
--- a/script/utils/check-task-statuses.sh
+++ b/script/utils/check-task-statuses.sh
[...]

PR #358, opened 2024-10-29
chore: set the status to `SIGNED`.

Ethnical:
**Description**

Update the status of `base-002-fp-upgrade` to **SIGNED** for now and should be set  to **EXECUTED** after execution.

### DIFF
--- a/tasks/eth/base-002-fp-upgrade/README.md
+++ b/tasks/eth/base-002-fp-upgrade/README.md
@@ -1,6 +1,6 @@
 # Base Mainnet Fault Proofs Upgrade
 
-Status: [READY TO SIGN]
+Status: [SIGNED]
 
 ## Objective
 


Ethnical:
I will close this PR because this is not necessary and the correct flow to change the status to `SIGNED` should be only for PSPs runbooks. 
From this conversation on discord: 
Ah interesting. So `SIGNED` was intended to be a terminal state for things like presigned pause, where we sign the playbook and don't have any immediate plans to execute. For this kind of playbook (sign + execute around the same time) we typically go straight from `READY TO SIGN` to `EXECUTED`. The reason being a "terminal status" matters is because we want to use those to determine which tasks to automatically exceute in CI: https://github.com/ethereum-optimism/superchain-ops/issues/337
Mentioning this issue -> https://github.com/ethereum-optimism/superchain-ops/issues/337

Ethnical:
I will close this PR because this is not necessary and the correct flow to change the status to `SIGNED` should be only for PSPs runbooks. 
From this conversation on discord: 
Ah interesting. So `SIGNED` was intended to be a terminal state for things like presigned pause, where we sign the playbook and don't have any immediate plans to execute. For this kind of playbook (sign + execute around the same time) we typically go straight from `READY TO SIGN` to `EXECUTED`. The reason being a "terminal status" matters is because we want to use those to determine which tasks to automatically exceute in CI: https://github.com/ethereum-optimism/superchain-ops/issues/337
Mentioning this issue -> https://github.com/ethereum-optimism/superchain-ops/issues/337

PR #363, opened 2024-10-31
Chain-independent script to generate setImplementation operations

ajsutton:
**Description**

Adds a template to the fp-recovery directory with tools to create operations that set the implementation for fault dispute games.  Multiple implementations can be set in a single batch (e.g. setting CANNON and PERMISSIONED game type as part of a hard fork upgrade.

This would replace #354 - it generates essentially the same task but in a much more automated way and with the ability to make multiple changes in a single operation.


tasks/sep/fp-recovery/005-set-game-implementation/templates/NestedSignFromJson.s.sol
```diff
@@ -0,0 +1,170 @@
...
+contract NestedSignFromJson is OriginalNestedSignFromJson {
+    using LibString for string;
+
+    // Chains for this task.
+    string l1ChainName = vm.envString("L1_CHAIN_NAME");
+    string l2ChainName = vm.envString("L2_CHAIN_NAME");
+
+    // Safe contract for this task.
+    GnosisSafe ownerSafe = GnosisSafe(payable(vm.envAddress("OWNER_SAFE")));
+    address livenessGuard = 0xc26977310bC89DAee5823C2e2a73195E85382cC7;
```
ajsutton, line 27: I suspect this is only correct for sepolia. In theory we can read this from the slot (https://github.com/ethereum-optimism/superchain-ops/pull/354#discussion_r1819854077) but I'm not sure how to actually do that.

mds1: Just replied in that other thread with info on how to do that!

tasks/sep/fp-recovery/005-set-game-implementation/justfile
```diff
@@ -0,0 +1,91 @@
...
+  @just --list
+
+clean:
+  rm -rf out
+
+prep l1='' l2='':
+  #!/usr/bin/env bash
+  set -euo pipefail
+  mkdir out
+  sed -e 's/<L2_CHAIN>/{{l2}}/g' -e 's/<L1_CHAIN>/{{l1}}/g' templates/README.md > out/README.md
```
mds1, line 14: Will this work cross platform? I know sed has some linux vs macOS differences but offhand I can't remember what they were

tasks/sep/fp-recovery/005-set-game-implementation/justfile
```diff
@@ -0,0 +1,91 @@
...
+  FOUNDATION_SAFE=0xDEe57160aAfCF04c34C887B5962D0a69676d3C8B
+  EOF
+  elif [[ "{{l1}}" == "mainnet" ]]; then
+    cat > out/.env << EOF
+  ETH_RPC_URL="https://ethereum.publicnode.com"
+  COUNCIL_SAFE=0xc2819DC788505Aac350142A7A707BF9D03E3Bd03
+  FOUNDATION_SAFE=0x847B5c174615B1B7fDF770882256e2D3E95b9D92
+  EOF
+  else
+    echo "Unknown L1 chain."
```
mds1, line 32: UX nit
```suggestion
    echo "Unknown L1 chain. Must be 'mainnet' or 'sepolia'"
```

tasks/sep/fp-recovery/005-set-game-implementation/justfile
```diff
@@ -0,0 +1,91 @@
...
+  SYSTEM_CONFIG="${SYSTEM_CONFIG}"
+  OWNER_SAFE="${OWNER_SAFE}"
+  EOF
+
+# Generate the `input.json` with the game type and implementation to set.
+set-implementation gameType='' implAddr='':
+  #!/usr/bin/env bash
+  set -euo pipefail
+  . out/.env
+  export ETH_RPC_URL # Make available to cast since the .env file is in the out subdir.
```
mds1, line 54: A lot of peolpe define an `ETH_RPC_URL` in their zshrc etc so it's always avaialble to cast by default. I've had issues in superchain-ops where the global one gets preferred, so this is just to be sure any default `ETH_RPC_URL` is not used
```suggestion
  unset ETH_RPC_URL 
  . out/.env
  export ETH_RPC_URL # Make available to cast since the .env file is in the out subdir.
```

tasks/sep/fp-recovery/005-set-game-implementation/justfile
```diff
@@ -0,0 +1,91 @@
...
+                (.data = \"$ENCODED_CALL\") |
+                (.to = \"${DGF_ADDR}\") |
+                (.contractInputsValues._gameType = \"{{gameType}}\") |
+                (.contractInputsValues._impl = \"{{implAddr}}\")" ./templates/tx/setImplementation.json)
+  jq --argjson tx "${TX}" --arg desc "${TX_DESC} " '(.metadata.description += $desc) | (.transactions += [$tx])' ./out/input.json > ./out/input.tmp.json
+  mv ./out/input.tmp.json ./out/input.json
+
+  # Add a precheck for this implementation change
+  sed -I "" -e 's|// INSERT NEW PRE CHECKS HERE|_precheckDisputeGameImplementation(GameType.wrap({{gameType}}), {{implAddr}});\n        // INSERT NEW PRE CHECKS HERE|' out/NestedSignFromJson.s.sol
+  sed -I "" -e 's|// INSERT NEW POST CHECKS HERE|_checkDisputeGameImplementation(GameType.wrap({{gameType}}), {{implAddr}});\n        // INSERT NEW POST CHECKS HERE|' out/NestedSignFromJson.s.sol
```
mds1, line 71: same cross-platform `sed` question

tasks/sep/fp-recovery/005-set-game-implementation/justfile
```diff
@@ -0,0 +1,91 @@
...
+  cat >> out/.env << EOF
+  SAFE_NONCE=""
+  L1_CHAIN_NAME="{{l1}}"
+  L2_CHAIN_NAME="{{l2}}"
+  SYSTEM_CONFIG="${SYSTEM_CONFIG}"
+  OWNER_SAFE="${OWNER_SAFE}"
+  EOF
+
+# Generate the `input.json` with the game type and implementation to set.
+set-implementation gameType='' implAddr='':
```
mds1, line 50: Should we do some basic validation on `implAddr` to verify that it is a fault dispute game, e.g. make a single call that should either not revert, or return some known value

tasks/sep/fp-recovery/005-set-game-implementation/templates/NestedSignFromJson.s.sol
```diff
@@ -0,0 +1,170 @@
...
+        allowed[0] = address(dgfProxy);
+        allowed[1] = address(ownerSafe);
+        allowed[2] = livenessGuard;
+        for (uint256 i = 0; i < nestedSafes.length; i++) {
+            allowed[3 + i] = nestedSafes[i];
+        }
+    }
+
+    /// @notice Checks the correctness of the deployment
+    function _nestedPostCheck(Vm.AccountAccess[] memory accesses, SimulationPayload memory /* simPayload */ )
```
mds1, line 143: Note this the dep update in https://github.com/ethereum-optimism/superchain-ops/pull/356 will require a few small changes to this script if #356 is merged first (see NestedSignFromJson diffs in that PR for an example)

tasks/sep/fp-recovery/005-set-game-implementation/justfile
ajsutton: The `NestedSignFromJson` checks call quite a range of methods on this contract. I've generally deferred most validation to there rather than trying to do it in bash, especially given that regenerating the task from scratch is so easy.

tasks/sep/fp-recovery/005-set-game-implementation/justfile
ajsutton: Yeah good call.  I tested it on linux this one works and is pretty basic sed syntax I've been using for years.  The ones below did need adjusting because the -I option wasn't on linux (just -i).

tasks/sep/fp-recovery/005-set-game-implementation/justfile
ajsutton: Yeah this one had to be adjusted - I couldn't remember if `-I` or `-i` was the standard option - turns out it was the lowercase one.

tasks/sep/fp-recovery/005-set-game-implementation/templates/NestedSignFromJson.s.sol
ajsutton: That got it working thanks.  I've written it so that it will allow a liveness guard from any of the nested safes to be accessed (and it detects when there isn't one so it doesn't add 0x00 to the allow list).

tasks/sep/fp-recovery/005-set-game-implementation/justfile
ajsutton: Upon further testing there's differences in which temporary files get left around so I gave up on having sed edit in place.

ajsutton:
Tested this on mainnet and it works with a simulation that looks right now that the livenessGuard is automatically determined.

tasks/sep/fp-recovery/005-set-game-implementation/templates/NestedSignFromJson.s.sol
mds1: Ok, I just merged https://github.com/ethereum-optimism/superchain-ops/pull/356 first, so we'll have to update this PR accordingly

mds1 (dismissed):
Will need to rebase against the latest main and re-tested, but once that is done feel free to resolve https://github.com/ethereum-optimism/superchain-ops/pull/363/files#r1828471807 and merge

ajsutton:
@mdehoog Could you help me with updating this one to handle the changes in https://github.com/ethereum-optimism/superchain-ops/pull/356 please? Getting the code to compile is easy, but now that it's rebased when I call `getOwners` on `0x1Eb2fFc903729a0F03966B917003800b145F56E2` (on sepolia L1) which should be the 2-of-2, it only gives me 1 address.  The actual contract on L1 still has two owners but it seems the simulation is now removing the council safe as an owner.

It seems when `SIMULATE_WITHOUT_LEDGER=true` is set, the owners of the 2-of-2 is set to the first owner account of the council safe.  But the council safe is still accessed somehow as part of the execution so it fails with "Unallowed Storage access". Was that deliberate? It seems like a weird configuration to be trying to simulate with.

ajsutton:
To reproduce the problem I'm hitting run:
```
just clean prep sepolia op set-implementation 0 0xaA511096675013995A2D58275af6a8880AC39F44
cd out
SIMULATE_WITHOUT_LEDGER=true just --dotenv-path $(pwd)/.env --justfile ../../../../../nested.just simulate council
```

ajsutton:
I get the same problem without `SIMULATE_WITHOUT_LEDGER=true` as well and when using the foundation safe.

ajsutton:
Worked around this by going back to using the hard coded council and foundation safe addresses (https://github.com/ethereum-optimism/superchain-ops/pull/363/commits/aa9543e9f7e4d81141f7b2c44b8beb18c01a6a3b) but it's a shame to have to embed that into the script as it means the script itself will have to be adjusted to deal with 3-of-3 safes and other variants.

Overriding the safe owners still seems kind of suspicious though, especially since it doesn't seem to completely eliminate the original owner safes.

mdehoog:
@ajsutton bug fix is here: base-org/contracts#105, apologies for the regression.

ajsutton:
Oh nice. Thanks.

mds1 (approved):

PR #365, opened 2024-11-01
Add ability to re-initialize AnchorStateRegistry with additional game types

ajsutton:
**Description**

Adds an additional transaction template to https://github.com/ethereum-optimism/superchain-ops/pull/363 that re-initializes the `AnchorStateRegistry` to add an initial anchor state for new game types.  The initial anchor state to use is taken from the current anchor state for another game type (at the time the `just` command is run so it's static in the resulting task definition).

This allows creating a task to enable permissionless fault proofs on a chain that previously only deployed permissioned games.

### DIFF
--- a/tasks/sep/fp-recovery/005-set-game-implementation/justfile
+++ b/tasks/sep/fp-recovery/005-set-game-implementation/justfile
@@ -37,6 +37,7 @@ prep l1='' l2='':
   CONFIG_TOML=$(curl -s --show-error --fail "${CONFIG_URL}")
   SYSTEM_CONFIG=$(echo "${CONFIG_TOML}" | yq -p toml .addresses.SystemConfigProxy)
   OWNER_SAFE=$(echo "${CONFIG_TOML}" | yq -p toml .addresses.ProxyAdminOwner)
+  PROXY_ADMIN=$(echo "${CONFIG_TOML}" | yq -p toml .addresses.ProxyAdmin)
 
   cat >> out/.env << EOF
   SAFE_NONCE=""
@@ -44,6 +45,7 @@ prep l1='' l2='':
   L2_CHAIN_NAME="{{l2}}"
   SYSTEM_CONFIG="${SYSTEM_CONFIG}"
   OWNER_SAFE="${OWNER_SAFE}"
+  PROXY_ADMIN="${PROXY_ADMIN}"
   EOF
 
 # Generate the `input.json` with the game type and implementation to set.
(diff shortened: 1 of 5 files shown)


tasks/sep/fp-recovery/005-set-game-implementation/justfile
```diff
@@ -94,3 +96,107 @@ set-implementation gameType='' implAddr='':
   INFO="* Set implementation for game type {{gameType}} to {{implAddr}} in \`DisputeGameFactory\` ${DGF_ADDR}: \`setImplementation({{gameType}}, {{implAddr}})\`"
   sed -e "s|<!--NEXT TASK DESCRIPTION-->|${INFO}\n<!--NEXT TASK DESCRIPTION-->|" out/README.md > out/README.md.tmp
   mv out/README.md.tmp out/README.md
+
+copy-anchor-state fromGameType='' toGameTypes='':
+  #!/usr/bin/env bash
+  set -euo pipefail
```
mds1, line 102: similar to above task
```suggestion
  set -euo pipefail
  unset ETH_RPC_URL
```

tasks/sep/fp-recovery/005-set-game-implementation/justfile
```diff
@@ -94,3 +96,107 @@ set-implementation gameType='' implAddr='':
...
+  DGF_ADDR=$(cast call "${SYSTEM_CONFIG}" "disputeGameFactory()(address)")
+
+  # Find the anchor state registry for the source game.
+  SOURCE_IMPL=$(cast call "${DGF_ADDR}" 'gameImpls(uint32)(address)' '{{fromGameType}}')
+  REGISTRY_PROXY=$(cast call "${SOURCE_IMPL}" 'anchorStateRegistry()(address)' "${SOURCE_IMPL}")
+  REGISTRY_IMPL=$(cast call "${REGISTRY_PROXY}" 'implementation()(address)')
+  SUPERCHAIN_CONFIG=$(cast call "${REGISTRY_PROXY}" 'superchainConfig()(address)')
+
+  # Load the current anchor state
+  ANCHOR_JSON=$(cast call -j "${REGISTRY_PROXY}" 'anchors(uint32)(bytes32,uint256)' '{{fromGameType}}')
```
mds1, line 115: Looks like this `--json` arg is only on foundry versions newer then what we specify in [`versions.json`](https://github.com/ethereum-optimism/optimism/blob/94ab846ab439ef26802ca32d6bc5921aadbafa0f/versions.json#L4)—I had to upgrade from the monorepo version to latest for this to work

mds1 (approved):
Similar to #363, just needs to be updated now that #356 was merged

ajsutton:
@mds1 Just a heads up, I wound up changing this one a bit.  Because the `initialize` method writes the supplied anchor states to it's storage but doesn't delete any other existing entries from the map, we can make the operation safer by only setting the new anchor states and not overwriting the existing source anchor state.  I added a post-check to confirm the source anchor state is still set.

tasks/sep/fp-recovery/005-set-game-implementation/justfile
ajsutton: hmm, `cast install-foundry` in monorepo gives me `cast 0.2.0 (143abd6 2024-09-04T00:24:42.542256000Z)` which does have the `-j` option for call and works for me.

mds1: ah maybe I was just on an even older version then (I've already upgraded, so can't confirm)

mds1 (approved):

PR #366, opened 2024-11-04
feat: sepolia OPCM admin swap

smartcontracts:
Adds a new task for swapping the OPCM proxy admin to be the ProxyAdmin contract for OP Sepolia. We decided that using the same ProxyAdmin everywhere was better than having a single contract where the owner was the Fake Foundation instead.

### DIFF
--- a/tasks/sep/018-opcm-admin-swap/README.md
+++ b/tasks/sep/018-opcm-admin-swap/README.md
@@ -1,0 +1,31 @@
+# Sepolia OPCM Admin Swap
+
+Status: DRAFT, NOT READY TO SIGN
+
+## Objective
+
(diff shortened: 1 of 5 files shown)


tasks/sep/018-opcm-admin-swap/README.md
```diff
@@ -0,0 +1,31 @@
+# Sepolia OPCM Admin Swap
+
+Status: DRAFT, NOT READY TO SIGN
+
+## Objective
+
+Changes the admin/owner address of the `OPContractsManager` proxy contract on Sepolia to be the
+`ProxyAdmin` contract on Sepolia instead of the 2/2 Safe. This configuration matches the setup on
```
mds1, line 8: Currently it should just be the foundation safe, not the 2/2

PR #367, opened 2024-11-05
Bump base-org/contracts dep

mdehoog:
**Description**

Bumps the `base-org/contracts` dep to pull in base-org/contracts#105, which fixes some issues we were having in #363.


ajsutton (approved):
Thank you. :) (but my approval doesn't count here...)

mds1 (approved):
Thank you!

PR #369, opened 2024-11-13
feat(sep-dev-0): Holocene upgrade task

clabby:
## Overview

Adds a task for upgrading the internal devnet to the Holocene smart contracts.

For simplicity in our operations, the `SystemConfig` and proofs contract upgrades are combined into this single task. This task is to be executed _prior_ to fork activation, to ensure that the `FaultDisputeGame` and `PermissionedDisputeGame` have a Holocene-compatible prestate.

Because we are upgrading the `SystemConfig` prior to fork activation, it is critical that the `SystemConfig` owner does not call `setEip1559Params` prior to the fork activation.


ajsutton:
I suspect we should upgrade devnet the same way we plan to do for testnet and mainnet so we can confirm that things work if we _don't_ upgrade SystemConfig ahead of time. I don't think there's anything else that would have tested that case to be honest since all the e2e tests use the latest versions of contracts.

tasks/sep-dev-0/005-holocene-proofs/.env
```diff
@@ -0,0 +1,4 @@
+ETH_RPC_URL="https://ethereum-sepolia.publicnode.com"
+SUPERCHAIN_CONFIG_ADDR=0xC2Be75506d5724086DEB7245bd260Cc9753911Be
```
ajsutton, line 2: This doesn't match the value in the superchain registry: https://github.com/ethereum-optimism/superchain-registry/blob/c8ef19dc6663aa337b72d2191abbc03681379b3b/superchain/configs/sepolia-dev-0/superchain.toml#L3
This looks like the sepolia testnet address.

tasks/sep-dev-0/006-holocene-sys-cfg/.env
```diff
@@ -0,0 +1,4 @@
+ETH_RPC_URL="https://ethereum-sepolia.publicnode.com"
+SUPERCHAIN_CONFIG_ADDR=0xC2Be75506d5724086DEB7245bd260Cc9753911Be
```
ajsutton, line 2: Also the wrong address but doesn't seem to be used in either case so can just be removed.

tasks/sep-dev-0/006-holocene-sys-cfg/README.md
```diff
@@ -0,0 +1,31 @@
...
+## Objective
+
+Upgrades the `SystemConfig` contract for the Holocene hardfork.
+
+Contracts upgraded are included within the
+[op-contracts/v1.8.0-rc.1](https://github.com/ethereum-optimism/optimism/tree/op-contracts/v1.8.0-rc.1) release.
+
+## Pre-deployments
+
+- `SystemConfig` - `0x29d06Ed7105c7552EFD9f29f3e0d250e5df412CD`
```
ajsutton, line 14: We're actually skipping quite a few versions of `SystemConfig` with this upgrade.  It seems the devnet hasn't been updated in the past.  So I think we will need to add a reinitialize step to the upgrade for devnet so it sets things like the dispute game factory address which it doesn't currently have. The current version appears to be pre the MCP upgrade.

ajsutton (commented):
Generally looks good and the new fault dispute game implementations are working as expected.  The SystemConfig is very out of date so I suspect the upgrade on devnet for that will be a bit more complex and need an initialize() call.

tasks/sep-dev-0/005-holocene-proofs/OVERVIEW.md
```diff
@@ -0,0 +1,43 @@
...
+
+**Value:** `0 WEI`
+
+**Raw Input Data:** `0x14f6b1a300000000000000000000000000000000000000000000000000000000000000010000000000000000000000006a962628aa48564b7c48d97e1a738044ffec686f`
+
+### Inputs
+
+**_impl:** `0x6A962628Aa48564B7C48D97E1A738044fFEc686F`
+
+**_gameType:** `1`
```
sebastianst, line 23: nit: wrong order (and tx2 has correct order)

tasks/sep-dev-0/006-holocene-sys-cfg/SignFromJson.s.sol
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    function _postCheck(Vm.AccountAccess[] memory accesses, Simulation.Payload memory) internal override {
+        console.log("Running post-deploy assertions");
+        checkStateDiff(accesses);
+        checkSystemConfigProxy();
+        console.log("All assertions passed!");
+    }
+
+    function checkSystemConfigProxy() internal {
```
sebastianst, line 62: could we add more assertions to make sure the new `SystemConfig` is indeed correctly initialized etc given that we're skipping over a few intermediate updates with this one?

tasks/sep-dev-0/006-holocene-sys-cfg/OVERVIEW.md
```diff
@@ -0,0 +1,21 @@
...
+**Function Signature:** `upgrade(address,address)`
+
+**To:** `0x18d890A46A3556e7F36f28C79F6157BC7a59f867`
+
+**Value:** `0 WEI`
+
+**Raw Input Data:** `0x99a88ec4000000000000000000000000a6b72407e2dc9ebf84b839b69a24c88929cf20f700000000000000000000000029d06ed7105c7552efd9f29f3e0d250e5df412cd`
+
+### Inputs
+**_implementation:** `0x29d06Ed7105c7552EFD9f29f3e0d250e5df412CD`
```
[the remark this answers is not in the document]
geoknee: I don't think we do for devnet, since the registry does not store a standard `SystemConfig` implementation  address for devnets (they aren't validated in this way). When we get to testnet, then yes we will need to add to the registry.

tasks/sep-dev-0/006-holocene-sys-cfg/README.md
clabby: Updated to set all of the MCP slots. The fee scalar slots and EIP-1559 parameter slots can be left alone, the old values will be used.

tasks/sep-dev-0/006-holocene-sys-cfg/OVERVIEW.md
```diff
@@ -0,0 +1,170 @@
...
+
+**Raw Input Data:** `0xca446dd952322a25d9f59ea17656545543306b7aef62bc0cc53a0e65ccfa0c75b97aa9060000000000000000000000002419423c72998eb1c6c15a235de2f112f8e38eff`
+
+### Inputs
+**_slot:** `0x52322a25d9f59ea17656545543306b7aef62bc0cc53a0e65ccfa0c75b97aa906`
+
+**_addr:** `0x2419423C72998eb1c6c15A235de2f112f8E38efF`
+
+
+## Tx #10: Upgrade `SystemConfig` proxy
```
mds1, line 155: Why are we manually setting all the slots instead of calling the `initialize` method? Is this because on sepolia/mainnet we just will use `upgrade` and not `upgradeAndCall`, so we want to preserve that exact call here and then just manually do what the initializer does to get the devnet up to date?

tasks/sep-dev-0/006-holocene-sys-cfg/README.md
```diff
@@ -0,0 +1,32 @@
...
+
```
mds1, line 15: What's the source of truth for this value, i.e. how can I verify this is correct?

tasks/sep-dev-0/006-holocene-sys-cfg/SignFromJson.s.sol
```diff
@@ -0,0 +1,97 @@
...
+import {Simulation} from "@base-contracts/script/universal/Simulation.sol";
+import {Types} from "@eth-optimism-bedrock/scripts/Types.sol";
+import {console2 as console} from "forge-std/console2.sol";
+import {stdJson} from "forge-std/StdJson.sol";
+import {LibString} from "solady/utils/LibString.sol";
+import {Vm, VmSafe} from "forge-std/Vm.sol";
+import "@eth-optimism-bedrock/src/dispute/lib/Types.sol";
+import {DisputeGameFactory} from "@eth-optimism-bedrock/src/dispute/DisputeGameFactory.sol";
+import {FaultDisputeGame} from "@eth-optimism-bedrock/src/dispute/FaultDisputeGame.sol";
+import {PermissionedDisputeGame} from "@eth-optimism-bedrock/src/dispute/PermissionedDisputeGame.sol";
```
mds1, line 14: nit, unused imports
```suggestion
import {Vm, VmSafe} from "forge-std/Vm.sol";
import {DisputeGameFactory} from "@eth-optimism-bedrock/src/dispute/DisputeGameFactory.sol";
import {FaultDisputeGame} from "@eth-optimism-bedrock/src/dispute/FaultDisputeGame.sol";
import {PermissionedDisputeGame} from "@eth-optimism-bedrock/src/dispute/PermissionedDisputeGame.sol";
```

tasks/sep-dev-0/006-holocene-sys-cfg/SignFromJson.s.sol
```diff
@@ -0,0 +1,97 @@
...
+import {SystemConfig} from "@eth-optimism-bedrock/src/L1/SystemConfig.sol";
+import {Proxy} from "@eth-optimism-bedrock/src/universal/Proxy.sol";
+
+contract NestedSignFromJson is OriginalSignFromJson {
+    using LibString for string;
+
+    // Safe contract for this task.
+    address immutable proxyAdminOwnerSafe = vm.envAddress("OWNER_SAFE");
+    address immutable livenessGuard = 0x24424336F04440b1c28685a38303aC33C9D14a25;
```
mds1, line 23: Let's read the guard from the storage slot instead of hardcoding the address. We almost ceratinly don't have this installed on devnet anyway, but I'm ok keeping it to minimize the diff between devnet and prod

You can see Adrian's use of the slot to fetch the guard here: https://github.com/ethereum-optimism/superchain-ops/blob/5bfc9ab82d4aacb854cacdb419b2862676fc7928/tasks/sep/fp-recovery/005-set-game-implementation/templates/NestedSignFromJson.s.sol#L33

tasks/sep-dev-0/006-holocene-sys-cfg/SignFromJson.s.sol
```diff
@@ -0,0 +1,97 @@
...
+        systemConfig = SystemConfig(stdJson.readAddress(inputJson, "$.transactions[9].contractInputsValues._proxy"));
+
+        l1XDMProxy = stdJson.readAddress(inputJson, "$.transactions[1].contractInputsValues._addr");
+        l1ERC721BridgeProxy = stdJson.readAddress(inputJson, "$.transactions[2].contractInputsValues._addr");
+        l1StandardBridgeProxy = stdJson.readAddress(inputJson, "$.transactions[3].contractInputsValues._addr");
+        optimismPortalProxy = stdJson.readAddress(inputJson, "$.transactions[4].contractInputsValues._addr");
+        optimismMintableERC20FactoryProxy = stdJson.readAddress(inputJson, "$.transactions[5].contractInputsValues._addr");
+        batchInbox = stdJson.readAddress(inputJson, "$.transactions[6].contractInputsValues._addr");
+        startBlock = stdJson.readUint(inputJson, "$.transactions[7].contractInputsValues._block");
+        dgfProxy = stdJson.readAddress(inputJson, "$.transactions[8].contractInputsValues._addr");
```
mds1, line 59: We should be using the superchain registry as the source of truth for these addresses, since that's how we can validate the addresses in this JSON file are correct

tasks/sep-dev-0/006-holocene-sys-cfg/OVERVIEW.md
clabby: It's easier than collecting all of the other configuration values that `initialize` requires. It's huge, requires the `ResourceConfig`, etc.

Doing this just allows a lot less verbosity / configuration.

tasks/sep-dev-0/006-holocene-sys-cfg/README.md
clabby: It was deployed prior to the upgrade. We don't update the addresses in `superchain-registry` until we've done the upgrade.

You can check the [code page](https://sepolia.etherscan.io/address/0x29d06Ed7105c7552EFD9f29f3e0d250e5df412CD#code) on etherscan to see the contract.

tasks/sep-dev-0/006-holocene-sys-cfg/SignFromJson.s.sol
```diff
@@ -0,0 +1,111 @@
...
+        proxyAdmin = readContractAddress("ProxyAdmin", l2ChainId);
+        systemConfigImpl = stdJson.readAddress(inputJson, "$.transactions[9].contractInputsValues._implementation");
+        systemConfig = SystemConfig(readContractAddress("SystemConfigProxy", l2ChainId));
+
+        l1XDMProxy = readContractAddress("L1CrossDomainMessengerProxy", l2ChainId);
+        l1ERC721BridgeProxy = readContractAddress("L1ERC721BridgeProxy", l2ChainId);
+        l1StandardBridgeProxy = readContractAddress("L1StandardBridgeProxy", l2ChainId);
+        optimismPortalProxy = readContractAddress("OptimismPortalProxy", l2ChainId);
+        optimismMintableERC20FactoryProxy = readContractAddress("OptimismMintableERC20FactoryProxy", l2ChainId);
+        batchInbox = stdJson.readAddress(inputJson, "$.transactions[6].contractInputsValues._addr");
```
mds1, line 57: This is also in the superchain registry: https://github.com/ethereum-optimism/superchain-registry/blob/a560627d56d9b12c7c649e782110101a9579a2c4/superchain/configs/sepolia-dev-0/oplabs-devnet-0.toml#L8

mds1 (approved):
LGTM pending https://github.com/ethereum-optimism/superchain-ops/pull/369#discussion_r1846927405

mds1 (approved):

PR #370, opened 2024-11-18
3-of-3 Proxy Admin Owner Support: Chain governor able to sign for nested safes

blmalone:
**Description**

As per the title, we want to add the ability for the L1 ProxyAdmin owner to be a 3-of-3 safe as well as a 2-of-2 safe. Some chains require an additional wallet in their L1PAO.

### DIFF
--- a/nested.just
+++ b/nested.just
@@ -9,6 +9,7 @@ export taskPath := invocation_directory()
 export councilSafe := env_var("COUNCIL_SAFE")
 export foundationSafe := env_var("FOUNDATION_SAFE")
 export ownerSafe := env_var('OWNER_SAFE')
+export chainGovernorSafe := env_var('CHAIN_GOVERNOR_SAFE')
 export randomPersonEoa := "0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045"
 
 simulate whichSafe hdPath='0':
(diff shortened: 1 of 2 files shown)


nested.just
```diff
@@ -9,6 +9,7 @@ export taskPath := invocation_directory()
 export councilSafe := env_var("COUNCIL_SAFE")
 export foundationSafe := env_var("FOUNDATION_SAFE")
 export ownerSafe := env_var('OWNER_SAFE')
+export chainGovernorSafe := env_var('CHAIN_GOVERNOR_SAFE')
```
mds1, line 12: Then in the justfile we should error if they pass `chain-governor` as the safe kind but don't specify the env var
```suggestion
export chainGovernorSafe := env_var_or_default('CHAIN_GOVERNOR_SAFE', '')
```

blmalone: sg, this is an example of it failing when a user passes chain-governor without setting the var.
```
blaine@blaines-MacBook-Pro 018-granite-upgrade % just --dotenv-path $(pwd)/.env --justfile ../../../nested.just simulate chain-governor            
Running script with assertions
Using script /Users/blaine/code/superchain-ops/tasks/eth/018-granite-upgrade/NestedSignFromJson.s.sol
getting signer address for chain-governor...
Error: CHAIN_GOVERNOR_SAFE is not set for chain-governor.
```

this was implemented in commit: https://github.com/ethereum-optimism/superchain-ops/pull/370/commits/23cac55edf801b8c8684f190da0c6025bc0c7732

PR #381, opened 2024-11-29
Fix `ApprovalHash` Incorrect Storage of task `020-holocene-system-config-upgrade-multi-chain`

Ethnical:
**Description**

 @geoknee Discover and Inform me of an inconsistency between the `approveHash` storage calculation from the task `020-holocene-system-config-upgrade-multi-chain`. 

After investigation, I missed the change of `nestedHash` value (so the value were calculated on the incorrect value but were matching the `VALIDATION.md` file that has incorrect values...).

Now, the value are matching the simulation correctly for example for security council below:

**State Precomputed**
ddda62b07befa8a55ccccfc521a57485360a393dbf0a4e2d93d1c3d4870d41f2

**Tenderly Simulation**
d607c91ff674481c6e8707e9f48b144be8fa47e07c2eee69ad675726e5d484ec

### DIFF
--- a/tasks/sep/020-holocene-system-config-upgrade-multi-chain/VALIDATION.md
+++ b/tasks/sep/020-holocene-system-config-upgrade-multi-chain/VALIDATION.md
@@ -90,14 +90,14 @@ And we do indeed see these entries:
 
   ### For the Council:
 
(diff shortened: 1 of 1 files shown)


geoknee (approved):

raffaele-oplabs (approved):
Approved as a codeowner to unblock. I trust @geoknee review who have already approved this PR.

PR #383, opened 2024-12-02
Swap `3` Previous Owners with `3` new Owners on the `FakeSecurityCouncil`. 

Ethnical:
**Description**

This PR will swap 3 old owners of the FakeSC on Sepolia using the function `swapOwners()` to remove 3 old signers and include 3 new signers  (See the `README.md`). 
To do so, we created the task `021-refresh-signers-forfakeSC-fakeFND` under `sep` folder. 

**Recap** 
Using `swapOwners()`:
This action will perform a swapping of the previous 3 owners that was not active anymore by the 3 new owners.

| PreviousOwner | newOwner |
| --- | --- |
| 0xad70Ad7Ac30Cee75EB9638D377EACD8DfDfE0C3c | 0x95E774787A63f145f7B05028a1479bDc9D055f3d |
| 0xE09d881A1A13C805ED2c6823f0C7E4443A260f2f | 0x41fb1d8c3262e88a056ee3099f5718405CC8cAdE |
| 0x78339d822c23d943e4a2d4c3dd5408f66e6d662d | 0xa03dafade71f1544f4b0120145eec9b89105951f |

**Tests**
I have run the simulation under tenderly and locally to make sure this is the desired state changes: 
**Tenderly:**
CleanShot 2024-12-03 at 17 08 19@2x

**Forge superchain-ops script**
CleanShot 2024-12-03 at 17 10 08@2x

**Additional context**

 This task is changing the number of the owner of the FakeSC but this shouldn't cause any liveness fallback activated (I don't know which invariant to check after the execution into the `_postCheck()` happy to implement any proposed solution here).  

**Metadata**
- Fixes https://github.com/ethereum-optimism/safe-wallets/pull/13

### DIFF
--- a/tasks/sep/024-refresh-signers-fakeSC/SignFromJson.s.sol
+++ b/tasks/sep/024-refresh-signers-fakeSC/SignFromJson.s.sol
@@ -107,7 +107,7 @@ contract SignFromJson is OriginalSignFromJson {
         }
         require(
             numberOwners == securityCouncilSafeOwners.length,
-            "The number of owner should have not be increased."
+            "The number of owners should not have been increased."
         );
 
         checkStateDiff(accesses);


tasks/sep/021-refresh-signers-fakeSC/.env
mds1, line 1: The description says:
> This task is changing the number of the owner of the FakeSC but this shouldn't cause any liveness fallback activated (I don't know which invariant to check after the execution into the _postCheck() happy to implement any proposed solution here).

But it seems this doesn't actually change the total number of signers, it just swaps them right?

Regarding the postCheck, that is a good idea. The specs might have some operational info on how to swap owners and what we should check: https://specs.optimism.io/protocol/safe-extensions.html#security-council-liveness-checking-extensions. 

I think we will need all new signers to show liveness right after the swap, since their `lastLive` time would be zero by default. @maurelian does that sound right?

tasks/sep/021-refresh-signers-fakeSC/README.md
```diff
@@ -0,0 +1,30 @@
+# Sepolia Update Signers `FakeSC` and `FakeFoS` and `FakeFuS`
+
+Status: [READY TO SIGN]
+
+## Objective
+
+This task **update** the signers of the `FakeSC`(`0xf64bc17485f0B4Ea5F06A96514182FC4cB561977`) and `FakeFuS`(`0xDEe57160aAfCF04c34C887B5962D0a69676d3C8B`) and `FakeFoS`(`0x837DE453AD5F21E89771e3c06239d8236c0EFd5E`) on Sepolia.
```
mds1, line 7: Seems this is not right, we only update it for one safe here?

tasks/sep/021-refresh-signers-fakeSC/SignFromJson.s.sol
```diff
@@ -0,0 +1,84 @@
...
+import {ModuleManager} from "safe-contracts/base/ModuleManager.sol";
+
+contract SignFromJson is OriginalSignFromJson {
+    using LibString for string;
+
+    // Safe contract for this task.
+    GnosisSafe securityCouncilSafe =
+        GnosisSafe(payable(vm.envAddress("OWNER_SAFE"))); // We take from the "OWNER_SAFE" as this is the "TARGET_SAFE".
+
+    // TODO: Get the livenessGuard from the SC for not hardcoding the address.
```
mds1, line 23: Yep, let's address this todo, you can see in other tasks how we read it from this slot (like the fp-recovery tasks)

tasks/sep/021-refresh-signers-fakeSC/.env
maurelian: Yes, I do think that'll be necessary.

maurelian: Actually, no, the LivenessGuard handles adding new signers with the current lastLive time in the checkAfterExecution() function

tasks/sep/021-refresh-signers-fakeSC/README.md
Ethnical: You are correct, I thought I forgot to update this task after multiples copy paste. 
Change addressed here.

tasks/sep/021-refresh-signers-fakeSC/.env
Ethnical: > But it seems this doesn't actually change the total number of signers, it just swaps them right?

That's correct. 

> I think we will need all new signers to show liveness right after the swap, since their lastLive time would be zero by default. @maurelian does that sound right?

This is not required @mds1, on the PR there is the screenshot (below) of the tenderly simulation. You can see that previous Owners are set to `0` and the new one are set with the  current value of the `block.timestamp()`. 
image

mds1: Perfect, thank you both. Resolving this thread!

mds1 (approved):

tasks/sep/021-refresh-signers-fakeSC/SignFromJson.s.sol
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    /// @notice Checks the correctness of the deployment
+    function _postCheck(
+        Vm.AccountAccess[] memory accesses,
+        Simulation.Payload memory /* simPayload */
+    ) internal view override {
+        console.log("Running post-deploy assertions");
+
+        checkStateDiff(accesses);
```
blmalone, line 90: My guess is you'll likely want some more post checks in here like the previous prs.

tasks/sep/021-refresh-signers-fakeSC/SignFromJson.s.sol
```diff
@@ -0,0 +1,94 @@
...
+import {ModuleManager} from "safe-contracts/base/ModuleManager.sol";
+
+contract SignFromJson is OriginalSignFromJson {
+    using LibString for string;
+
+    // Safe contract for this task.
+    GnosisSafe securityCouncilSafe =
+        GnosisSafe(payable(vm.envAddress("OWNER_SAFE"))); // We take from the "OWNER_SAFE" as this is the "TARGET_SAFE".
+
+    // TODO: Get the livenessGuard from the SC for not hardcoding the address.
```
blmalone, line 23: do you need this todo to remain?

blmalone (approved):
lgtm pending some comments.

tasks/sep/021-refresh-signers-fakeSC/SignFromJson.s.sol
Ethnical: Thanks will remove and merge it :)

Ethnical: Fixed in commit: 4d44fe3d1a892180b9ba99832eaca3efa6d0446a

tasks/sep/021-refresh-signers-fakeSC/SignFromJson.s.sol
Ethnical: Correct!  Commit 8736779b0a8da0d1357d62eae7aba0eea465f484.

blmalone: nice, resolving.

PR #384, opened 2024-12-02
Mark sep/020 task as done

geoknee:
I got a bunch os json artifacts in `records/...`: should I commit these?


Inphi (approved):
Needs security-reviewers approval for cci changes.

The broadcast directory is an artifact of the SystemConfig update tx. We dont need to commit those.

Ethnical (approved):
Perfect!

PR #407, opened 2024-12-13
set the respected game type

sbvegan:
**Description**

This task sets the respected game type to cannon, the permissionless FaultDisputeGame.


tasks/eth/ink-002-set-respected-game-type/.env
```diff
@@ -0,0 +1,4 @@
+ETH_RPC_URL="https://ethereum.publicnode.com"
+#`OWNER_SAFE` represents the address of the `deputyGuardian` gnosis safe attached to the `DeputyGuardianModule`
+OWNER_SAFE=0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A # guardian multisig
+SAFE_NONCE=
```
mds1, line 4: Should be 99 per our notion doc

sbvegan: updated to 98, as shown in the notion's pages new ordering

runbooks/enable-permissionless/bundle.sh
mds1, line 1: Is this a new script that needs to be reviewed by proofs team, or is this copied from somewhere? If the latter let's add a code comment linking to the original source

runbooks/enable-permissionless/bundle.sh
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+OPTIMISM_PORTAL_PROXY=0x5d66c1782664115999c47c9fa5cd031f495d3e4f
+GUARDIAN=0x09f7150D8c019BeF34450d6920f6B3608ceFdAf2
+CHAINID=57073
+#DGM=0xc6901F65369FC59fC1B4D6D6bE7A2318Ff38dB5B
```
mds1, line 7: This is a runbook (so it sounds like this should generalize) but we have hardcoded data in here for one chain, and the runbook README has no instructions to change these values. My suggestion would be to delete these four declarations, and in the bundle readme, including instructions for setting those env vars and having the bundle read them. It would also be ideal if we can simply only include chain ID, and fetch the portal and guardian addreses from the superchain registry from that.

We may want to split this up into two PRs (this runbook, which we can iterate on more, and the ink task) to avoid the ink task being blocked by the runbook comments

mds1: Below in this file we also have hardcoded `data` and `signatures` which seems wrong. This script feels like it's not fully templatized?e

tasks/eth/ink-002-set-respected-game-type/.env
```diff
@@ -0,0 +1,4 @@
+ETH_RPC_URL="https://ethereum.publicnode.com"
+#`OWNER_SAFE` represents the address of the `deputyGuardian` gnosis safe attached to the `DeputyGuardianModule`
+OWNER_SAFE=0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A # guardian multisig
+SAFE_NONCE=98
```
mds1, line 4: Going forward let's only use this format, since the `SAFE_NONCE=` format is not always supported (it is off by default for nested sigs). Please double check the address though!

```suggestion
SAFE_NONCE_0X9BA6E03D8B90DE867373DB8CF1A58D2F7F006B3A=98
```

tasks/eth/ink-002-set-respected-game-type/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+**To:** `0x09f7150D8c019BeF34450d6920f6B3608ceFdAf2`
+
+**Value:** `0 WEI`
+
+**Raw Input Data**
+
+|Function Selector Signature| OptimismPortalProxyAddress (left-padded to 32 bytes) | `CANNON` game type, hardcoded to 0 and left-padded |
+|--------------------------|--------------------------------------------|--------------------------------------------|
+|0xa1155ed9|0000000000000000000000005d66c1782664115999c47c9fa5cd031f495d3e4f|0000000000000000000000000000000000000000000000000000000000000000|
```
mds1, line 29: let's delete this, it's redundant with other data and just another thing that we have to review that could be incorrect
```suggestion
```

tasks/eth/ink-002-set-respected-game-type/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+### State Validations
```
mds1, line 31: This section is redundant with the `VALIDATION.md` file. Let's take any info from here and move it into the `VALIDATION.md` file

tasks/eth/ink-002-set-respected-game-type/SignFromJson.s.sol
```diff
@@ -0,0 +1,49 @@
...
+import {LibString} from "solady/utils/LibString.sol";
+import {GnosisSafe} from "safe-contracts/GnosisSafe.sol";
+import "@eth-optimism-bedrock/src/dispute/lib/Types.sol";
+import {Simulation} from "@base-contracts/script/universal/Simulation.sol";
+
+contract SignFromJson is OriginalSignFromJson {
+    using LibString for string;
+
+    // Chains for this task.
+    address InkMainnetOptimismPortalProxy = 0x5d66C1782664115999C47c9fA5cd031f495D3e4F;
```
mds1, line 20: We should be reading this from the registry like we do in other tasks, instead of hardcoding addresses

tasks/eth/ink-002-set-respected-game-type/.env
sbvegan: it matches the foundation operations safe and this is the respected game type so it looks good

tasks/eth/ink-002-set-respected-game-type/SignFromJson.s.sol
sbvegan: got it, I'll update, I think Ink wasn't in the registry yet

runbooks/enable-permissionless/bundle.sh
sbvegan: @Inphi gave me this script and you're right its not fully templatized so I'm going to remove this

runbooks/enable-permissionless/bundle.sh
sbvegan: this was provided by mofi, I'm removing it from this task for simplicity

tasks/eth/ink-002-set-respected-game-type/VALIDATION.md
```diff
@@ -0,0 +1,38 @@
...
+  **Meaning:** Enables the simulation by setting the threshold to 1. The key can be validated by the location of the `threshold` variable in the [Safe's Storage Layout](https://github.com/safe-global/safe-smart-account/blob/v1.3.0/contracts/examples/libraries/GnosisSafeStorage.sol#L14).
+
+## State Changes
+
+Note: The changes listed below do not include safe nonce updates.
+
+###  `0x5d66c1782664115999c47c9fa5cd031f495d3e4f` (`OptimismPortalProxy`)
+
+- **Key**: `0x000000000000000000000000000000000000000000000000000000000000003b`
+  **Value**: `0x00000000000000000000000000000000000000000000000TIMESTAMP00000000`
```
mds1, line 36:
```suggestion
  **Before**: `0x00000000000000000000000000000000000000000000000TIMESTAMP00000001`
  **After**: `0x00000000000000000000000000000000000000000000000TIMESTAMP00000000`
```

tasks/eth/ink-002-set-respected-game-type/VALIDATION.md
```diff
@@ -0,0 +1,38 @@
...
+
+Overrides:
+
+- **Key:** `0x0000000000000000000000000000000000000000000000000000000000000004` <br/>
+  **Value:** `0x0000000000000000000000000000000000000000000000000000000000000001` <br/>
+  **Meaning:** Enables the simulation by setting the threshold to 1. The key can be validated by the location of the `threshold` variable in the [Safe's Storage Layout](https://github.com/safe-global/safe-smart-account/blob/v1.3.0/contracts/examples/libraries/GnosisSafeStorage.sol#L14).
+
+## State Changes
+
+Note: The changes listed below do not include safe nonce updates.
```
mds1, line 31: We should include the nonces, especially given the recent nonce mistake
```suggestion
```

tasks/eth/ink-002-set-respected-game-type/VALIDATION.md
```diff
@@ -0,0 +1,38 @@
...
+
+###  `0x5d66c1782664115999c47c9fa5cd031f495d3e4f` (`OptimismPortalProxy`)
+
+- **Key**: `0x000000000000000000000000000000000000000000000000000000000000003b`
+  **Value**: `0x00000000000000000000000000000000000000000000000TIMESTAMP00000000`
+  **Description**: Sets the `respectedGameType` to `0` (permissionless cannon game) and sets the `respectedGameTypeUpdatedAt` timestamp to the time when the upgrade transaction was executed (this will be a dynamic value).
+
```
mds1, line 38:
```suggestion

Additionally, the following nonce changes are present:
- The Foundation safe, `0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A`, has a nonce increment from 98 to 99
- The address of your account should show a nonce increment
```

tasks/eth/ink-002-set-respected-game-type/VALIDATION.md
```diff
@@ -0,0 +1,41 @@
...
+  **Meaning:** Enables the simulation by setting the threshold to 1. The key can be validated by the location of the `threshold` variable in the [Safe's Storage Layout](https://github.com/safe-global/safe-smart-account/blob/v1.3.0/contracts/examples/libraries/GnosisSafeStorage.sol#L14).
+
+## State Changes
+
+
+###  `0x5d66c1782664115999c47c9fa5cd031f495d3e4f` (`OptimismPortalProxy`)
+
+- **Key**: `0x000000000000000000000000000000000000000000000000000000000000003b` <br/>
+  **Before**: `0x0000000000000000000000000000000000000000000000006753162b00000001` <br/>
+  **After**: `0x00000000000000000000000000000000000000000000000TIMESTAMP00000000`
```
mds1, line 36:
```suggestion
  **After**: `0x00000000000000000000000000000000000000000000000TIMESTAMP00000000` <br/>
```

mds1 (approved):

PR #415, opened 2024-12-16
FPS Init - Resubmission

blmalone:
Please see https://github.com/solidity-labs-io/superchain-ops/pull/2. I'm resubmitting this PR so that CI will work.


test/mock/chainList1.toml
mds1, line 1: Which superchain registry files are these a mock of?

test/mock/data1/chainList.toml
mds1, line 1: Similarly what is this supposed to mock compared to `test/mock/chainList*.toml`?

Issue #394, opened 2024-12-10
Bug in SystemConfig Holocene Upgrade


tynes:
The [Holocene SystemConfig upgrade](https://github.com/ethereum-optimism/superchain-ops/blob/main/tasks/sep/020-holocene-system-config-upgrade-multi-chain/input.json) simply modified the implementations of the `SystemConfig` contracts rather than calling `initialize` on them. This means that there are two functions that currently do not work as expected. 

- [basefeeScalar()](https://github.com/ethereum-optimism/optimism/blob/34d5f66ade24bd1f3ce4ce7c0a6cfc1a6540eca1/packages/contracts-bedrock/src/L1/SystemConfig.sol#L117)
- [blobBasefeeScalar()](https://github.com/ethereum-optimism/optimism/blob/34d5f66ade24bd1f3ce4ce7c0a6cfc1a6540eca1/packages/contracts-bedrock/src/L1/SystemConfig.sol#L120)

This is because these values are their own storage slots rather than functions that read from a single source of truth (the `scalar`). Without calling `initialize` or calling `setGasConfigEcotone`, they will return `0`.

We have a few options:
- redo the upgrade and call `initialize` as part of the upgrade
- do a new task that calls `setGasConfigEcotone` to set the values equal to what they are currently
- leave them as 0 and update the implementation of the `SystemConfig` to parse from the `scalar` value when calling these functions

These functions would implement the inverse of this:

```solidity
scalar = (uint256(0x01) << 248) | (uint256(_blobbasefeeScalar) << 32) | _basefeeScalar;
```

So the `basefeeScalar` function would sload the `scalar`, and mask out the least significant 32 bits. The `blobBasefeeScalar` function would mask out the next least most significant 32 bits.

geoknee:
@tynes what should we do for chains that have a `scalar` set to something we can't decode? See here https://sepolia.etherscan.io/address/0x5D63A8Dc2737cE771aa4a6510D063b6Ba2c4f6F2#readProxyContract#F33

tynes:
Can you give more details? I don't understand why you cant decode `684000`. Encoded as `bytes32`, it is version 0 so it should be fine

sebastianst:
[...]

PR #418, opened 2024-12-17
sep/027: upgrade and re-initialize `SystemConfig`  across sepolia superchain

geoknee:
Closes https://github.com/ethereum-optimism/superchain-ops/issues/394

See also #454 

---
TODO 
- [x] upgrade to a multi chain task when things are working for OP Sepolia
- [x] add some breadcrumbs to the previous task `020` to signify that it was incorrect and that this task supercedes it
- [x] Build on top of @sebastianst 's new validation framework [3591eec](https://github.com/ethereum-optimism/superchain-ops/pull/418/commits/3591eecb26448799ae758ec6414f66230813c339)


tasks/sep/026-holocene-system-config-upgrade-and-init-multi-chain/VALIDATION.md
```diff
@@ -0,0 +1,19 @@
...
+This document can be used to validate the state diff resulting from the execution of the upgrade
+transaction.
+
+For each contract listed in the state diff, please verify that no contracts or state changes shown in the Tenderly diff are missing from this document. Additionally, please verify that for each contract:
+
+- The following state changes (and none others) are made to that contract. This validates that no unexpected state changes occur.
+- All addresses (in section headers and storage values) match the provided name, using the Etherscan and Superchain Registry links provided. This validates the bytecode deployed at the addresses contains the correct logic.
+- All key values match the semantic meaning provided, which can be validated using the storage layout links provided.
+
+## State Changes
```
geoknee, line 12: This should also ideally show the `baseFeeScalar` and `blobbaseFeeScalar` changes.

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/NestedSignFromJson.s.sol
```diff
@@ -0,0 +1,49 @@
...
+import {VerificationBase, SuperchainRegistry} from "script/verification/Verification.s.sol";
+import {SystemConfigUpgradeEcotoneScalars as SystemConfigUpgrade} from
+    "script/verification/SystemConfigUpgradeEcotoneScalars.s.sol";
+
+contract NestedSignFromJson is OriginalNestedSignFromJson, CouncilFoundationNestedSign {
+    string constant l1ChainName = "sepolia";
+    string constant release = "v1.8.0-rc.4";
+    string[4] l2ChainNames = ["op", "metal", "mode", "zora"];
+
+    SystemConfigUpgrade[1] sysCfgUpgrades;
```
sebastianst, line 19:
```suggestion
    SystemConfigUpgrade[] sysCfgUpgrades;
```

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/NestedSignFromJson.s.sol
```diff
@@ -0,0 +1,49 @@
...
+
+    constructor() {
+        for (uint256 i = 0; i < l2ChainNames.length; i++) {
+            sysCfgUpgrades[i] = new SystemConfigUpgrade(l1ChainName, l2ChainNames[i], release);
```
sebastianst, line 23:
```suggestion
            sysCfgUpgrades.push(new SystemConfigUpgrade(l1ChainName, l2ChainNames[i], release));
```

script/verification/Verification.s.sol
```diff
@@ -100,12 +100,14 @@ contract SuperchainRegistry is CommonBase {
         proxies.L1CrossDomainMessenger = stdToml.readAddress(toml, "$.addresses.L1CrossDomainMessengerProxy");
         proxies.L1StandardBridge = stdToml.readAddress(toml, "$.addresses.L1StandardBridgeProxy");
         proxies.SystemConfig = stdToml.readAddress(toml, "$.addresses.SystemConfigProxy");
-        proxies.AnchorStateRegistry = stdToml.readAddress(toml, "$.addresses.AnchorStateRegistryProxy");
-        proxies.DisputeGameFactory = stdToml.readAddress(toml, "$.addresses.DisputeGameFactoryProxy");
+
+        // TODO handle chains which are missing these values:
```
geoknee, line 104: Flagging this TODO

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/OVERVIEW.md
```diff
@@ -0,0 +1,57 @@
+# Holocene Hardfork - SystemConfig Upgrade
+Upgrades the `SystemConfig.sol` contract for Holocene.
+
+The batch will be executed on chain ID `11155111`, and contains `3` transactions.
```
geoknee, line 4: TODO: There are more transactions now since we made it a multi chain task.

script/verification/Verification.s.sol
geoknee: [3741e88](https://github.com/ethereum-optimism/superchain-ops/pull/418/commits/3741e88f01384df8b2ea69d3533cf25c0abffd2f)

tasks/sep/026-holocene-system-config-upgrade-and-init-multi-chain/VALIDATION.md
geoknee: [c93994a](https://github.com/ethereum-optimism/superchain-ops/pull/418/commits/c93994a645d12455919360f83ded1b9b1fdf3910)

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/OVERVIEW.md
geoknee: [0916f19](https://github.com/ethereum-optimism/superchain-ops/pull/418/commits/0916f198c3cfab43c3ba83d1f73b528fb7895e77)

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/NestedSignFromJson.s.sol
```diff
@@ -0,0 +1,52 @@
...
+contract NestedSignFromJson is OriginalNestedSignFromJson, CouncilFoundationNestedSign {
+    string constant l1ChainName = "sepolia";
+    string constant release = "v1.8.0-rc.4";
+    string[4] l2ChainNames = ["op", "metal", "mode", "zora"];
+
+    SystemConfigUpgradeEcotoneScalars[] sysCfgUpgrades;
+
+    constructor() {
+        for (uint256 i = 0; i < l2ChainNames.length; i++) {
+            sysCfgUpgrades.push(new SystemConfigUpgradeEcotoneScalars(l1ChainName, l2ChainNames[i], release));
```
geoknee, line 22: Here we deploy a new `SystemConfigUpgradeEcotoneScalars` per chain, which each contains its own bindings to the appropriate SuperchainRegistry data.

geoknee: Each instance caches some information when it is created, and this cached data is used to verify that (aside from the Ecotone scalars) no data changed in the `SystemConfig` after the task is executed.

script/verification/SystemConfigUpgrade.s.sol
```diff
@@ -0,0 +1,82 @@
...
+        address disputeGameFactory;
+        address optimismPortal;
+        address optimismMintableERC20Factory;
+    }
+
+    address public systemConfigAddress;
+    SysCfgVars previous;
+
+    constructor(string memory l1ChainName, string memory l2ChainName, string memory release)
+        SuperchainRegistry(l1ChainName, l2ChainName, release)
```
mds1, line 39: Since `l1ChainName` and `l2ChainName` are storage variable names in the `SuperchainRegistry` contract, using them as argument names in the constructors shadows those variable names
```suggestion
    constructor(string memory _l1ChainName, string memory _l2ChainName, string memory _release)
        SuperchainRegistry(_l1ChainName, _l1ChainName, _release)
```

script/verification/SystemConfigUpgrade.s.sol
```diff
@@ -0,0 +1,82 @@
...
+import {LibString} from "solady/utils/LibString.sol";
+import {SuperchainRegistry} from "script/verification/Verification.s.sol";
+import "@eth-optimism-bedrock/src/dispute/lib/Types.sol";
+import {ISystemConfig} from "./ISystemConfig.sol";
+import {IResourceMetering} from "./IResourceMetering.sol";
+import {MIPS} from "@eth-optimism-bedrock/src/cannon/MIPS.sol";
+import {Simulation} from "@base-contracts/script/universal/Simulation.sol";
+import {NestedMultisigBuilder} from "@base-contracts/script/universal/NestedMultisigBuilder.sol";
+
+contract SystemConfigUpgrade is SuperchainRegistry {
```
mds1, line 15: Can you add comments explaining what this contract is supposed to do and when it should be used? Would be good to have explanatory comments in `Verification.s.sol` as well

script/verification/SystemConfigUpgrade.s.sol
```diff
@@ -0,0 +1,82 @@
...
+import {console2 as console} from "forge-std/console2.sol";
+import {Vm} from "forge-std/Vm.sol";
+import {LibString} from "solady/utils/LibString.sol";
+import {SuperchainRegistry} from "script/verification/Verification.s.sol";
+import "@eth-optimism-bedrock/src/dispute/lib/Types.sol";
+import {ISystemConfig} from "./ISystemConfig.sol";
+import {IResourceMetering} from "./IResourceMetering.sol";
+import {MIPS} from "@eth-optimism-bedrock/src/cannon/MIPS.sol";
+import {Simulation} from "@base-contracts/script/universal/Simulation.sol";
+import {NestedMultisigBuilder} from "@base-contracts/script/universal/NestedMultisigBuilder.sol";
```
mds1, line 13: removing unused imports
```suggestion
import {LibString} from "solady/utils/LibString.sol";
import {SuperchainRegistry} from "script/verification/Verification.s.sol";
import {ISystemConfig} from "./ISystemConfig.sol";
import {IResourceMetering} from "./IResourceMetering.sol";
```

script/verification/SystemConfigUpgradeEcotoneScalars.s.sol
```diff
@@ -0,0 +1,48 @@
...
+import {MIPS} from "@eth-optimism-bedrock/src/cannon/MIPS.sol";
+import {Simulation} from "@base-contracts/script/universal/Simulation.sol";
+import {NestedMultisigBuilder} from "@base-contracts/script/universal/NestedMultisigBuilder.sol";
+import {SystemConfigUpgrade} from "script/verification/SystemConfigUpgrade.s.sol";
+
+contract SystemConfigUpgradeEcotoneScalars is SystemConfigUpgrade {
+    using LibString for string;
+
+    constructor(string memory l1ChainName, string memory l2ChainName, string memory release)
+        SystemConfigUpgrade(l1ChainName, l2ChainName, release)
```
mds1, line 20: same comment on variable shadowing
```suggestion
    constructor(string memory _l1ChainName, string memory _l2ChainName, string memory _release)
        SystemConfigUpgrade(_l1ChainName, _l2ChainName, _release)
```

script/verification/SystemConfigUpgradeEcotoneScalars.s.sol
```diff
@@ -0,0 +1,48 @@
...
+import {console2 as console} from "forge-std/console2.sol";
+import {Vm} from "forge-std/Vm.sol";
+import {LibString} from "solady/utils/LibString.sol";
+import {SuperchainRegistry} from "script/verification/Verification.s.sol";
+import "@eth-optimism-bedrock/src/dispute/lib/Types.sol";
+import {ISystemConfig} from "./ISystemConfig.sol";
+import {IResourceMetering} from "./IResourceMetering.sol";
+import {MIPS} from "@eth-optimism-bedrock/src/cannon/MIPS.sol";
+import {Simulation} from "@base-contracts/script/universal/Simulation.sol";
+import {NestedMultisigBuilder} from "@base-contracts/script/universal/NestedMultisigBuilder.sol";
```
mds1, line 13: removing unused imports

script/verification/SystemConfigUpgradeEcotoneScalars.s.sol
```diff
@@ -0,0 +1,48 @@
...
+import {SystemConfigUpgrade} from "script/verification/SystemConfigUpgrade.s.sol";
+
+contract SystemConfigUpgradeEcotoneScalars is SystemConfigUpgrade {
```
mds1, line 16: Same comment about adding explanatory code comments

script/verification/SystemConfigUpgradeEcotoneScalars.s.sol
```diff
@@ -0,0 +1,48 @@
...
+
+    constructor(string memory l1ChainName, string memory l2ChainName, string memory release)
+        SystemConfigUpgrade(l1ChainName, l2ChainName, release)
+    {}
+
+    /// @notice Public function that must be called by the verification script.
+    function checkSystemConfigUpgrade() public view override {
+        ISystemConfig sysCfg = ISystemConfig(proxies.SystemConfig);
+        uint256 reencodedScalar =
+            (uint256(0x01) << 248) | (uint256(sysCfg.blobbasefeeScalar()) << 32) | sysCfg.basefeeScalar();
```
mds1, line 27: Is this only needed temporarily since we are upgrading the SystemConfig to a version that no longer needs encoding?

script/verification/SystemConfigUpgradeEcotoneScalars.s.sol
```diff
@@ -0,0 +1,48 @@
...
+        console.log(
+            "checking baseFeeScalar and blobbaseFeeScalar ",
+            LibString.toString(sysCfg.basefeeScalar()),
+            LibString.toString(sysCfg.blobbasefeeScalar())
+        );
+        if (
+            uint8(previous.scalar >> 248) == 1 // most significant bit
+        ) {
```
mds1, line 35: Why is this behind an `if` block? A code comment for what this method is testing would be helpful for people not familiar with this encoding

script/verification/Verification.s.sol
```diff
@@ -100,16 +101,28 @@ contract SuperchainRegistry is CommonBase {
...
     }
 
+    function tryReadAddress(string memory toml, string memory key) internal pure returns (address) {
+        try vmSafe.parseTomlAddress(toml, key) returns (address a) {
+            return a;
+        } catch {
+            console.log("failed to read address for ", key);
+            return address(0);
+        }
+    }
```
mds1, line 124: We can delete this, update our forge-std version to the latest release, and instead use: https://github.com/foundry-rs/forge-std/blob/b5a86914561f38735ef1fc357685de3e7c92dc48/src/StdToml.sol#L148

You can either use it like this:
```solidity
stdToml.readAddressOr(toml, "$.addresses.AnchorStateRegistryProxy", address(0));
```
or like this, similar to LibString
```solidity
using StdToml for string;

// --- snip ---

toml.readAddressOr("$.addresses.AnchorStateRegistryProxy", address(0));
```

tasks/sep/020-holocene-system-config-upgrade-multi-chain/README.md
```diff
@@ -2,6 +2,8 @@
 
 Status: [EXECUTED](https://sepolia.etherscan.io/tx/0x88bd1d85740af3741e2ed96d6fd07f2abb4541afc667625480bf6a28451c4d6d)
 
+> ⚠️ **Warning:** This task is incorrect and is superceded by [Task 027](../027-holocene-system-config-upgrade-and-init-multi-chain/)
```
mds1, line 5: nit (according to my spellchecker extension, I also thought your spelling was correct)
```suggestion
> ⚠️ **Warning:** This task is incorrect and is superseded by [Task 027](../027-holocene-system-config-upgrade-and-init-multi-chain/)
```

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/NestedSignFromJson.s.sol
```diff
@@ -0,0 +1,52 @@
...
+import {Simulation} from "@base-contracts/script/universal/Simulation.sol";
+import {NestedSignFromJson as OriginalNestedSignFromJson} from "script/NestedSignFromJson.s.sol";
+import {DisputeGameUpgrade} from "script/verification/DisputeGameUpgrade.s.sol";
+import {CouncilFoundationNestedSign} from "script/verification/CouncilFoundationNestedSign.s.sol";
+import {VerificationBase, SuperchainRegistry} from "script/verification/Verification.s.sol";
+import {SystemConfigUpgradeEcotoneScalars} from "script/verification/SystemConfigUpgradeEcotoneScalars.s.sol";
+
+contract NestedSignFromJson is OriginalNestedSignFromJson, CouncilFoundationNestedSign {
+    string constant l1ChainName = "sepolia";
+    string constant release = "v1.8.0-rc.4";
```
mds1, line 15: The gov proposal has passed so we don't need to use the RC anymore and should be using the finalized release, which is what we want for mainnet also
```suggestion
    string constant release = "v1.8.0";
```

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/OVERVIEW.md
mds1, line 1: I'd suggest deleting this file since we don't need it, just one less thing to review

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/README.md
```diff
@@ -0,0 +1,30 @@
+# Holocene Hardfork Upgrade - `SystemConfig`
+
+Status: DRAFT, NOT READY TO SIGN
```
mds1, line 3: what do we need for this to become ready to sign?
```suggestion
Status: DRAFT, NOT READY TO SIGN
```

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Objective
+
+Upgrades the `SystemConfig` contracts for the Holocene hardfork across multiple chains in the sepolia superchain.
+
+This upgrades the `SystemConfig` in the
+[v1.8.0-rc.4](https://github.com/ethereum-optimism/optimism/tree/v1.8.0-rc.4) release.
```
mds1, line 10: we should just use 1.8.0 now that it's passed
```suggestion
This upgrades the `SystemConfig` in the
[v1.8.0](https://github.com/ethereum-optimism/optimism/tree/v1.8.0) release.
```

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Pre-deployments
+
+- `SystemConfig` - `0x33b83E4C305c908B2Fc181dDa36e230213058d7d`
```
mds1, line 14: Any messages you can link to so I can verify this address?

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Simulation
+
+Please see the "Simulating and Verifying the Transaction" instructions in [NESTED.md](../../../NESTED.md).
+When simulating, ensure the logs say `Using script /your/path/to/superchain-ops/tasks/<path>/NestedSignFromJson.s.sol`.
```
mds1, line 19:
```suggestion
When simulating, ensure the logs say `Using script /your/path/to/superchain-ops/tasks/027-holocene-system-config-upgrade-and-init-multi-chain/NestedSignFromJson.s.sol`.
```

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/VALIDATION.md
```diff
@@ -0,0 +1,85 @@
+# Validation
```
mds1, line 1: I have not yet validated this file or run the simulation

script/verification/SystemConfigUpgrade.s.sol
geoknee: [5098753](https://github.com/ethereum-optimism/superchain-ops/pull/418/commits/5098753ad37ac09680f6978ec39371d9a961e403)

script/verification/SystemConfigUpgrade.s.sol
geoknee: [ab72af4](https://github.com/ethereum-optimism/superchain-ops/pull/418/commits/ab72af4dca4dbdbcdc782b32a41a21bbd7275eb8)

script/verification/SystemConfigUpgrade.s.sol
geoknee: [ab72af4](https://github.com/ethereum-optimism/superchain-ops/pull/418/commits/ab72af4dca4dbdbcdc782b32a41a21bbd7275eb8)

script/verification/SystemConfigUpgradeEcotoneScalars.s.sol
geoknee: [ddad3df](https://github.com/ethereum-optimism/superchain-ops/pull/418/commits/ddad3df754d97266685eda3b6f0eac63fcc37ee1)

script/verification/SystemConfigUpgradeEcotoneScalars.s.sol
geoknee: [ab72af4](https://github.com/ethereum-optimism/superchain-ops/pull/418/commits/ab72af4dca4dbdbcdc782b32a41a21bbd7275eb8)

script/verification/SystemConfigUpgradeEcotoneScalars.s.sol
geoknee: [ab72af4](https://github.com/ethereum-optimism/superchain-ops/pull/418/commits/ab72af4dca4dbdbcdc782b32a41a21bbd7275eb8)

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/NestedSignFromJson.s.sol
geoknee: Added these explanations into the source code itself, as comments.

script/verification/Verification.s.sol
geoknee: Nice. [8a770be](https://github.com/ethereum-optimism/superchain-ops/pull/418/commits/8a770be124b6d7de8ee085e31dc7eaac0fa43ba3)

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/README.md
geoknee: I updated it now, there are no blockers other than review of this PR.

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/README.md
geoknee: I agree that passing governance process is a prompt for us finalizing the release. We have some options on the ordering of next steps however: 
1. We can (should) make the final tag now in any case.
2. We can update the superchain registry TOML files with the finalized tag, either 
    a. by adding a new release to `standard-versions-mainnet.toml` and keeping the existing `v1.8.0-rc.4` entry. We would then want to go back and remove the latter when all tasks referencing it have been executed.
    b. by modifying the existing release but waiting to do that until all tasks including this one are executed
    c. by modifying the existing release and modifying any existing tasks that reference it.
    
Curious what you think makes the most sense? I'd prefer if we can settle on a flow which is as clear and consistent as possible given we will go through this on each and every hard fork / upgrade. I'm inclined towards option b.

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/OVERVIEW.md
geoknee: [8a770be](https://github.com/ethereum-optimism/superchain-ops/pull/418/commits/8a770be124b6d7de8ee085e31dc7eaac0fa43ba3)

script/verification/SystemConfigUpgradeEcotoneScalars.s.sol
```diff
@@ -0,0 +1,44 @@
...
+            console.log(
+                "reencode to previous scalar: ",
+                LibString.toString(reencodedScalar),
+                LibString.toString(previous.scalar)
+            );
+            require(reencodedScalar == previous.scalar, "scalar-100 (reencoding produced incorrect result)");
+            // We do this check last because it would fail if the scalar is wrong, and we get less debug info from it.
+            // It checks all of the other fields which should not have changed (via a hash).
+            super.checkSystemConfigUpgrade();
```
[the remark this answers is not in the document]
geoknee: [58bc135](https://github.com/ethereum-optimism/superchain-ops/pull/418/commits/58bc135aea37965cb7c3a61baafe836ae4a3b86b) and [5253651](https://github.com/ethereum-optimism/superchain-ops/pull/418/commits/5253651eb9ce4c764cbc22e6cd32c0c3a26d01eb)

script/verification/SystemConfigUpgradeEcotoneScalars.s.sol
geoknee: Tidied that up here [5253651](https://github.com/ethereum-optimism/superchain-ops/pull/418/commits/5253651eb9ce4c764cbc22e6cd32c0c3a26d01eb) PTAL hopefully it is clear now.

script/verification/SystemConfigUpgrade.s.sol
```diff
@@ -0,0 +1,84 @@
...
+import {LibString} from "solady/utils/LibString.sol";
+import {SuperchainRegistry} from "script/verification/Verification.s.sol";
+import {ISystemConfig} from "./ISystemConfig.sol";
+import {IResourceMetering} from "./IResourceMetering.sol";
+
+// SystemConfigUpgrade is a contract that can be used to verify that an upgrade of the SystemConfig contract
+// has not changed any of the storage variables that are not expected to change.
+// It stores the prior storage variables EXCLUDING:
+// - scalar
+// - basedeeScalar
```
geoknee, line 14:
```suggestion
// - basefeeScalar
```

script/verification/SystemConfigUpgradeEcotoneScalars.s.sol
geoknee: The `sysCfg.blobbasefeeScalar()` and `sysCfg.basefeeScalar()` getters read from dedicated storage slots. When we did the upgrade incorrectly previously, these storage slots were not updated causing the getters to return `0`. Now, we are checking they return values consistent with what you get if you call `sysCfg.scalar()` and decode them by hand. 

Essentially we implicitly construct the expected values for the two new getters by reencoding what we get after the upgrade and checking it matches the previous `scalar` value.

I could try redesigning this so it is more explicit, and decode the expected values from the previous.scalar and compare the new getter's results with those?

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/README.md
geoknee: See also https://github.com/ethereum-optimism/optimism/issues/13009

script/verification/SystemConfigUpgrade.s.sol
```diff
@@ -0,0 +1,97 @@
...
+import {ISystemConfig} from "./ISystemConfig.sol";
+import {IResourceMetering} from "./IResourceMetering.sol";
+import {console2 as console} from "forge-std/console2.sol";
+
+// SystemConfigUpgrade is a contract that can be used to verify that an upgrade of the SystemConfig contract
+// has not changed any of the storage variables that are not expected to change.
+// It stores the prior storage variables EXCLUDING:
+// - scalar
+// - basefeeScalar
+// - blobbasefeeScalar
```
geoknee, line 16: On reflection, this base contract seems less and less useful as a reusable component. So I might look to combine it with `SystemConfigUpgradeEcotoneScalars` and rename that to `SystemConfigUpgradeHolocene`.

geoknee: (It will still be reusable across L1 and L2 chain contexts, just not across upgrades to future versions).

geoknee: Done

script/verification/HoloceneSystemConfigUpgrade.s.sol
```diff
@@ -0,0 +1,187 @@
...
+
+    // Target values for gasPayingToken and its decimals. This should represent ETH
+    // on all upgrade paths:
+    address constant targetGasPayingToken = 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE;
+    uint8 constant targetDecimals = 18;
+
+    // The target version of the SystemConfig contract, read from the SCR @ specified release:
+    string targetVersion;
+
+    // The constructor caches some information from the SCR and from the existing SystemConfig contract
```
geoknee, line 62:
```suggestion
    // The constructor caches some information from the Superchain Registry and from the existing SystemConfig contract
```

script/verification/HoloceneSystemConfigUpgrade.s.sol
```diff
@@ -0,0 +1,187 @@
...
+    function checkTargetVersion() internal view {
+        require(
+            keccak256(abi.encode(getSysCfgVersion())) == keccak256(abi.encode(targetVersion)),
+            "system-config-050: targetVersion"
+        );
+        console.log("confirmed SystemConfig upgraded to version", targetVersion);
+    }
+
+    // Checks scalar, basefeeScalar, blobbasefeeScalar are set consistently:
+    function checkScalar() internal view {
```
geoknee, line 120:
```suggestion
    // Checks scalar, basefeeScalar, blobbasefeeScalar are set consistently.
    // The upgrade will modify (via the initialize() method) the storage slots for scalar (pre-existing), basefeeScalar (new) and blobbasefeeScalar (new). 
    // checkScalar() reads the variables from the new slot and reencodes them into reecondedScalar (which is a packed version). It then checks that this is equal to the scalar value before the upgrade (previousScalar). In this way, we indirectly prove that the new slots were set correctly (in a way consistent with existing data in the SystemConfig contract). 
    function checkScalar() internal view {
```

script/verification/HoloceneSystemConfigUpgrade.s.sol
```diff
@@ -0,0 +1,187 @@
...
+            (uint256(0x01) << 248) | (uint256(sysCfg.blobbasefeeScalar()) << 32) | sysCfg.basefeeScalar();
+        console.log(
+            "checking baseFeeScalar and blobbaseFeeScalar ",
+            LibString.toString(sysCfg.basefeeScalar()),
+            LibString.toString(sysCfg.blobbasefeeScalar())
+        );
+        if (
+            // If the scalar version (i.e. the most significant bit of the scalar)
+            // is 1, we expect it to be unchanged during the upgrade.
+            uint8(previousScalar >> 248) == 1
```
geoknee, line 131: Something like this can be clearer:

```solidity
scalarVersion =uint8(previousScalar >> 248);
if  scalarVersion == 1 {
```

script/verification/HoloceneSystemConfigUpgrade.s.sol
```diff
@@ -0,0 +1,187 @@
...
+        } else {
+            // Otherwise, if the scalar version is 0,
+            // and the blobbasefeeScalar is 0,
+            // the upgrade will migrate the scalar version to 1 and preserve
+            // everything else.
+            // See https://specs.optimism.io/protocol/system-config.html?highlight=ecotone%20scalar#ecotone-scalar-overhead-uint256uint256-change
+            require(previousScalar >> 248 == 0, "scalar-101 previous scalar version != 0 or 1");
+            require(reencodedScalar >> 248 == 1, "scalar-102 reenconded scalar version != 1");
+            require(sysCfg.blobbasefeeScalar() == uint32(0), "scalar-103 blobbasefeeScalar !=0");
+            require(reencodedScalar << 8 == previousScalar << 8, "scalar-104 scalar mismatch");
```
geoknee, line 143: Two comments on this: 
1. It's not very clear, if we define a variable or at least put in the revert reason "`baseFeeScalar`" that would help a lot
2. The bit shift looks wrong, we probably want  `<< 32` instead of `<< 8`

lib/forge-std
Ethnical, line 1: - [ ] Make sure this is not causing issue during simulation + broadcast (cf: This issue we had on the previous upgrades: https://github.com/foundry-rs/foundry/issues/9648)

script/verification/HoloceneSystemConfigUpgrade.s.sol
Ethnical: Looks like this is not working indeed, i tried with the mask `0xfffff` and this seems to have a more reasonable value @geoknee: 
14aa2bca398d7e38779daed1c1b54d44fb9d852fb8378e0b4641ebcf9c47b124

script/verification/IResourceMetering.sol
```diff
@@ -0,0 +1,27 @@
...
+        uint128 maximumBaseFee;
+    }
+
+    error OutOfGas();
+
+    event Initialized(uint8 version);
+
+    function params() external view returns (uint128 prevBaseFee, uint64 prevBoughtGas, uint64 prevBlockNum); // nosemgrep
+
+    function __constructor__() external;
```
Ethnical, line 26: NIT: curious to know if this is required?

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/VALIDATION.md
```diff
@@ -0,0 +1,85 @@
...
+# 0x6baf11815ec9e3d4dc6cfa30c0e72dd31a00e45c9ca6d8a40e52f5cc78635009
+```
+
+## State Changes
+
+### `0x034edD2A225f7f429A63E0f1D2084B9E0A93b538` (`SystemConfigProxy` for op-sepolia)
+
+- **Key**: `0x0000000000000000000000000000000000000000000000000000000000000068`
+  **Before**: `0x0000000000000000000000000000000000000000000000000000000003938700`
+  **After**: `0xx00000000000000000000000000000000000d273000001db00000000003938700`
```
Ethnical, line 43: There is a typo 
```suggestion
  **After**: `0x00000000000000000000000000000000000d273000001db00000000003938700`
```

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/VALIDATION.md
```diff
@@ -0,0 +1,85 @@
...
+  **Meaning**: Updates the `basefeeScalar` and `blobbasefeeScalar` storage variables to `7600` and `86200` respectively. These share a slot with the `gasLimit` which remans at `60000000`
```
Ethnical, line 44: @geoknee there is an error here the state is updating with `862000`. 
We should make sure this is desired and this is a simple typo here: 

```
`86200` respectively.
```

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/VALIDATION.md
```diff
@@ -0,0 +1,85 @@
...
+
+
+### `0x1Eb2fFc903729a0F03966B917003800b145F56E2` (`ProxyAdminOwner` for all chains in this task)
+
+- **Key**: `0x0000000000000000000000000000000000000000000000000000000000000005`
+  **Before**: `0x000000000000000000000000000000000000000000000000000000000000000f`
+  **After**: `0x0000000000000000000000000000000000000000000000000000000000000010`
+  **Meaning**: Nonce increments by 1
+
+- **Key**: See above.
```
Ethnical, line 80: imo "see above" is not clear enough here. 
we should faciliate the access to the information to make sure signers can make sure they can highlight any difference directly. 
```suggestion
- **Key**: In the [`approvedHashes` section line 18](/blob/main/tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/VALIDATION.md#L18)
```

Ethnical (commented):
Super work with the framework for the SuperConfig upgrade! 🔥

The validation "storage" **matched** the simulation. But some value into the validation file need to be revisited.  
There is few things to complete before merge but would be fast!
 
CleanShot 2025-01-09 at 22 26 06@2x

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/input.json
```diff
@@ -0,0 +1,393 @@
+{
+  "chainId": 11155111,
+  "metadata": {
+    "name": "Holocene Hardfork - Multichain SystemConfig Upgrade",
+    "description": "Upgrades the 'SystemConfig' contract for Holocene for {op,mode,metal,zora}-sepolia"
+  },
+  "transactions": [
+    {
+      "metadata": {
+        "name": "ProxyAdmin.upgrade(SystemConfigProxy, StorageSetter)",
```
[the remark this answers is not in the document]
geoknee: It is uncertain whether this step is necessary after all.

script/verification/HoloceneSystemConfigUpgrade.s.sol
geoknee: Fixed this here. In fact what I was trying to do was compare the scalars _without_ the scalar version (first byte). [556d1d5](https://github.com/ethereum-optimism/superchain-ops/pull/418/commits/556d1d51bc7b73b13cc9aeda26cb54c042f589ad)

script/verification/IResourceMetering.sol
geoknee: [cd7ec86](https://github.com/ethereum-optimism/superchain-ops/pull/418/commits/cd7ec86dcb90083996058af5ca57e1233c4374cf)

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/VALIDATION.md
geoknee: Yes you are right this is a typo. I decoded it again by hand and have corrected the doc.

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/VALIDATION.md
```diff
@@ -0,0 +1,85 @@
...
+  **Before**: `0x0000000000000000000000000000000000000000000000000000000003938700`
+  **After**: `0xx00000000000000000000000000000000000d273000001db00000000003938700`
+  **Meaning**: Updates the `basefeeScalar` and `blobbasefeeScalar` storage variables to `7600` and `86200` respectively. These share a slot with the `gasLimit` which remans at `60000000`
+
+### `0x5D63A8Dc2737cE771aa4a6510D063b6Ba2c4f6F2` (`SystemConfigProxy` for metal-sepolia)
+
+- **Key**: `0x0000000000000000000000000000000000000000000000000000000000000068`
+  **Before**: `0x0000000000000000000000000000000000000000000000000000000001c9c380`
+  **After**: `0x0000000000000000000000000000000000000000000a6fe00000000001c9c380`
+  **Meaning**: Updates the `basefeeScalar` and `blobbasefeeScalar` storage variables to `68400` and `0` respectively. These share a slot with the `gasLimit` which remans at `30000000`
```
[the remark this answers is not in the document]
geoknee: Yes you are right this is a typo. I decoded it again by hand and have corrected the doc.

Ethnical (dismissed):
LGTM! rechecked the tasks validation + simulation make sense to me!

script/verification/HoloceneSystemConfigUpgrade.s.sol
Ethnical: Hmm, 
Just to make sure, you want to compare the if the first byte (`0x0`) are correctly incremented to (`0x1`) when the value `scalarVersion` is `0` right (if yes for me it works as expected). 
CleanShot 2025-01-10 at 13 05 23@2x

geoknee: Yes that's right.

lib/forge-std
sebastianst, line 1: is the forge-std upgrade intended?

script/verification/Verification.s.sol
```diff
@@ -4,8 +4,9 @@ pragma solidity ^0.8.15;
 import {console2 as console} from "forge-std/console2.sol";
 import {LibString} from "solady/utils/LibString.sol";
 import {Types} from "@eth-optimism-bedrock/scripts/Types.sol";
-import {CommonBase} from "forge-std/Base.sol";
+import {ScriptBase} from "forge-std/Base.sol";
 import {stdToml} from "forge-std/StdToml.sol";
+import {console2 as console} from "forge-std/console2.sol";
```
sebastianst, line 9: this is already happening on line 4?

script/verification/Verification.s.sol
```diff
@@ -21,7 +22,7 @@ contract VerificationBase {
     }
 }
 
-contract SuperchainRegistry is CommonBase {
+contract SuperchainRegistry is ScriptBase {
```
sebastianst, line 25: what's the difference? why don't we want or need this for all other script contacts?

script/verification/Verification.s.sol
```diff
@@ -102,12 +103,15 @@ contract SuperchainRegistry is CommonBase {
...
         proxies.L1StandardBridge = stdToml.readAddress(toml, "$.addresses.L1StandardBridgeProxy");
         proxies.SystemConfig = stdToml.readAddress(toml, "$.addresses.SystemConfigProxy");
-        proxies.AnchorStateRegistry = stdToml.readAddress(toml, "$.addresses.AnchorStateRegistryProxy");
-        proxies.DisputeGameFactory = stdToml.readAddress(toml, "$.addresses.DisputeGameFactoryProxy");
+
+        // Not all chains have the following values specified in the registry, so we will
+        // set them to the zero address if they are not found.
+        proxies.AnchorStateRegistry = stdToml.readAddressOr(toml, "$.addresses.AnchorStateRegistryProxy", address(0));
+        proxies.DisputeGameFactory = stdToml.readAddressOr(toml, "$.addresses.DisputeGameFactoryProxy", address(0));
+        chainConfig.unsafeBlockSigner = stdToml.readAddressOr(toml, "$.addresses.UnsafeBlockSigner", address(0));
```
sebastianst, line 111: I'd still put this line back to where it was, so it's grouped with the other chainConfig setters as before.

script/verification/HoloceneSystemConfigUpgrade.s.sol
```diff
@@ -0,0 +1,279 @@
...
+        uint32 maxResourceLimit;
+        uint8 elasticityMultiplier;
+        uint8 baseFeeMaxChangeDenominator;
+        uint32 minimumBaseFee;
+        uint32 systemTxMaxGas;
+        uint128 maximumBaseFee;
+    }
+}
+
+interface ISystemConfig {
```
sebastianst, line 213: Isn't the interface available from the imported `@eth-optimism-bedrock` package at `@eth-optimism-bedrock/src/L1/interfaces/ISystemConfig` or so? Edit: doesn't seem so. But it would be in a more recent commit of the monorepo. E.g. https://github.com/ethereum-optimism/optimism/blob/proposal/op-contracts/v1.8.0/packages/contracts-bedrock/src/L1/interfaces/ISystemConfig.sol has it. On `develop` the `interfaces` got moved into a directory tree next to the `src` tree, if we want to use a recent monorepo commit instead...

script/verification/HoloceneSystemConfigUpgrade.s.sol
```diff
@@ -0,0 +1,279 @@
...
+
+        // Read target version from SCR @ specified release
+        targetVersion = standardVersions.SystemConfig.version;
+
+        if (sysCfg.version().eq("2.3.0")) {
+            // Supported initial version
+            targetDGF = sysCfg.disputeGameFactory();
+        } else if (sysCfg.version().eq("2.2.0")) {
+            // Supported initial version
+            targetDGF = sysCfg.disputeGameFactory();
```
sebastianst, line 80: could merge them
```suggestion
        if (sysCfg.version().eq("2.3.0") || sysCfg.version().eq("2.2.0")) {
            // Supported initial version
            targetDGF = sysCfg.disputeGameFactory();
```

script/verification/HoloceneSystemConfigUpgrade.s.sol
```diff
@@ -0,0 +1,279 @@
...
+    address constant targetGasPayingToken = 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE;
+    uint8 constant targetDecimals = 18;
+
+    // The target version of the SystemConfig contract, read from the SCR @ specified release:
+    string targetVersion;
+
+    // The constructor caches some information from the Superchain Registry and from the existing SystemConfig contract
+    // before the upgrade is executed.
+    constructor(string memory _l1ChainName, string memory _l2ChainName, string memory _release)
+        SuperchainRegistry(_l1ChainName, _l2ChainName, _release)
```
sebastianst, line 63: You can make the contract `abstract` and then don't need to pass down initialization of the `SuperchainRegistry` contract. This is more modular. See my recent fp upgrade tasks, where the actual check script instantiation directly calls the `SuperchainRegistry` constructor.

```suggestion
    constructor() # and make contract abstract
```

script/verification/HoloceneSystemConfigUpgrade.s.sol
```diff
@@ -0,0 +1,279 @@
...
+        if (block.chainid != 1) exceptions[i++] = previous.owner;
+        exceptions[i++] = address(uint160(uint256((previous.batcherHash))));
+        exceptions[i++] = previous.unsafeBlockSigner;
+        exceptions[i++] = previous.batchInbox;
+    }
+
+    // Checks semver of SystemConfig is correct after the upgrade
+    function checkTargetVersion() internal view {
+        require(
+            keccak256(abi.encode(getSysCfgVersion())) == keccak256(abi.encode(targetVersion)),
```
sebastianst, line 111:
```suggestion
            getSysCfgVersion().eq(targetVersion),
```

script/verification/HoloceneSystemConfigUpgrade.s.sol
```diff
@@ -0,0 +1,279 @@
...
+
+    // Checks the remaining storage variables are unchanged after the upgrade
+    function checkBaseSysCfgVars() internal view {
+        // Check remaining storage variables didn't change
+        require(keccak256(abi.encode(getBaseSysCfgVars())) == keccak256(abi.encode(previous)), "system-config-100");
+    }
+
+    // Reads the semantic version of the SystemConfig contract
+    function getSysCfgVersion() internal view returns (string memory) {
+        return sysCfg.version();
```
sebastianst, line 180: if you change to use an imported `ISystemConfig`, you can get the `version()` function with interface `ISemver`.

script/verification/HoloceneSystemConfigUpgrade.s.sol
```diff
@@ -0,0 +1,279 @@
...
+        } else if (sysCfg.version().eq("1.12.0")) {
+            // Supported initial version
+            targetDGF = address(0);
+        } else {
+            revert("unsupported SystemConfig version");
+        }
+    }
+
+    /// @notice Public function that must be called by the verification script.
+    function checkSystemConfigUpgrade() public view {
```
sebastianst, line 90: Do you check for parity of the `resourceConfig`, or is this not necessary? If not, and you can't import the `ISystemConfig` interface from the monorepo, you could trim down the interface you define below in the file to only those functions you need, and also remove the `IResourceMetering` and its contained struct.

script/verification/HoloceneSystemConfigUpgrade.s.sol
```diff
@@ -0,0 +1,279 @@
...
+interface IResourceMetering {
+    struct ResourceConfig {
+        uint32 maxResourceLimit;
+        uint8 elasticityMultiplier;
+        uint8 baseFeeMaxChangeDenominator;
+        uint32 minimumBaseFee;
+        uint32 systemTxMaxGas;
+        uint128 maximumBaseFee;
+    }
+}
```
sebastianst, line 211: I think you could also just define the struct directly
```suggestion
struct ResourceConfig {
    uint32 maxResourceLimit;
    uint8 elasticityMultiplier;
    uint8 baseFeeMaxChangeDenominator;
    uint32 minimumBaseFee;
    uint32 systemTxMaxGas;
    uint128 maximumBaseFee;
}
```

script/verification/HoloceneSystemConfigUpgrade.s.sol
```diff
@@ -0,0 +1,279 @@
...
+
+    function getCodeExceptions() public view returns (address[] memory exceptions) {
+        uint256 len = block.chainid == 1 ? 3 : 4; // Mainnet doesn't need owner exception.
+        exceptions = new address[](len);
+        uint256 i = 0;
+        if (block.chainid != 1) exceptions[i++] = previous.owner;
+        exceptions[i++] = address(uint160(uint256((previous.batcherHash))));
+        exceptions[i++] = previous.unsafeBlockSigner;
+        exceptions[i++] = previous.batchInbox;
+    }
```
sebastianst, line 106: A better pattern with the verification framework is to make the contract also a `VerificationBase` and then here do
```suggestion
    function _getCodeExceptions() internal {
        if (block.chainid != 1) addCodeException(previous.owner);
        addCodeException(address(uint160(uint256(previous.batcherHash))));
        addCodeException(previous.unsafeBlockSigner);
        addCodeException(previous.batchInbox);
    }
```
then call `_getCodeExceptions()` inside the `constructor`, and in the actual check script instance, you do
```solidity
    function getAllowedStorageAccess() internal view override returns (address[] memory) {
        return allowedStorageAccess;
    }

    function getCodeExceptions() internal view override returns (address[] memory) {
        return codeExceptions;
    }
```
see for example the check script for task eth/022.

sebastianst (commented):
Part 1

script/verification/HoloceneSystemConfigUpgrade.s.sol
sebastianst: Ahh I see that you need to make multiple instances and don't inherit this contract directly... ok then it makes sense.

script/verification/HoloceneSystemConfigUpgrade.s.sol
sebastianst: OK I realize you need to collect all those code exceptions from multiple instances. Still you could inherit the `VerificationBase` contract and then do something like
```suggestion
    // call this from main script
    function getCodeExceptions() public view returns (address[] memory exceptions) {
        return codeExceptions
    }
    
    //  call this from the constructor
    function _addCodeExceptions() internal {
        if (block.chainid != 1) addCodeException(previous.owner);
        addCodeException(address(uint160(uint256(previous.batcherHash))));
        addCodeException(previous.unsafeBlockSigner);
        addCodeException(previous.batchInbox);
    }
```

lib/forge-std
geoknee: Yes. It pulls in `stdToml.readAddressOr()` which lets us handle chains which have missing data in the superchain registry.

script/verification/Verification.s.sol
geoknee: I think I may be able to revert this. I had it previously because it exposed `external` TOML utilities which I could wrap with a try/catch.

script/verification/HoloceneSystemConfigUpgrade.s.sol
```diff
@@ -0,0 +1,279 @@
...
+    function checkGasPayingToken() internal view {
+        (address t, uint8 d) = sysCfg.gasPayingToken();
+        require(t == targetGasPayingToken, "scalar-107");
+        require(d == targetDecimals, "scalar-108");
+    }
+
+    // Checks the remaining storage variables are unchanged after the upgrade
+    function checkBaseSysCfgVars() internal view {
+        // Check remaining storage variables didn't change
+        require(keccak256(abi.encode(getBaseSysCfgVars())) == keccak256(abi.encode(previous)), "system-config-100");
```
sebastianst, line 172:
```suggestion
        require(getBaseSysCfgVars().eq(previous), "system-config-100");
```

script/verification/HoloceneSystemConfigUpgrade.s.sol
geoknee: I actually think I don't want to float this dependency. This task is for a very specific SystemConfig interface so I'm pretty happy with hardcoding the interface here.

script/verification/HoloceneSystemConfigUpgrade.s.sol
geoknee: I do check for parity for `resourceConfig`.

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/NestedSignFromJson.s.sol
```diff
@@ -0,0 +1,54 @@
...
+            }
+        }
+    }
+
+    function _postCheck(Vm.AccountAccess[] memory accesses, Simulation.Payload memory) internal view override {
+        console.log("Running post-deploy assertions");
+        checkStateDiff(accesses);
+        for (uint256 i = 0; i < l2ChainNames.length; i++) {
+            console.log("");
+            console.log("Running post-deploy assertions for chain", l2ChainNames[i], "-", l1ChainName);
```
sebastianst, line 41: nit, I think the convention is the other way round
```suggestion
            console.log("Running post-deploy assertions for chain", l1ChainName, "-", l2ChainNames[i]);
```

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/VALIDATION.md
```diff
@@ -0,0 +1,89 @@
...
+# 0xc18fefc0a6b81265cf06017c3f1f91c040dc3227321d73c608cfbcf1c5253e5c
+cast index bytes32 0x28342fccf7308fc0967d8303fd5289550a30acff2de8754cf384b524ebe9ca0a 0xc18fefc0a6b81265cf06017c3f1f91c040dc3227321d73c608cfbcf1c5253e5c
+# 0x6baf11815ec9e3d4dc6cfa30c0e72dd31a00e45c9ca6d8a40e52f5cc78635009
+```
+
+## State Changes
+
+### `0x034edD2A225f7f429A63E0f1D2084B9E0A93b538` (`SystemConfigProxy` for op-sepolia)
+
+* **Key**: `0x0000000000000000000000000000000000000000000000000000000000000068`
```
sebastianst, line 41: it may make sense to link to the storage layout so that the keys can be verified.

script/verification/HoloceneSystemConfigUpgrade.s.sol
geoknee: I'm not sure this will work for my multi chain task though?

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/VALIDATION.md
```diff
@@ -0,0 +1,89 @@
...
+* **Before**: `0x0000000000000000000000000000000000000000000000000000000003938700`
+* **After**: `0x00000000000000000000000000000000000d273000001db00000000003938700`
+* **Meaning**: Updates the `basefeeScalar` and `blobbasefeeScalar` storage variables to `7600` and `862000` respectively. These share a slot with the `gasLimit` which remains at `60000000`
```
sebastianst, line 44: it would help to say that `cast th 60000000` equals `0x3938700`. same for other int-hex comparisons

script/verification/HoloceneSystemConfigUpgrade.s.sol
geoknee: Perhaps, but then I need to change a bunch of references and this moves more towards a completely hand rolled interface for the SystemConfig. I like it better like this where we have a snapshot of how the interface actually looks.

script/verification/HoloceneSystemConfigUpgrade.s.sol
geoknee: I don't think this one will work because I'm not ` using LibString for BaseSysCfgVars` and I don't think I can either.

sebastianst: ah lol, I realize this is not a string comparison, sorry!

script/verification/HoloceneSystemConfigUpgrade.s.sol
sebastianst: OK makes sense.

script/verification/HoloceneSystemConfigUpgrade.s.sol
sebastianst: My proposal still gives you that. It just has two advantages
1. you use the `VerificationBase` to give you `addCodeException` to manage the array
2. if you use this contract in a single-verification (not multi-instance like this one), it's automatically compatible with that pattern.

Note that my proposal gives you the exact same `function getCodeExceptions() public view returns (address[] memory exceptions)` so same api.

script/verification/HoloceneSystemConfigUpgrade.s.sol
sebastianst: sorry yes, by hashing the whole thing 👑

script/verification/HoloceneSystemConfigUpgrade.s.sol
sebastianst: I see you want to retain the explicit interface.

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/NestedSignFromJson.s.sol
sebastianst: sorry, this is wrong, your way is correct 🙈

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/VALIDATION.md
geoknee: [1e1fc2b](https://github.com/ethereum-optimism/superchain-ops/pull/418/commits/1e1fc2b128be4404cd8245e3075ccaa61700cdaf)

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/VALIDATION.md
geoknee: [3eec9d8](https://github.com/ethereum-optimism/superchain-ops/pull/418/commits/3eec9d8bfbdb740f8e7f5f5c3c0fd48e477d0e60)

script/verification/HoloceneSystemConfigUpgrade.s.sol
```diff
@@ -0,0 +1,279 @@
...
+    // In this way, we indirectly prove that the new slots were set correctly (in a way consistent with existing data in the SystemConfig contract).
+    function checkScalar() internal view {
+        // Check that basefeeScalar and blobbasefeeScalar are correct by re-encoding them and comparing to the new scalar value.
+        uint256 reencodedScalar =
+            (uint256(0x01) << 248) | (uint256(sysCfg.blobbasefeeScalar()) << 32) | sysCfg.basefeeScalar();
+        console.log(
+            "checking baseFeeScalar and blobbaseFeeScalar reencoded to ",
+            LibString.toString(sysCfg.basefeeScalar()),
+            LibString.toString(sysCfg.blobbasefeeScalar()),
+            LibString.toString(reencodedScalar)
```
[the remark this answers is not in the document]
geoknee: [c904f92](https://github.com/ethereum-optimism/superchain-ops/pull/418/commits/c904f925609bcfb52c7d453b71eaf54fa9c86c54)

script/verification/HoloceneSystemConfigUpgrade.s.sol
geoknee: [31ab749](https://github.com/ethereum-optimism/superchain-ops/pull/418/commits/31ab749a1b08412a1bb4cc5b5083d54b4e30f368)

script/verification/HoloceneSystemConfigUpgrade.s.sol
```diff
@@ -0,0 +1,279 @@
...
+            // the blobbasefeeScalar is implicitly 0 and
+            // the upgrade will migrate the scalar version to 1 and preserve
+            // everything else.
+            // See https://specs.optimism.io/protocol/system-config.html?highlight=ecotone%20scalar#ecotone-scalar-overhead-uint256uint256-change
+            require(previousScalarEncodingVersion == 0, "scalar-101 previous scalar version != 0 or 1");
+            require(newScalarEncodingVersion == 1, "scalar-102 reenconded scalar version != 1");
+            require(sysCfg.blobbasefeeScalar() == uint32(0), "scalar-103 blobbasefeeScalar !=0");
+
+            // The scalars should match if we add the new scalar version byte to the previous scalar.
+            require(reencodedScalar == previousScalar + (uint256(0x01) << 248), "scalar-104 scalar mismatch");
```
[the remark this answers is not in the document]
geoknee: We do not guarantee that. This would be to cover an invalid scalar encoding, correct? I'm assuming the scalar was encoded correctly. Do you think extra checks are warranted to cover encodings which violate the spec?

script/verification/Verification.s.sol
geoknee: Good catch, removed.

script/verification/Verification.s.sol
geoknee: I reverted this change in the end.

script/verification/HoloceneSystemConfigUpgrade.s.sol
geoknee: I see, took a crack here [acd812e](https://github.com/ethereum-optimism/superchain-ops/pull/418/commits/acd812e94dcacc0638d05501af8788fa759d657d)

script/verification/Verification.s.sol
geoknee: [17bff2d](https://github.com/ethereum-optimism/superchain-ops/pull/418/commits/17bff2d0b9a2931594f6084d6284ebe88aeaa947)

script/verification/HoloceneSystemConfigUpgrade.s.sol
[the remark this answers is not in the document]
sebastianst: It would force us to take a closer look at any chain that somehow managed to set an invalid scalar. It could also be a pre-check (called from the ctor of the check contract).

script/verification/HoloceneSystemConfigUpgrade.s.sol
sebastianst: yes thanks, that what I meant :)

script/verification/HoloceneSystemConfigUpgrade.s.sol
```diff
@@ -0,0 +1,276 @@
...
+    /// @notice Public function that must be called by the verification script.
+    function checkSystemConfigUpgrade() public view {
+        checkTargetVersion();
+        checkScalar();
+        checkDGF();
+        checkGasPayingToken();
+        checkBaseSysCfgVars();
+    }
+
+    function _getCodeExceptions() internal {
```
sebastianst, line 97: nit, this is now an _adder_ not a _getter_.
```suggestion
    function _addCodeExceptions() internal {
```

lib/forge-std
sebastianst: ah nice

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/README.md
```diff
@@ -0,0 +1,47 @@
...
+[v1.8.0-rc.4](https://github.com/ethereum-optimism/optimism/tree/v1.8.0-rc.4) release.
+
+## Pre-deployments
+
+- `SystemConfig` - `0x33b83E4C305c908B2Fc181dDa36e230213058d7d`
+
+## Simulation
+
+Please see the "Simulating and Verifying the Transaction" instructions in [NESTED.md](../../../NESTED.md).
+When simulating, ensure the logs say `Using script /your/path/to/superchain-ops/tasks/027-holocene-system-config-upgrade-and-init-multi-chain/NestedSignFromJson.s.sol`.
```
sebastianst, line 19:
```suggestion
When simulating, ensure the logs say `Using script /your/path/to/superchain-ops/tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/NestedSignFromJson.s.sol`.
```
you fix this in the [mainnet pr](https://github.com/ethereum-optimism/superchain-ops/pull/454/files#diff-6e1818466a6f803978bd56a0a754e9b9de17381edc235a5f7f1a08f8f012a484)

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/NestedSignFromJson.s.sol
```diff
@@ -0,0 +1,54 @@
...
+            // which each contains its own bindings to an individual chain's SuperchainRegistry data.
+            sysCfgUpgrades.push(new HoloceneSystemConfigUpgrade(l1ChainName, l2ChainNames[i], release));
+            console.log("");
+            console.log("Set up verification data for chain", l2ChainNames[i], "-", l1ChainName);
+            console.log("with SystemConfigProxy @", sysCfgUpgrades[i].systemConfigAddress());
+            addAllowedStorageAccess(sysCfgUpgrades[i].systemConfigAddress());
+            address[] memory exceptions = sysCfgUpgrades[i].getCodeExceptions();
+            for (uint256 j = 0; j < exceptions.length; j++) {
+                addCodeException(exceptions[j]);
+            }
```
sebastianst, line 32: We could add the array adder to the `VerificationBase`, then we could just do
```suggestion
            addCodeExceptions(sysCfgUpgrades[i].getCodeExceptions());
```
here and in following tasks.

geoknee: [540baec](https://github.com/ethereum-optimism/superchain-ops/pull/418/commits/540baec4113f4bed0358a24912ba6eb18d5050df)

script/verification/HoloceneSystemConfigUpgrade.s.sol
geoknee: [a99ddbf](https://github.com/ethereum-optimism/superchain-ops/pull/418/commits/a99ddbfdc6c3b0c8a5522db44001a9d3be8473c6)

script/verification/HoloceneSystemConfigUpgrade.s.sol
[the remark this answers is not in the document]
geoknee: [7bb795b](https://github.com/ethereum-optimism/superchain-ops/pull/418/commits/7bb795b6aab31b87178d37958f86c40fc3b44cce)

sebastianst (approved):

script/verification/HoloceneSystemConfigUpgrade.s.sol
```diff
@@ -0,0 +1,280 @@
...
+            // Otherwise, if the scalar version is 0,
+            // the blobbasefeeScalar is implicitly 0 and
+            // the upgrade will migrate the scalar version to 1 and preserve
+            // everything else.
+            // See https://specs.optimism.io/protocol/system-config.html?highlight=ecotone%20scalar#ecotone-scalar-overhead-uint256uint256-change
+            require(previousScalarEncodingVersion == 0, "scalar-101 previous scalar version != 0 or 1");
+            require(newScalarEncodingVersion == 1, "scalar-102 reenconded scalar version != 1");
+            require(sysCfg.blobbasefeeScalar() == uint32(0), "scalar-103 blobbasefeeScalar !=0");
+
+            // Sanity check the scalar "padding" is zero for legacy encodings (see spec)
```
Ethnical, line 152: Maybe also add the link to the specs for the legacy encoding here?

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/VALIDATION.md
```diff
@@ -0,0 +1,89 @@
...
+* **Before**: `0x0000000000000000000000000000000000000000000000000000000001c9c380`
+* **After**: `0x00000000000000000000000000000000000941ad000003f40000000001c9c380`
+* **Meaning**: Updates the `basefeeScalar` and `blobbasefeeScalar` storage variables to `1012` (`cast td 0x000003f4`) and `606637` (cast td `0x000941ad`) respectively. These share a slot with the `gasLimit` which remains at `30000000` (`cast td 0x0000000001c9c380`).
+
+
+### `0x1Eb2fFc903729a0F03966B917003800b145F56E2` (`ProxyAdminOwner` for all chains in this task)
+
+* **Key**: `0x0000000000000000000000000000000000000000000000000000000000000005`
+* **Before**: `0x000000000000000000000000000000000000000000000000000000000000000f`
+* **After**: `0x0000000000000000000000000000000000000000000000000000000000000010`
```
Ethnical, line 79: Invalid state validation here has been incremented. 
Need to recheck why and update the correct nonce expected.

Ethnical (changes requested):
Validation state is now incorrect with the nonce.

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/VALIDATION.md
geoknee: [85bca50](https://github.com/ethereum-optimism/superchain-ops/pull/418/commits/85bca50362715920d34ac96536de114b590f5fa9)

script/verification/HoloceneSystemConfigUpgrade.s.sol
geoknee: It is already linked on L147 just above.

Ethnical (approved):

tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/input.json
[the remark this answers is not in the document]
geoknee: We won't do this step.

PR #421, opened 2024-12-17
sep/026: Add task to fix sepolia fp upgrade

sebastianst:
**Description**

Adds task to fix the Sepolia FP upgrade, since the old one (https://github.com/ethereum-optimism/superchain-ops/pull/374) referenced a wrongly newly deployed `DelayedWETH` implementation.

~Also adds a framework for creating reusable verification scripts.~ The framework was split into its own PR with #429 

**Additional context**

~This task may actually be obsolete because of #410~ update: it's necessary because #410 reused the wrong `DelayedWETH`s.

Based on https://github.com/ethereum-optimism/superchain-ops/pull/442 which adds additional checks to DG upgrades.

**Metadata**

Towards https://github.com/ethereum-optimism/optimism/issues/13033

PR #429, opened 2024-12-19
Add verification framework & FP dispute game module

sebastianst:
Developed in #421 and now split into its own PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,6 +20,28 @@ parameters:
     type: integer
     default: 5
 
+commands:
+  simulate_nested:
+    description: "Runs simulations of a nested task"
+    parameters:
+      task:
+        type: string
+    steps:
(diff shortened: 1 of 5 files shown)


lib/superchain-registry
Ethnical, line 1: I am just worried about the modification of the submodules `superchain-registry and `solady` that can potentially break stuffs but overall looks great to me!

Ethnical (approved):
I am just worried about the modification of the submodules `superchain-registry` and `solady` that can potentially lead to issue. 
 But overall looks great to me!

lib/superchain-registry
sebastianst: The SCR needs to be updated for all following Holocene-related tasks anyways. so I just did it already in this PR. In general, we should update dependencies like the SCR and the optimism monorepo more frequently.

PR #352, opened 2024-10-23
Base mainnet fp upgrade

sbvegan:
**Description**

Adding the mainnet task to upgrade Base Mainnet to using permissionless fault proofs.

PR #430, opened 2024-12-19
base-003: Base Mainnet Holocene FP upgrade

sebastianst:
**Description**

Upgrades the new PDG and FDG for Base mainnet.

Deployments:
* [FaultDisputeGame](https://etherscan.io/address/0xc5f3677c3C56DB4031ab005a3C9c98e1B79D438e#code): `0xc5f3677c3C56DB4031ab005a3C9c98e1B79D438e`
* [PermissionedDisputeGame](https://etherscan.io/address/0xF62c15e2F99d4869A925B8F57076cD85335832A2#code): `0xF62c15e2F99d4869A925B8F57076cD85335832A2`

**Tests**

Uses reusable check scripts.


tasks/eth/base-003-holocene-fp-upgrade/.env
```diff
@@ -0,0 +1,12 @@
...
+
+# Base L1 PAO
+OWNER_SAFE=0x7bB41C3008B3f03FE483B28b8DB90e19Cf07595c
+# This is Base's safe but the Nested.just file uses council as a keyword and
+# we want to minimize changes
+COUNCIL_SAFE=0x9855054731540A48b28990B63DcF4f33d8AE46A1
+# Foundation Operations Safe
+FOUNDATION_SAFE=0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A
+
+# No nonce overrides needed, this is the next task for B1PAO, BOS, FOS
```
Ethnical, line 11: We should hardcode the NONCE of the safes following our notion doc. and also update the nested validation with the storage increment. 
```
SAFE_NONCE=4
SAFE_NONCE_0X9855054731540A48B28990B63DCF4F33D8AE46A1=16
SAFE_NONCE_0X9BA6E03D8B90DE867373DB8CF1A58D2F7F006B3A=97
```

tasks/eth/base-003-holocene-fp-upgrade/input.json
```diff
@@ -0,0 +1,67 @@
+{
+  "chainId": 8453,
```
Ethnical, line 2: The chainId should be mainnet (`1`) imo. 
I don't know why the chainId is 8453, the task would be executed on mainnet. 
In the current script, I don't know where the `chainId` is used but for consistency with previous [task](https://github.com/ethereum-optimism/superchain-ops/pull/352/files#diff-9fe6757affc75ddec67b6b5965a760112ff204a06de6a16bce41252b1e87e61a) probably better to have 1.
```suggestion
  "chainId": 1,
```

tasks/eth/base-003-holocene-fp-upgrade/input.json
```diff
@@ -0,0 +1,67 @@
...
+  "metadata": {
+    "name": "Base Mainnet Holocene - Proof Contract Upgrades",
+    "description": "Upgrades the `MIPS.sol`, `FaultDisputeGame.sol`, and `PermissionedDisputeGame.sol` contracts for Holocene."
```
Ethnical, line 5: I don't see where the `MIPS.sol` is upgraded in this task.

tasks/eth/base-003-holocene-fp-upgrade/input.json
sebastianst: Ah true, will fix. This field is meaningless anyways 😂

tasks/eth/base-003-holocene-fp-upgrade/.env
sebastianst: There won't be increments/overrides in the validation, even if we set these explicitly. I'll add them anyways.

tasks/eth/base-003-holocene-fp-upgrade/input.json
sebastianst: Same response as last time, MIPS isn't upgraded directly at a proxy, but implicitly because the new games reference a new MIPS implementation.

Ethnical: Thanks! Sorry, I didn't saw the previous answer that make sense to me now :)

NESTED-VALIDATION.md
```diff
@@ -88,15 +92,15 @@ The GnosisSafe `approvedHashes` mapping is updated to indicate approval of this
...
-When the Security Council executes a transaction, the liveness timestamp are updated for each owner that signed the tasks.
-This is updating at the moment of the transaction is submitted (`block.timestamp`) into the [`lastLive`](https://github.com/ethereum-optimism/optimism/blob/e84868c27776fd04dc77e95176d55c8f6b1cc9a3/packages/contracts-bedrock/src/safe/LivenessGuard.sol#L41) mapping located at the slot `0`.
+When the Security Council executes a transaction, the liveness timestamps are updated for each owner that signed the task.
+This is updating at the moment when the transaction is submitted (`block.timestamp`) into the [`lastLive`](https://github.com/ethereum-optimism/optimism/blob/e84868c27776fd04dc77e95176d55c8f6b1cc9a3/packages/contracts-bedrock/src/safe/LivenessGuard.sol#L41) mapping located at the slot `0`.
 
 ### Nonce increments
 
 The only other state changes related to the nested execution are _three_ nonce increments:
 
 - One on the `ProxyAdminOwner` 2/2. If this is not decoded, it corresponds to key `0x05` on a `GnosisSafeProxy`.
```
Ethnical, line 104:
```suggestion
- One increment on the slot of the  `ProxyAdminOwner` 2/2 also known as (L1PAO), located at the storage key:  `0x0000000000000000000000000000000000000000000000000000000000000005` of the `Safe`.
```

NESTED-VALIDATION.md
```diff
@@ -88,15 +92,15 @@ The GnosisSafe `approvedHashes` mapping is updated to indicate approval of this
...
-- One on the Council or Foundation Safe. If this is not decoded, it corresponds to key `0x05` on a `GnosisSafeProxy`.
+- One on the Safe Signer. If this is not decoded, it corresponds to key `0x05` on a `GnosisSafeProxy`.
```
Ethnical, line 105:
```suggestion
- One increment on the slot of the  _Safe Signer_ (that is a owner of the ProxyAdminOwner above) located at the storage key:  `0x0000000000000000000000000000000000000000000000000000000000000005` of the `Safe` contract.
```

tasks/eth/base-003-holocene-fp-upgrade/.env
```diff
@@ -0,0 +1,15 @@
+ETH_RPC_URL="https://ethereum.publicnode.com"
+
+# Base L1 PAO
+OWNER_SAFE=0x7bB41C3008B3f03FE483B28b8DB90e19Cf07595c
+# This is Base's safe but the Nested.just file uses council as a keyword and
+# we want to minimize changes
```
Ethnical, line 6: Agreed here! Thanks for putting this at comment for the future readers!

tasks/eth/base-003-holocene-fp-upgrade/.env
```diff
@@ -0,0 +1,15 @@
...
+COUNCIL_SAFE=0x9855054731540A48b28990B63DcF4f33d8AE46A1
+# Foundation Operations Safe
+FOUNDATION_SAFE=0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A
+
+# No nonce overrides needed, this is the next task for B1PAO, BOS, FOS.
```
Ethnical, line 11: Thanks!! we do that in case that the `BOS` do an action to be notified that the nonce changed and could require investigation.

tasks/eth/base-003-holocene-fp-upgrade/input.json
```diff
@@ -0,0 +1,67 @@
...
+      },
+      "contractInputsValues": {
+        "_gameType": "1",
+        "_impl": "0xF62c15e2F99d4869A925B8F57076cD85335832A2"
+      }
+    },
+    {
+      "metadata": {
+        "name": "Upgrade `CANNON` game type in `DisputeGameFactory`",
+        "description": "Upgrades the `CANNON` game type to the new Holocene deployment, with an updated version of `op-program` as the absolute prestate hash."
```
Ethnical, line 40:
```suggestion
        "description": "Upgrade the `CANNON` game type to the new Holocene deployment, with an updated version of `op-program` as the absolute prestate hash."
```

Ethnical (commented):
Nice work!
The state looks correct and expected with the 2 games added.
CleanShot 2024-12-20 at 23 21 27@2x

Maybe possible adding more information about the `MIPS.sol` that is located into the PDG & FDG contracts (with the var `vm`)  for the reader in the future. 
To make sure signers can also check.

tasks/eth/base-003-holocene-fp-upgrade/input.json
```diff
@@ -0,0 +1,67 @@
...
+  "chainId": 1,
+  "metadata": {
+    "name": "Base Mainnet Holocene - Proof Contract Upgrades",
+    "description": "Upgrades the `MIPS.sol`, `FaultDisputeGame.sol`, and `PermissionedDisputeGame.sol` contracts for Holocene."
+  },
+  "transactions": [
+    {
+      "metadata": {
+        "name": "Upgrade `PERMISSIONED_CANNON` game type in `DisputeGameFactory`",
+        "description": "Upgrades the `PERMISSIONED_CANNON` game type to the new Holocene deployment, with an updated version of `op-program` as the absolute prestate hash."
```
Ethnical, line 11:
```suggestion
        "description": "Upgrade the `PERMISSIONED_CANNON` game type to the new Holocene deployment, with an updated version of `op-program` as the absolute prestate hash."
```

tasks/eth/base-003-holocene-fp-upgrade/.env
sebastianst: I copied the comment too 😂

tasks/eth/base-003-holocene-fp-upgrade/input.json
sebastianst: This just omits the `It` which is common in short descriptive sentences. Also, the `"description"` fields has no impact at all, so I'll just leave it as is, also because it's auto generated and I don't want to change the auto-generation script.

tasks/eth/base-003-holocene-fp-upgrade/input.json
sebastianst: This just omits the `It` which is common in short descriptive sentences. Also, the `"description"` fields has no impact at all, so I'll just leave it as is, also because it's auto generated and I don't want to change the auto-generation script.

NESTED-VALIDATION.md
sebastianst: I think it's a matter of taste. I like short sentences of rather trivial facts over long sentences that require more brain-parsing 😅 I also deliberately didn't mention the L1PAO or BL1PAO here again, because the `ProxyAdminOwner` is already defined at the top. This just reduces the chance of us forgetting to update this doc here when we add additional Safes in the future. I'll still change it to be slightly more descriptive.

Ethnical: Ok sounds good to me thanks for the explanation.

PR #438, opened 2025-01-02
Adding the Anvil Fork on the multiples tasks.

Ethnical:
This PR simulated multiples tasks in a row within the same anvil fork.
Really useful for ensuring multiples upgrade in the same batch and avoid error that would make all the signers resign again. 

https://github.com/user-attachments/assets/a0ba8bf4-d5cd-48fc-96ba-ffe53154343f

To execute the script just need to do: 
```
 ./script/utils/sim-sequence.sh eth "021 022 base-003 ink-001"
```

and to kill anvil if needed because stuck. 
```
ps aux | grep anvil | grep -v grep | awk '{print $2}' | xargs kill; ./script/utils/sim-sequence.sh eth "021 022 base-003 ink-001"
```
## TESTS
Tests made are covering all the necessary case we need using this:
```
#   - expected to pass: ./script/utils/sim-sequence.sh eth "021 022 base-003 ink-001" ✅
#   - expected to pass: ./script/utils/sim-sequence.sh eth "021 022 base-003" ✅
#   - expected to fail: ./script/utils/sim-sequence.sh eth "021 base-003 ink-001 022" ✅
#   - expected to fail: ./script/utils/sim-sequence.sh eth "021 base-003 ink-001" ✅
#   - expected to fail: ./script/utils/sim-sequence.sh eth "021 025" ✅
# - expected to fail: create a dir called 021-tmp and run ./script/utils/sim-sequence.sh eth "021 022 base-003 ink-001" ✅
```

Would merge this one to close the PR #437 

---
OLD content: 
_Trying to add the Anvil fork feature, however I am stuck when the simulation **broadcast** into the anvil fork. 
Somehow this is only deploying the contract and create an inconsistency between the simulation and the execution onchain. 
cf screenshot -> 
**The trace before** broadcast: 
image

**The trace during the broadcast**, only deploying the contract (Maybe I should call a function into this new contract created for now, I am stuck). 
594a0d86598527e0b6a599408b12ee3fae979aed40def75127794c908ea40c50

If anyone has an idea why?_


Ethnical:
I have fixed the previous issue, now I am stuck in the part to generate a correct signature since the threshold need to be `1` at least. 
My plan is to: 

- [ ] 1.Set the Threshold to of owner of the safe at `1`.

- [ ] 2. **Use the anvil cheatcode to set an owner address with a known private keys** (so through the sentinel_address and remove the rest of the signers). In this version we already set the storage of the threshold to `1`.
- [ ] 3.  generate a signature of the datahash previously get from the simulation with the known privatekey and send it instead of the dummy signature.

Ethnical:
I have fixed the previous issues, now I am stuck in the part to generate a correct signature since the threshold need to be `1` at least with the an owner that will match the signatures. 
 
My plan is to: 
- [x] 1. Set the Threshold to of owner of the safe at `1`.
- [x] 2. **Use the anvil cheatcode to set an owner address with a known private keys** (so through the sentinel_address and remove the rest of the signers). In this version we already set the storage of the threshold to `1`.
- [x] 3. Use the `approveHash()` method to approve the hash of the transaction. 
- [x] 4. Execute from a well known EoA (because the hash is already approved) the tx. 

---
---
Alternative plan that we could consider is using the: `vm.createSelectFork('mainnet-rpc')`. 
To avoid to have all the bash and rpc call for anvil. 

 // 1. import all 4 task scripts
 // 2. Create ascript
 // 3. vm.createSelectFork('mainnet-rpc');
 // task1.run()
 // task2.run()
 // task3.run()
 // task4.run()

PR #431, opened 2024-12-20
Add additional pre-checks to the `DisputeGameUpgade`

sebastianst:
**Description**

Adds additional pre-checks to the `DisputeGameUpgade`, in particular comparing the existing and new games' configuration for parity (or differences, where expected).

Stacked on top of  #421 but otherwise good to merge.

**Additional context**

Taken mostly from https://github.com/ethereum-optimism/superchain-ops/pull/410

PR #410, opened 2024-12-13
sepolia-025: Deploy MTCannon to op-sepolia

mbaxter:
**Description**

Add scripts to configure MTCannon dispute games on the op sepolia.

**Tests**

Tested the simulation manually. 

**Metadata**
Part of: https://github.com/ethereum-optimism/optimism/issues/12814

PR #442, opened 2025-01-06
Add additional pre-checks to the `DisputeGameUpgade`

sebastianst:
**Description**

Adds additional pre-checks to the `DisputeGameUpgade`, in particular comparing the existing and new games' configuration for parity (or differences, where expected).

**Additional context**

Taken mostly from https://github.com/ethereum-optimism/superchain-ops/pull/410 and reopening of https://github.com/ethereum-optimism/superchain-ops/pull/431 which got closed by github automatically when I was moving around branches...


script/verification/DisputeGameUpgrade.s.sol
```diff
@@ -86,4 +94,60 @@ abstract contract DisputeGameUpgrade is VerificationBase, SuperchainRegistry {
...
+        require(currentGame.maxGameDepth() == newGame.maxGameDepth(), "pre-60");
+        require(
+            uint64(Duration.unwrap(currentGame.maxClockDuration()))
+                == uint64(Duration.unwrap(newGame.maxClockDuration())),
+            "pre-70"
+        );
+        require(
+            uint64(Duration.unwrap(currentGame.clockExtension())) == uint64(Duration.unwrap(newGame.clockExtension())),
+            "pre-80"
+        );
```
mbaxter, line 132: (nit) Looks like we could also use `val.raw()` instead of `Duration.unwrap(val)` here.

script/verification/DisputeGameUpgrade.s.sol
```diff
@@ -86,4 +94,60 @@ abstract contract DisputeGameUpgrade is VerificationBase, SuperchainRegistry {
 
         require(address(faultDisputeGame.weth()) != address(permissionedDisputeGame.weth()), "weth-200");
     }
+
+    function precheckDisputeGames() internal view {
```
mbaxter, line 98: (nit) It probably makes sense to rename `checkDisputeGameUpgrade` to `postcheckDisputeGames` for clarity.

mbaxter: It also looks to me like most of the post-checks could be pre-checks?  If so, would be nice to clean that up, but can be for a future PR.

tasks/eth/022-holocene-fp-upgrade/.env
```diff
@@ -8,3 +8,6 @@ SAFE_NONCE_0XC2819DC788505AAC350142A7A707BF9D03E3BD03=8  # noop
 SAFE_NONCE_0X847B5C174615B1B7FDF770882256E2D3E95B9D92=11 # +1, task 021
 
 SIMULATE_WITHOUT_LEDGER=0 # 1
+
+# we change the WETH in this task, which is non-standard
+DISPUTE_GAME_CHANGE_WETH=true
```
mbaxter, line 13: Do we need this flag for `base-003-holocene-fp-upgrade` ?

mbaxter:
My main question is about `base-003-holocene-fp-upgrade` - other comments are not blockers.

tasks/eth/022-holocene-fp-upgrade/.env
sebastianst: No, because Base correctly _didn't_ change their DelayedWETH. The default during an upgrade is to not change it, but if it does need to change, setting this overwrite allows it.

script/verification/DisputeGameUpgrade.s.sol
sebastianst: Yes quite a few checks could be done as pre-checks instead. But it doesn't matter too much because all is happening during simulations first. This is how most FP upgrade tasks did it in the past too.

Worth fixing in a follow-up. Would like to get this one in as is.

script/verification/DisputeGameUpgrade.s.sol
sebastianst: nice, fixed.

script/verification/DisputeGameUpgrade.s.sol
sebastianst: I decided to keep the name `checkDisputeGameUpgrade` because it says "check ... upgrade" which implies that it runs checks against an upgrade that has happened. I also want to keep the convention of naming the function `check...` for the function that should be run from the inheriting task contract.

mbaxter (dismissed):

tasks/eth/022-holocene-fp-upgrade/.env
```diff
@@ -8,3 +8,6 @@ SAFE_NONCE_0XC2819DC788505AAC350142A7A707BF9D03E3BD03=8  # noop
 SAFE_NONCE_0X847B5C174615B1B7FDF770882256E2D3E95B9D92=11 # +1, task 021
 
 SIMULATE_WITHOUT_LEDGER=0 # 1
```
Ethnical, line 10: We need to remove this since the just file doesn't check if the value is `0` or `1` cf https://github.com/ethereum-optimism/superchain-ops/issues/448

script/verification/DisputeGameUpgrade.s.sol
```diff
@@ -86,4 +94,53 @@ abstract contract DisputeGameUpgrade is VerificationBase, SuperchainRegistry {
...
+
+    // _precheckDisputeGameImplementation checks that the new game being set has the same
+    // configuration as the existing implementation.
+    function _precheckDisputeGameImplementation(GameType _targetGameType, address _newImpl) internal view {
+        console.log("pre-check new game implementation", _targetGameType.raw());
+
+        FaultDisputeGame currentGame = FaultDisputeGame(address(dgfProxy.gameImpls(GameType(_targetGameType))));
+        FaultDisputeGame newGame = FaultDisputeGame(_newImpl);
+
+        if (vm.envOr("DISPUTE_GAME_CHANGE_WETH", false)) {
```
Ethnical, line 111: seems that this is not ready the value `true` or `false` from the `DISPUTE_GAME_CHANGE_WETH` into the `.env`.
 If this is the case we should comment it to make sure in the future there is no issue failling with it.

Ethnical (dismissed):
It would be nice to remove and add a comment on the DelayedWETH change mechanism + fix the simulation_without_ledger otherwise the rest is looking super nice thanks for the work!

script/verification/DisputeGameUpgrade.s.sol
sebastianst: `envOr` with the second argument being a `bool` is reading the env var value as a boolean, see [docs](https://book.getfoundry.sh/cheatcodes/env-or). Or what do you mean?

tasks/eth/022-holocene-fp-upgrade/.env
sebastianst: Removed and will make sure to not use it/remove it from other tasks as well. Note that I didn't add it in this PR but it got already merged with task 022.

script/verification/DisputeGameUpgrade.s.sol
Ethnical: Thanks for confirming, I was incorrect I have tried this and this is working as expected like in the doc. 
I will resolve this comment.

Ethnical (approved):

PR #449, opened 2025-01-07
bump: the version to the latest version of base-contracts.

Ethnical:
A known bug was causing the task to deploy two contracts instead during the execution of the tasks. 
After investigation, this is coming from a library that contains a method in `public` instead of `internal`. 

cf the this thread on discord of @mds1:  https://discord.com/channels/1244729134312198194/1326232324215668817


Ethnical:
Already merge there
https://github.com/ethereum-optimism/superchain-ops/pull/450

PR #456, opened 2025-01-08
adding ink set respected game type task

sbvegan:
**Description**

This task updates the respected game time on Ink Mainnet's OptimismPortal to use CANNON. This task must not be executed before `eth/ink-001-permissionless-proofs`.

**Additional context**

We ran into a weird issue where the `superchain-ops` repo was not using the latest from `main`. This showed up when I wasn't able to utilize the [ink.toml[(https://github.com/ethereum-optimism/superchain-registry/blob/main/superchain/configs/mainnet/ink.toml) file because the submodule was pinned at a specific branch.

I also needed to update one of the dependencies in the script from what was included in the template. I'll follow up with a commit that fixes that problem.


sbvegan:
See the additional context for why the ci is failing. It is passing locally, but the superchain-registry is pinned to a certain commit for some reason

sbvegan:
wow I can't believe I did this one twice... https://github.com/ethereum-optimism/superchain-ops/pull/407

PR #461, opened 2025-01-13
sep/027: recompute `approvedHashes` keys

geoknee:
Discovered when running through the validation.


Ethnical (approved):
LGTM! 
For posterity, This is happening following the nonce bump.

Issue #397, opened 2024-12-10
Integrate Forge Proposal Simulator (FPS)


blmalone:
https://github.com/solidity-labs-io/forge-proposal-simulator

blmalone:
This was closed with https://github.com/ethereum-optimism/superchain-ops/pull/406.

PR #459, opened 2025-01-13
OPSC Rehearsal setup 

futurealisha:
**Description**

There is a folder under Rehearsals for Rehearsal 1 that needs to be merged — 2025-01-12-r1-hello-council-250112.

PR #467, opened 2025-01-15
[PT: 1] FPS Import - Multisig Proposal

ElliotFriedman:
**Description**

This PR adds the engine that will do all calldata shaping for task creators and allow proposals to be written in plain Solidity. It aims to be as simple as possible and does not include a runnable script to cut down on scope and reduce overhead during review.

- Add AddressRegistry getChains function, remove network addresses files to simplify this PR.
- Add Multisig Proposal.

**Tests**

Testing will be done in a separate PR with templates that execute a previously run task.

**Metadata**

Issues related to these changes are here https://github.com/ethereum-optimism/superchain-ops/issues/397


.gitignore
```diff
@@ -23,4 +23,6 @@ cache/
 # ignore unwanted output from presigning runbooks
 tmp-*.json
 draft*signer*.json
-ready-*
\ No newline at end of file
+ready-*
+
+records/
```
mds1, line 28: We actually recently [changed](https://github.com/ethereum-optimism/superchain-ops/pull/459/files#diff-bac743fe3d899998e32393af53af8cc7d28c89c3d7fabfa48b01361cf8d42d9bL10) our `foundry.toml` to just use the regular `broadcast` default, so this files diff should no longer be needed

src/fps/AddressRegistry.sol
```diff
@@ -3,8 +3,10 @@ pragma solidity 0.8.15;
 import {EnumerableSet} from "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
 
 import {Test} from "forge-std/Test.sol";
+
 import {IAddressRegistry} from "src/fps/IAddressRegistry.sol";
 import {SUPERCHAIN_REGISTRY_PATH} from "src/fps/utils/Constants.sol";
+import {ETHEREUM_CHAIN_ID, SEPOLIA_CHAIN_ID} from "src/fps/utils/Constants.sol";
```
mds1, line 9: Now that I think of it, we probably can reuse forge-std's `StdChains` to avoid having to maintain our own list of constants. And if there are chains forge-std is missing that we need, we can PR those into forge-std. Looks like forge-std doesn't have a way to just fetch chain ID from a chain alias but that's an easy helper method to upstream as well

src/fps/AddressRegistry.sol
```diff
@@ -161,6 +166,12 @@ contract AddressRegistry is IAddressRegistry, Test {
         );
     }
 
+    /// @notice Returns the list of supported superchains
+    /// @return An array of Superchain structs representing the supported superchains
+    function getSuperchains() public view returns (Superchain[] memory) {
+        return superchains;
+    }
```
mds1, line 173: This terminology should just be "chains" and not "superchains". In this context a superchain would be a set of chains that share a config. All chains in a superchain have the same upgrader role (also known as the L1 ProxyAdmin owner, this is the role that can upgrade contracts), and they all use the same SuperchainConfig (meaning if you pause withdrawals on one chain you pause them all). 

A given superchain-ops task would only ever operate on a set of chains that are all in the same superchain. Specifically, there is one root multisig (such as the upgrader role) that is sending `m` transactions to the `n` chains in a single task, since that root multisig is (for example) the upgrader role for all those chains.

I know it can be a bit confusing, so let me know if that makes sense!

src/fps/addresses/1.toml
```diff
@@ -0,0 +1,4 @@
+[[addresses]]
+addr = "0x240B3bd6b95cE40497Aafd71aD4705d0345A33CD"
+identifier = "SYSTEM_CONFIG_LOGIC"
+isContract = true
```
mds1, line 4: In the spirit of avoiding hardcoding addresses as much as possible, do we have to hardcode these logic addresses? They can be read onchain so we should be able to delete this. We often upgrade the SystemConfig logic address, and now this file becomes outdated

src/fps/addresses/10.toml
```diff
@@ -1,14 +1,34 @@
...
+
+[[addresses]]
+addr = "0xc641A33cab81C559F2bd4b21EA34C290E2440C2B"
+identifier = "DISPUTE_GAME_FACTORY_LOGIC"
+isContract = true
+
+[[addresses]]
+addr = "0xf691F8A6d908B58C534B624cF16495b491E633BA"
+identifier = "FAULT_DISPUTE_GAME_IMPLEMENTATION"
+isContract = true
```
mds1, line 14: One thing that is a bit confusing is that these addresses all exist on L1, but the file is `10.toml`, making it seem like these addresses exist on L2. I think these need to exist in `1.toml`, but `1.toml` supports specifying which L2s a given address is used for.

But similar to above comment, it feels like none of these addresses need to be hardcoded this file, as they are all discoverable onchain starting from the data in the superchain registry

src/fps/example/IDisputeGameFactory.sol
mds1, line 1: Instead of redeclaring interfaces, we should import them from the Optimism monorepo submodule (we may need to update that dep, and we should just ensure this doesn't cause problems for any existing tasks—I recall @sebastianst or @geoknee may have flagged an issue preventing us from easily upgrading it?)

src/fps/AddressRegistry.sol
mds1: I would say this change is low priority and we shouldn't block on it. So I created https://github.com/ethereum-optimism/superchain-ops/issues/469 to track

@blmalone Can you update the tracking issue with this?

src/fps/addresses/10.toml
mds1: > You're saying that for each superchain we are talking to, we can query it's superchain registry onchain and see this and other addresses. We will look into this.

Exactly. The most minimal form of this is that you can uniquely identify a chain by it's SystemConfig and discover all other addresses. To reduce RPC calls, instead of discovering *everything* from the SystemConfig, we can read in everything from the `addresses` key of a given `superchain/configs/{L1}/{L2}.toml`. https://github.com/ethereum-optimism/superchain-registry/blob/e2d3490729b20a649281899c2c286e6e12db57f3/superchain/configs/mainnet/op.toml#L38-L60

Regarding `AddressManager`, that is a legacy contract that we only use in the L1CrossDomainMessenger as you've noticed. Basically, for legacy reasons, we have 3 kinds of proxies: `Proxy.sol` is the regular one, `ResolvedDelegateProxy` (RDP) is a unique kind used for the L1CrossDomainMessenger, and `L1ChugsplashProxy` is a unique one for the L1StandardBridge. The RDP uses an AddressManager to look up the implementation, so you are right that AddressManager will therefore only contain the implementation address for the L1CrossDomainMessenger

To find implementations, the logic is basically:
- If the contract is an L1CrossDomainMessenger, go through the AddressManager to find the implementation
- Otherwise, read the standard EIP-1967 storage slot: https://github.com/ethereum-optimism/optimism/blob/547ea72d9849e13ce169fd31df0f9197651b3f86/packages/contracts-bedrock/src/libraries/Constants.sol#L26

src/fps/AddressRegistry.sol
blmalone: I've added to the parent tracking issue: https://github.com/ethereum-optimism/superchain-ops/issues/396

src/fps/proposal/MultisigProposal.sol
```diff
@@ -0,0 +1,968 @@
...
+
+    /// @notice Sets the L2 networks configuration
+    /// @param networkConfigFilePath Path to the network configuration file
+    /// @param _addresses Address registry contract
+    function setL2NetworksConfig(string memory networkConfigFilePath, Addresses _addresses) public override {
+        addresses = _addresses;
+        string memory networkConfigFileContents = vm.readFile(networkConfigFilePath);
+
+        nonce = abi.decode(vm.parseToml(networkConfigFileContents, ".safeNonce"), (uint256));
+        isNestedSafe = abi.decode(vm.parseToml(networkConfigFileContents, ".isNestedSafe"), (bool));
```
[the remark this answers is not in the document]
blmalone: thanks, I've added to the parent tracking [issue](https://github.com/ethereum-optimism/superchain-ops/issues/396)

src/fps/utils/Constants.sol
```diff
@@ -20,7 +24,23 @@ uint256 constant METAL_SEPOLIA_CHAIN_ID = 1740;
...
+// offset for the nonce variable in Gnosis Safe
+bytes32 constant NONCE_OFFSET = 0x0000000000000000000000000000000000000000000000000000000000000005;
+
+bytes32 constant SAFE_NONCE_SLOT = bytes32(uint256(5));
+
+// the amount of modules to fetch from the Gnosis Safe
+uint256 constant MODULES_FETCH_AMOUNT = 1_000;
+
+// storage slot for the fallback handler keccak256("fallback_manager.handler.address")
+bytes32 constant FALLBACK_HANDLER_STORAGE_SLOT = 0x6c9a6c4a39284e37ed1cf53d337577d14212a4870fb976a4366c693b939918d5;
```
blmalone, line 46: This constant doesn't seem to be used.

src/fps/utils/Constants.sol
```diff
@@ -20,7 +24,23 @@ uint256 constant METAL_SEPOLIA_CHAIN_ID = 1740;
...
+
+address constant MULTICALL3_ADDRESS = 0xcA11bde05977b3631167028862bE2a173976CA11;
+
+// offset for the nonce variable in Gnosis Safe
+bytes32 constant NONCE_OFFSET = 0x0000000000000000000000000000000000000000000000000000000000000005;
+
+bytes32 constant SAFE_NONCE_SLOT = bytes32(uint256(5));
+
+// the amount of modules to fetch from the Gnosis Safe
+uint256 constant MODULES_FETCH_AMOUNT = 1_000;
```
blmalone, line 43: This constant doesn't seem to be used.

src/fps/utils/Constants.sol
```diff
@@ -20,7 +24,23 @@ uint256 constant METAL_SEPOLIA_CHAIN_ID = 1740;
...
 uint256 constant LOCAL_CHAIN_ID = 31337;
 
+string constant ADDRESSES_PATH = "src/fps/addresses";
+
 string constant SUPERCHAIN_REGISTRY_PATH = "lib/superchain-registry/superchain/extra/addresses/addresses.json";
+
+address constant MULTICALL3_ADDRESS = 0xcA11bde05977b3631167028862bE2a173976CA11;
+
+// offset for the nonce variable in Gnosis Safe
+bytes32 constant NONCE_OFFSET = 0x0000000000000000000000000000000000000000000000000000000000000005;
```
blmalone, line 38: This constant doesn't seem to be used.

src/fps/utils/Constants.sol
```diff
@@ -20,7 +24,23 @@ uint256 constant METAL_SEPOLIA_CHAIN_ID = 1740;
 uint256 constant RACE_SEPOLIA_CHAIN_ID = 6806;
 uint256 constant ZORA_SEPOLIA_CHAIN_ID = 999999999;
 uint256 constant OPLABS_DEVNET_CHAIN_ID = 11155421;
+uint256 constant BINARY_SEPOLIA_CHAIN_ID = 625;
 
 uint256 constant LOCAL_CHAIN_ID = 31337;
 
+string constant ADDRESSES_PATH = "src/fps/addresses";
```
blmalone, line 31: This constant doesn't seem to be used.

src/fps/utils/BytesHelper.sol
```diff
@@ -0,0 +1,13 @@
+pragma solidity 0.8.15;
+
+library BytesHelper {
+    /// @notice function to grab the first 32 bytes of returned memory
+    /// @param toSlice the calldata payload
+    function getFirstWord(bytes memory toSlice) public pure returns (uint256 value) {
```
blmalone, line 6: I don't think this function is used in MultisigProposal.sol, so can we remove this library and introduce it again in the future if we need it?

src/fps/proposal/IProposal.sol
```diff
@@ -0,0 +1,62 @@
+pragma solidity 0.8.15;
+
+import {AddressRegistry as Addresses} from "src/fps/AddressRegistry.sol";
+
+abstract contract IProposal {
```
blmalone, line 5: I wonder do we want to call these things 'Tasks' instead of 'Proposals'. I know this terminology is brought in from FPS so I understand it's origins.

Anyway, I don't think the change needs to be made in this PR. Just a callout for the future. Internally we don't really use the term 'proposal' for these things.

src/fps/proposal/IProposal.sol
```diff
@@ -0,0 +1,62 @@
...
+    /// @notice proposal name, e.g. "BIP15".
+    /// @dev override this to set the proposal name.
+
+    function name() external view virtual returns (string memory);
+
+    /// @notice proposal description.
+    /// @dev override this to set the proposal description.
```
blmalone, line 12: I don't even think we will ever use this, right? Our tasks aren't onchain and the `README.md` files associated with each task should give us all the information we need to understand the context of the given upgrade task.

src/fps/proposal/IProposal.sol
```diff
@@ -0,0 +1,62 @@
...
+        virtual
+        returns (address[] memory targets, uint256[] memory values, bytes[] memory arguments);
+
+    /// @notice return contract identifiers whose storage is modified by the proposal
+    function getAllowedStorageAccess() external view virtual returns (address[] memory);
+
+    /// @notice return proposal calldata
+    function getCalldata() external virtual returns (bytes memory data);
+
+    /// @notice helper function to mock on-chain data
```
blmalone, line 33: For my own understanding, what's the purpose of the mock function? Can you explain an example where I might use this? 
I _don't think_ we have this functionality right now in the superchain-ops repo.

src/fps/proposal/MultisigProposal.sol
```diff
@@ -0,0 +1,854 @@
...
+import {
+    NONCE_OFFSET,
+    SAFE_NONCE_SLOT,
+    MODULES_FETCH_AMOUNT,
+    FALLBACK_HANDLER_STORAGE_SLOT,
+    MULTICALL3_ADDRESS
+} from "src/fps/utils/Constants.sol";
+
+abstract contract MultisigProposal is Test, Script, IProposal {
+    using BytesHelper for bytes;
```
blmalone, line 23: Not sure this is actually used.

src/fps/proposal/MultisigProposal.sol
```diff
@@ -0,0 +1,854 @@
+pragma solidity 0.8.15;
```
blmalone, line 1: should we be putting the license in these files? 

e.g. 
```
// SPDX-License-Identifier: MIT
```
```suggestion
// SPDX-License-Identifier: MIT
pragma solidity 0.8.15;
```

src/fps/proposal/IProposal.sol
```diff
@@ -0,0 +1,62 @@
+pragma solidity 0.8.15;
```
blmalone, line 1: Do we need the license in these files? 
```suggestion
// SPDX-License-Identifier: MIT
pragma solidity 0.8.15;
```

src/fps/proposal/MultisigProposal.sol
```diff
@@ -0,0 +1,854 @@
...
+    /// they all follow the same structure
+    Action[] public actions;
+
+    /// @notice proposal name, e.g. "OIP15".
+    /// @dev set in the proposal config file
+    string public override name;
+
+    /// @notice proposal description.
+    /// @dev set in the proposal config file
+    string public override description;
```
blmalone, line 129: Similar comment as before, our tasks aren't stored on chain and our docs should contain all the info/context a dev needs to understand the scope of an upgrade.

src/fps/proposal/MultisigProposal.sol
```diff
@@ -0,0 +1,854 @@
...
+                    vm.getLabel(addresses.getAddress(config.safeAddressString, chains[i].chainId))
+                )
+            );
+        }
+
+        /// Fetch starting owners
+        IGnosisSafe safe = IGnosisSafe(multisig);
+        startingOwners = safe.getOwners();
+
+        for (uint256 i = 0; i < config.allowedStorageWriteAccesses.length; i++) {
```
blmalone, line 238: Some comments around why we need this would be useful. We use this later in a state comparison right? Would be good to get that detail in here.

src/fps/proposal/MultisigProposal.sol
```diff
@@ -0,0 +1,854 @@
...
+                    addresses.getAddress(config.allowedStorageWriteAccesses[i], chains[j].chainId)
+                );
+            }
+        }
+    }
+
+    /// @notice function to be used by forge script.
+    /// @dev use flags to determine which actions to take
+    ///      this function shoudn't be overriden.
+    function _processProposal() internal override {
```
blmalone, line 250: Thinking back to an earlier comment on naming here, 'processTask' feels more accurate but I don't want this to be a blocker. Just throwing this idea around as something to be considered later.

src/fps/proposal/MultisigProposal.sol
```diff
@@ -0,0 +1,854 @@
...
+        mock();
+        build();
+        simulate();
+        validate();
+        print();
+    }
+
+    /// @notice get the calldata to be executed by safe
+    /// @return data The calldata to be executed
+    function getCalldata() public view override returns (bytes memory data) {
```
blmalone, line 260: Seems like this function is only callable when the actions have been set. Otherwise it reverts with `No actions found`. Is it possible to make the function description a little clearer on when this function would be eligible for successful execution.  Seems like `build();` _must_ be called first.

src/fps/proposal/MultisigProposal.sol
```diff
@@ -0,0 +1,854 @@
...
+    }
+
+    /// @notice Sets the L2 networks configuration
+    /// @param networkConfigFilePath Path to the network configuration file
+    /// @param _addresses Address registry contract
+    function setL2NetworksConfig(string memory networkConfigFilePath, Addresses _addresses) public override {
+        addresses = _addresses;
+        string memory networkConfigFileContents = vm.readFile(networkConfigFilePath);
+
+        nonce = abi.decode(vm.parseToml(networkConfigFileContents, ".safeNonce"), (uint256));
```
blmalone, line 212: IIRC `parseToml` will fail if the nonce is not defined right? Just curious what happens in this case.

src/fps/proposal/MultisigProposal.sol
```diff
@@ -0,0 +1,854 @@
...
+            gasToken: address(0),
+            refundReceiver: address(0),
+            _nonce: nonce
+        });
+    }
+
+    /// @notice simulate the proposal by approving from owners and then executing
+    function simulate() public override {
+        bytes memory data = getCalldata();
+        bytes32 hash = keccak256(_getDataToSign(multisig, data));
```
blmalone, line 309: This logic happens a bunch of times in this file, is there an opportunity to refactor this into a function?

src/fps/proposal/MultisigProposal.sol
```diff
@@ -0,0 +1,854 @@
...
+        Call3Value[] memory calls = new Call3Value[](1);
+        calls[0] = call;
+        return abi.encodeWithSignature("aggregate3Value((address,bool,uint256,bytes)[])", calls);
+    }
+
+    /// @notice helper function to prepare the signatures to be executed
+    /// @param _safe The address of the parent multisig
+    /// @param hash The hash to be approved
+    /// @return The signatures to be executed
+    function prepareSignatures(address _safe, bytes32 hash) internal view returns (bytes memory) {
```
blmalone, line 615: Base has a Signatures.sol library for this that maybe we can reuse here to reduce the size of this file. https://github.com/base-org/contracts/blob/main/script/universal/Signatures.sol

src/fps/proposal/MultisigProposal.sol
```diff
@@ -0,0 +1,854 @@
...
+        return subset;
+    }
+
+    /// --------------------------------------------------------------------
+    /// --------------------------------------------------------------------
+    /// ------------------------- Private functions ------------------------
+    /// --------------------------------------------------------------------
+    /// --------------------------------------------------------------------
+
+    /// @notice to be used by the build function to create a governance proposal
```
blmalone, line 671: I'm not sure if this language is accurate for us. 

What makes more sense to me would be something like:

```
to be used by the build function to capture the state changes applied by a given task. These state changes will inform whether or not the task will be executed onchain.
```

blmalone (commented):
I've given this PR an initial pass. I'll come back to it for another review soon.

src/fps/proposal/MultisigProposal.sol
mds1: It looks like this code is copy/pasted from the base repo, so we may as well just continue using the base repo as a dep to make this PR easier to read and this file easier to maintain. We don't need to move away from deps just to reduce the number of dependencies, we only need to move away from deps if doing so solves a problem for us

src/fps/proposal/MultisigProposal.sol
mds1: Sounds like a good test case :)

src/fps/proposal/IProposal.sol
mds1: +1 for tasks in a future PR

src/fps/proposal/IProposal.sol
mds1: > we did need to mock some onchain data for one of the tasks around safe modules

I believe that was a misunderstanding of either how to fetch the modules from the safe, or what the expected module address was: https://discord.com/channels/1244729134312198194/1301293771153080331/1326611480585699328

Let's get to the e2e MVP first before we think about what the code and UX should look like for simulating sequences of tasks though

src/fps/utils/Constants.sol
blmalone: resolving

src/fps/utils/Constants.sol
blmalone: resolving

src/fps/utils/BytesHelper.sol
blmalone: resolving

src/fps/proposal/IProposal.sol
blmalone: Tracking this [here](https://github.com/ethereum-optimism/superchain-ops/issues/475) and resolving.

src/fps/proposal/IProposal.sol
blmalone: I see this file has been removed, resolving this thread.

src/fps/proposal/MultisigProposal.sol
blmalone: Awesome - I'll resolve this thread and look for more testing in a later PR.

src/fps/proposal/MultisigProposal.sol
blmalone: Ah, I was thinking more about this being reused in multiple places:
```
bytes32 hash = keccak256(_getDataToSign(multisig, getCalldata()))
```

A useful function might be:

```solidity
    function getHash() internal view returns (bytes32, bytes memory) {
        bytes memory data = getCalldata();
        bytes32 hash = keccak256(_getDataToSign(multisig, data));
        return (hash, data);
    }
```

src/fps/proposal/MultisigProposal.sol
blmalone: Nice, I think it would be nice to keep this file as minimal as possible. Being able to reason about the logic in here easily will pay dividends down the line as we want to add more features.

src/fps/proposal/IProposal.sol
blmalone: Getting rid of the mock functionality for now. If we find that we need it going forward we can add it back in.

src/fps/proposal/MultisigProposal.sol
blmalone: Awesome, resolving.

src/fps/proposal/MultisigProposal.sol
blmalone: This is completed so resolving this comment.

blmalone (approved):
LGTM, nice work.

PR #454, opened 2025-01-08
eth/027: upgrade and re-initialize SystemConfig 

geoknee:
~Stacked on #418~
Reuses the `HoloceneSystemConfigUpgrade` contract committed in #418.

The tx bundle was generated with the code on https://github.com/ethereum-optimism/optimism/pull/13504. 

Closes https://github.com/ethereum-optimism/optimism/issues/13354

PR #476, opened 2025-01-21
sep/028: upgrade and re-initialize Base's SystemConfig to v2.3.0

xenoliss:
**Description**

Upgrade Base's `SystemConfig` on Sepolia to v2.3.0.
Duplicate of https://github.com/ethereum-optimism/superchain-ops/pull/454 with slight modifications for Base.

**Additional context**

Base specifics:
1. Added `checkSystemConfigUpgradeWithPreviousGasLimitOverride` to `script/verification/HoloceneSystemConfigUpgrade.s.sol` to allow updating the `gasLimit` which will be needed because Base `_gasLimit` will change on mainnet (PR to come). On Sepolia the `_gasLimit` is kept as is.
2. Whitelisted the `2.2.0+max-gas-limit-400M` version in `script/verification/HoloceneSystemConfigUpgrade.s.sol`.
3. Manually added a `SystemConfigProxy.setEIP1559Params(1,4)` call to the end of the `tasks/sep/028-holocene-system-config-upgrade-and-init-base/input.json` via `just add-transaction tasks/sep/028-holocene-system-config-upgrade-and-init-base/input.json 0xf272670eb55e895584501d564AfEB048bEd26194 'setEIP1559Params(uint32,uint32)' 1 4`


tasks/sep/028-holocene-system-config-upgrade-and-init-base/VALIDATION.md
```diff
@@ -0,0 +1,36 @@
...
+  **After**:  `0x00000000000000000000000033b83e4c305c908b2fc181dda36e230213058d7d`
+  **Meaning**: Updates the `SystemConfig`'s implementation to version 2.3.0 at `0x33b83E4C305c908B2Fc181dDa36e230213058d7d`.
+
+- **Key**: `0x0000000000000000000000000000000000000000000000000000000000000068`
+  **Before**: `0x0000000000000000000000000000000000000000000000000000000003938700`
+  **After**: `0x00000000000000000000000000000000000a118b0000044d0000000003938700`
+  **Meaning**: Sets the new `SystemConfig`'s variables `blobbasefeeScalar` to `659851` and `basefeeScalar` to `1101`.
+
+- **Key**: `0x000000000000000000000000000000000000000000000000000000000000006a`
+  **Before**: `0x000000000000000000000000000000000000000000000000000000000042b1d7`
```
xenoliss, line 34: Slot 106 was previously used to store a `startBlock` in previous versions which is why it has a non-zero value of 0x42b1d7 (4370903)

tasks/sep/028-holocene-system-config-upgrade-and-init-base/SignFromJson.s.sol
```diff
@@ -0,0 +1,70 @@
...
+        //
+        //  2. sets the owner back to the OwnerSafe multisig when calling `transferOwnership(_owner)`
+        //     https://github.com/ethereum-optimism/optimism/blob/39e9f1a4693912af8bdef428d246c15a6cf44ec7/packages/contracts-bedrock/src/L1/SystemConfig.sol#L172
+        //
+        //  /!\ For this reason we need NOT to include the code exception for the OwnerSafe multisig /!\
+        address[] memory codeExceptions = sysCfgUpgrade.getCodeExceptions();
+        for (uint256 i; i < codeExceptions.length; i++) {
+            if (codeExceptions[i] == ownerSafe) {
+                continue;
+            }
```
geoknee, line 44: Curious about this. Is it that we don't need to include the `ownerSafe` here, or that we **need to not** include the `ownerSafe` here? Is there something that makes this task different from original multichain task this was based on (in this particular way)? I would have thought the ownership transfer gymnastics would have been the same? Also, adding an exception would usually not cause the underlying validation check to fail -- so confused about what is going on here.

tasks/sep/028-holocene-system-config-upgrade-and-init-base/SignFromJson.s.sol
```diff
@@ -0,0 +1,70 @@
...
+pragma solidity ^0.8.15;
+
+import {console2 as console} from "forge-std/console2.sol";
+import {Vm} from "forge-std/Vm.sol";
+import {Simulation} from "@base-contracts/script/universal/Simulation.sol";
+import {SignFromJson as OriginalSignFromJson} from "script/SignFromJson.s.sol";
+import {ISystemConfig, HoloceneSystemConfigUpgrade} from "script/verification/HoloceneSystemConfigUpgrade.s.sol";
+import {VerificationBase} from "script/verification/Verification.s.sol";
+
+contract SignFromJson is OriginalSignFromJson, VerificationBase {
```
geoknee, line 11: It was my understanding that this task ought to be a `NestedSignFromJson is OriginalNestedSignFromJson`, because the `ProxyAdminOwner` is a "safe of safes".

xenoliss: [On Sepolia](https://sepolia.etherscan.io/address/0x0fe884546476dDd290eC46318785046ef68a0BA9#readProxyContract), we don't use a nested multisig and instead the `OWNER_SAFE` is directly owned by some EOAs.

tasks/sep/028-holocene-system-config-upgrade-and-init-base/SignFromJson.s.sol
xenoliss: I was trying to replicate the simple logic [here](https://github.com/ethereum-optimism/superchain-ops/blob/main/tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/NestedSignFromJson.s.sol#L29).

The issue in our specific case is that simply doing `addCodeExceptions(sysCfgUpgrades[i].getCodeExceptions())` does not work as it adds a [code exception on Sepolia for the `previous.owner`](https://github.com/ethereum-optimism/superchain-ops/blob/main/script/verification/HoloceneSystemConfigUpgrade.s.sol#L98) (not sure why).

My logic was to get all code exceptions returned by `sysCfgUpgrade.getCodeExceptions()` and simply skip the one relative to the `ownerSafe`.

geoknee: Your logic makes sense. 

I don't think the codeExceptions API is well documented at all. By adding an address to the scripts `codeExceptions`, I realise now that we are telling the verification system that: if any storage slot is updated to that address, the address _should have no code_.  Which is why doing so with the `ownerSafe` causes verification to fail (it's a safe, it has code).  

(Previously I had thought that adding the exception would just skip the usual check that if the address is found in a storage access, it should have some code). 

We already have some special casing for this in the base `HoloceneSystemConfigUpgrade` contract (the superchain SystemConfigOwner is a safe on mainnet but an EOA on sepolia). So I think we should consolidate the exceptions there. I can make a PR with what I have in mind.

tasks/sep/028-holocene-system-config-upgrade-and-init-base/.env
```diff
@@ -0,0 +1,5 @@
+ETH_RPC_URL="https://ethereum-sepolia.publicnode.com"
+OWNER_SAFE=0x0fe884546476dDd290eC46318785046ef68a0BA9
```
geoknee, line 2: Confirmed this with the superchain registry 👍

tasks/sep/028-holocene-system-config-upgrade-and-init-base/OVERVIEW.md
```diff
@@ -0,0 +1,57 @@
...
+
+**Raw Input Data:** `0x4e91db0800000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000`
+
+### Inputs
+**_slot:** `0x0000000000000000000000000000000000000000000000000000000000000000`
+
+**_value:** `0x0000000000000000000000000000000000000000000000000000000000000000`
+
+
+## Tx #3: ProxyAdmin.upgradeAndCall(SystemConfigProxy, SystemConfigImplementation, Initialize())
```
geoknee, line 40: Can you add Tx 4?

tasks/sep/028-holocene-system-config-upgrade-and-init-base/README.md
```diff
@@ -0,0 +1,31 @@
+# Holocene Hardfork Upgrade - `SystemConfig`
+
+Status: READY TO SIGN
+
+## Objective
+
+Upgrades the `SystemConfig` for the Holocene hardfork.
+
+This upgrades the `SystemConfig` in the
```
geoknee, line 9: Perhaps somewhere here you could commit some of the context around this being a somewhat special case for base (with the extra transaction).

script/verification/HoloceneSystemConfigUpgrade.s.sol
```diff
@@ -94,6 +97,16 @@ contract HoloceneSystemConfigUpgrade is SuperchainRegistry, VerificationBase {
...
+    /// @notice Public function that must be called by the verification script.
+    function checkSystemConfigUpgradeWithPreviousGasLimitOverride(uint256 previousGasLimitOverride) public view {
+        checkTargetVersion();
+        checkScalar();
+        checkDGF();
+        checkGasPayingToken();
+
+        checkBaseSysCfgVarsWithGasLimitOverride(previousGasLimitOverride);
+    }
+
```
geoknee, line 109: I think your overall approach here is fine, but am I right in understanding that you are committing this code without actually using it yet? If so, might be best to move this to the next (mainnet) PR.

geoknee (commented):
Looking good so far. Got one suggestion for you here https://github.com/ethereum-optimism/superchain-ops/pull/478

I reviewed this mostly from the perspective of the `HoloceneSystemConfigUpgrade` framework point of view. I did not yet check the simulation results in depth.

tasks/sep/028-holocene-system-config-upgrade-and-init-base/VALIDATION.md
geoknee: Do we have any links to the source code for the version of the contract you are upgrading _from_?

xenoliss: Yes, for both Sepolia and Mainnet the current version is [this patch](https://github.com/ethereum-optimism/optimism/blob/23fe5994587a636a1656c9f4c975529472030d2d/packages/contracts-bedrock/src/L1/SystemConfig.sol). 

The storage slot 106 on Sepolia was set previously (from an old version that I don't know too much about) and the initializer did not clean it that's why it is non-zero. On Mainnet this slot is set to 0 though.

script/verification/HoloceneSystemConfigUpgrade.s.sol
xenoliss: Yes you are right this will only be used in for the Mainnet task. Moving this change to the next PR.

tasks/sep/028-holocene-system-config-upgrade-and-init-base/README.md
xenoliss: Added some context. Note that the extra tx is really only needed on Mainnet to match with our current 252M gas limit. On Sepolia we had a gas limit of 60M which we are okay to increase here (via an `_elasticity` of 4).

script/verification/HoloceneSystemConfigUpgrade.s.sol
```diff
@@ -176,6 +181,16 @@ contract HoloceneSystemConfigUpgrade is SuperchainRegistry, VerificationBase {
...
+    // Checks the remaining storage variables are unchanged after the upgrade.
+    // NOTE: The `_gasLimit` is permitted to be updated to accommodate the new EIP-1559 parameters.
+    function checkBaseSysCfgVarsWithGasLimitOverride(uint256 previousGasLimitOverride) internal view {
+        BaseSysCfgVars memory _previous = previous;
+        _previous.gasLimit = previousGasLimitOverride;
+
+        // Check remaining storage variables didn't change
+        require(keccak256(abi.encode(getBaseSysCfgVars())) == keccak256(abi.encode(_previous)), "system-config-110.5");
+    }
+
```
geoknee, line 193: Let's move this extra code to the next PR?

script/verification/HoloceneSystemConfigUpgrade.s.sol
```diff
@@ -72,7 +72,10 @@ contract HoloceneSystemConfigUpgrade is SuperchainRegistry, VerificationBase {
         // Read target version from SCR @ specified release
         targetVersion = standardVersions.SystemConfig.version;
 
-        if (sysCfg.version().eq("2.3.0") || sysCfg.version().eq("2.2.0")) {
+        if (
+            sysCfg.version().eq("2.3.0") || sysCfg.version().eq("2.2.0")
+                || sysCfg.version().eq("2.2.0+max-gas-limit-400M")
+        ) {
```
geoknee, line 78: I don't want to risk changes to this file upsetting in-flight tasks. In particular `eth/028`. @Ethnical do you think it might change the approvedHash of that task? The safest thing to do would be to wait until `eth/028` is executed, which I believe is due to happen in the next day or so.

Ethnical: This **can't** change the `approvalHash` to changes the approvalHash we need to modified the current system on L1 since the task has been signed. 

However, this can decrease the security checks if a bug is introduce into the current checks if a tests is missed for some reason.

I suggest for resolving both of the concern to override this function and apply the check only for this task @geoknee what do you think about it?

geoknee: I think it's cleaner to wait for eth/028 to be executed and continue with this PR mostly as is. The changes to the base contract make sense here, I think they are very unlikely to reduce the security of future tasks as long as we do a thorough job reviewing (and they make the base contract significantly more reusable).

tasks/sep/028-holocene-system-config-upgrade-and-init-base/SignFromJson.s.sol
```diff
@@ -0,0 +1,54 @@
...
+
+    function _postCheck(Vm.AccountAccess[] memory accesses, Simulation.Payload memory) internal view override {
+        console.log("Running post-deploy assertions");
+
+        checkStateDiff(accesses);
+        sysCfgUpgrade.checkSystemConfigUpgrade();
+
+        ISystemConfig systemConfig = ISystemConfig(sysCfgUpgrade.systemConfigAddress());
+        vm.assertEq(systemConfig.eip1559Denominator(), 1, "invalid EIP1559 denominator");
+        vm.assertEq(systemConfig.eip1559Elasticity(), 4, "invalid EIP1559 elasticity");
```
geoknee, line 42: Nit: is it better to say "incorrect" rather than "invalid" here?

tasks/sep/028-holocene-system-config-upgrade-and-init-base/OVERVIEW.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Tx #4: SystemConfigProxy.setEIP1559Params(1,4)
+Sets the EIP1559 parameters: `_denominator = 1`, `_elasticity_ = 4`
```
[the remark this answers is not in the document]
geoknee: Answering my own question -- because for this chain, the gas limit is being reduced from 400M to 200M, some changes to the EIP1559 params are required to restore existing functionality. Could you provide details of how you calculated `1,4` as the right values to do that?

tasks/sep/028-holocene-system-config-upgrade-and-init-base/VALIDATION.md
```diff
@@ -0,0 +1,36 @@
...
+  **Meaning**: Increments the `GnosisSafeProxy`'s `nonce`.
+
+### `0xf272670eb55e895584501d564AfEB048bEd26194` (`SystemConfigProxy`)
+
+- **Key**: `0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc`
+  **Before**: `0x00000000000000000000000031aad1062c25ce545c14bd9ee64dedef6c6b6fac`
+  **After**:  `0x00000000000000000000000033b83e4c305c908b2fc181dda36e230213058d7d`
+  **Meaning**: Updates the `SystemConfig`'s implementation to version 2.3.0 at `0x33b83E4C305c908B2Fc181dDa36e230213058d7d`.
+
+- **Key**: `0x0000000000000000000000000000000000000000000000000000000000000068`
```
geoknee, line 28: Nit: the formatting doesn't quite work when rendered, can you add some line breaks or more bullets or something?

tasks/sep/028-holocene-system-config-upgrade-and-init-base/VALIDATION.md
```diff
@@ -0,0 +1,36 @@
...
+  **Before**: `0x0000000000000000000000000000000000000000000000000000000003938700`
+  **After**: `0x00000000000000000000000000000000000a118b0000044d0000000003938700`
+  **Meaning**: Sets the new `SystemConfig`'s variables `blobbasefeeScalar` to `659851` and `basefeeScalar` to `1101`.
```
geoknee, line 31: Nit, it's great if you can show how to validate the decimals against the hex, like this https://github.com/ethereum-optimism/superchain-ops/blob/d556d220609202f85412d6f0e97f0b0f1c97260a/tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain/VALIDATION.md?plain=1#L44

tasks/sep/028-holocene-system-config-upgrade-and-init-base/SignFromJson.s.sol
geoknee: Resolved here [a037e61](https://github.com/ethereum-optimism/superchain-ops/pull/476/commits/a037e61fa46e6dc331e0841b0865e2930c3b6324)

tasks/sep/028-holocene-system-config-upgrade-and-init-base/VALIDATION.md
```diff
@@ -0,0 +1,36 @@
...
+This document can be used to validate the state diff resulting from the execution of the upgrade
+transaction.
+
+For each contract listed in the state diff, please verify that no contracts or state changes shown in the Tenderly diff are missing from this document. Additionally, please verify that for each contract:
+
+- The following state changes (and none others) are made to that contract. This validates that no unexpected state changes occur.
+- All addresses (in section headers and storage values) match the provided name, using the Etherscan and Superchain Registry links provided. This validates the bytecode deployed at the addresses contains the correct logic.
+- All key values match the semantic meaning provided, which can be validated using the storage layout links provided.
+
+## State Changes
```
mds1, line 12: For mainnet, let's make sure to include the state overrides as well so signers can verify those

tasks/sep/028-holocene-system-config-upgrade-and-init-base/VALIDATION.md
```diff
@@ -0,0 +1,36 @@
...
+
+### `0x0fe884546476dDd290eC46318785046ef68a0BA9` (`GnosisSafeProxy`)
+
+- **Key**: `0x0000000000000000000000000000000000000000000000000000000000000005`
+  - **Before**: `0x000000000000000000000000000000000000000000000000000000000000000b`
+  - **After**: `0x000000000000000000000000000000000000000000000000000000000000000c`
+  - **Meaning**: Increments the `GnosisSafeProxy`'s `nonce`.
```
mds1, line 19: it will render better this way, let's make sure to apply this to all the other state changes below too
```suggestion
- **Key**: `0x0000000000000000000000000000000000000000000000000000000000000005`<br>
- **Before**: `0x000000000000000000000000000000000000000000000000000000000000000b`<br>
- **After**: `0x000000000000000000000000000000000000000000000000000000000000000c`<br>
- **Meaning**: Increments the `GnosisSafeProxy`'s `nonce`.
```

tasks/sep/028-holocene-system-config-upgrade-and-init-base/OVERVIEW.md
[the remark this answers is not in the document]
mds1: We should add a few sentences to the README explaining the before/after situation and justifying the 1559 and gas limit values. I think this is something that is going to be confusing to both PR reviewers and signers

tasks/sep/028-holocene-system-config-upgrade-and-init-base/OVERVIEW.md
mds1, line 1: I usually advocate for deleting this file and therefore have not reviewed it. The `input.json` and state diff is the source of truth, this is just another file that can become out of date

mds1 (commented):
State diffs look good to me, though I don't yet understand the rationale for everything (see https://github.com/ethereum-optimism/superchain-ops/pull/476#discussion_r1928792073). So will wait for those questions and comments above to be resolved before approving, but generally see no major issues here

script/verification/HoloceneSystemConfigUpgrade.s.sol
geoknee: eth/028 has now been executed.

xenoliss: See https://github.com/ethereum-optimism/superchain-ops/pull/476/commits/1faf8520c82fa48e1aaefe38f75ac6f6f022ba35

tasks/sep/028-holocene-system-config-upgrade-and-init-base/README.md
xenoliss: See https://github.com/ethereum-optimism/superchain-ops/pull/476/commits/1faf8520c82fa48e1aaefe38f75ac6f6f022ba35.

tasks/sep/028-holocene-system-config-upgrade-and-init-base/OVERVIEW.md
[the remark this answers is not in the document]
xenoliss: Does https://github.com/ethereum-optimism/superchain-ops/pull/476/commits/1faf8520c82fa48e1aaefe38f75ac6f6f022ba35 answer this?

tasks/sep/028-holocene-system-config-upgrade-and-init-base/VALIDATION.md
xenoliss: Doing it right now but I would argue that having the nested tabs below a `key` helps reading, unless I'm missing something:

Currently:
image

With the recommended change:
image

mds1 (approved):

geoknee (approved):

geoknee (dismissed):

mds1 (approved):

PR #485, opened 2025-01-27
[FPS] multisig task tests

prateek105:
**Description**
Add tests for Single and Nested Multisig Tasks.

**Tests**

**Additional context**
Using `GasConfigTemplate` for Single multisig tests and `DisputeGameUpgradeTemplate` for nested multisig tests

**Metadata**


blmalone (commented):
I don't think we need `test/registry/.gitkeep` in this PR. 

Made a first pass through this. Still need to go through `SingleMultisig.t.sol`.

test/task/mock/IncorrectGasConfigTemplate1.sol
```diff
@@ -0,0 +1,70 @@
+pragma solidity 0.8.15;
+
+import {SystemConfig} from "@eth-optimism-bedrock/src/L1/SystemConfig.sol";
+
+import {MultisigTask} from "src/fps/task/MultisigTask.sol";
+import {AddressRegistry as Addresses} from "src/fps/AddressRegistry.sol";
+
+/// @title IncorrectGasConfigTemplate1
+/// @notice allowed storage write to incorrect address
+contract IncorrectGasConfigTemplate1 is MultisigTask {
```
blmalone, line 10: I'm thinking that we can cut a lot of code down in these `IncorrectGasConfigTemplate*`files by making the relevant functions inside GasConfigTemplate `vitrual`. e.g. 

```
contract A {
    function foo() internal pure virtual returns (string[] memory);
}

contract B is A {
    function foo() internal pure virtual override returns (string[] memory) {
        // Using virtual twice like this, what do people think?
    }
}

contract C is B {
    function foo() internal pure override returns (string[] memory) {
        // Further override
    }
}
```

`IncorrectGasConfigTemplate2.sol` would become: 

```
pragma solidity 0.8.15;

import {SystemConfig} from "@eth-optimism-bedrock/src/L1/SystemConfig.sol";

import {GasConfigTemplate} from "src/fps/example/template/GasConfigTemplate.sol";
import {AddressRegistry as Addresses} from "src/fps/AddressRegistry.sol";

/// @title IncorrectGasConfigTemplate2
/// @notice not all allowed storages writes are written to
contract IncorrectGasConfigTemplate2 is GasConfigTemplate {

    /// @notice Modifies the storage write permissions for the task
    /// @return Array of storage write permissions
    function _taskStorageWrites() internal pure override returns (string[] memory) {
        string[] memory storageWrites = new string[](2);
        storageWrites[0] = "SystemConfigProxy"; // Original permission
        storageWrites[1] = "CustomConfigProxy"; // Additional permission
        return storageWrites;
    }
}
```

blmalone (approved):

Issue #475, opened 2025-01-17
Rename Proposal to Task


blmalone:
See more information here: https://github.com/ethereum-optimism/superchain-ops/pull/467#discussion_r1919205208

PR #486, opened 2025-01-28
Fix/naming

ElliotFriedman:
Change all naming from Proposal to Task.

**Tests**

This is only a naming convention change, no tests needed.

**Metadata**

Should close https://github.com/ethereum-optimism/superchain-ops/issues/475


ControlCplusControlV (commented):
Why is this necessary?

blmalone (approved):

Issue #343, opened 2024-10-09
Run monorepo test suite against state resulting from proposal


mds1:
This is discussed in https://github.com/ethereum-optimism/superchain-ops/issues/140#issuecomment-2075435988 through the use of [forge-proposal-simulator](https://github.com/solidity-labs-io/forge-proposal-simulator) (FPS). How to best do this needs some thought, and we should make sure our approach is forward compatible with batched upgrades (i.e. currently superchain-ops is focused on upgrading one task at a time but soon we will want to upgrade L1 contracts of many chains in 1 transaction—in that case we want to run the monorepo tests for each chain).

This is a big task, and likely will require changes both to this repo and the monorepo. We should check out how FPS does this, consider using that to avoid rolling our own solution, and may @ElliotFriedman will be kind enough to lend some time to help us out here :)

mds1:
Yes definitely! So I recall last time we chatted about this a tricky part was that the upgrade transactions are executed in this repo, but our test suite is in the monorepo (https://github.com/ethereum-optimism/optimism/tree/develop/packages/contracts-bedrock). However this repo does already have the monorepo as a [dependency](https://github.com/ethereum-optimism/superchain-ops/blob/808004cd933e0bfa36e6d9124701e296639a6d41/.gitmodules#L7-L10), so that might help

First, some background on the monorepo setup:
- The core test suite is a standard forge test suite. We also have a few kontrol proofs, and a few differential ffi tests to ensure go and solidity implementations match.
- We run all tests (including the go ffi tests) with `just test`, and the proofs are run with `just test-kontrol`.
- Only running `just test` would be be a great start and kontrol can be added later.

Now on this repo, superchain-ops
- "Tasks"are transactions to broadcast, like contract upgrade transactions or transactions that just change a config value. This is essentially the same as what Maker calls "spells"
- All tasks live in a directory of `tasks/{L1-network}/{taskName}`, and contain `input.json` bundles that define the transaction to be sent, along with a single solidity script that simulates the task based on that JSON bundle. 
- Some tasks are "single" meaning it's a regular safe that needs to send the transaction, like the 5/7 Foundation Safe. Other tasks are "nested" meaning it's a 2/2 Safe that needs to send the transaction (where each signer on that 2/2 is a Safe, specifically the Foundation and the Security Council)
[...]

PR #420, opened 2024-12-17
Simulate non-terminal tasks in CI - @bordolot contribution

blmalone:
We have a draft PR that was a good first shot at adding the simulation of non-terminal tasks to CI [here](https://github.com/ethereum-optimism/superchain-ops/pull/391). 

This PR builds upon that branch, preserving the open source contributors (@bordolot) commits. 

Closes: https://github.com/ethereum-optimism/superchain-ops/issues/337

PR #505, opened 2025-01-30
[PT: 2] Monorepo Integration Testing

ElliotFriedman:
**Run Monorepo Integration Tests**

This PR adds commands to run monorepo integration tests against state resulting from running all non-terminal tasks.

**Tests**

This PR runs monorepo upgrade integration tests against the resulting state after running all non-terminal tasks.

**Metadata**

this should close issue https://github.com/ethereum-optimism/superchain-ops/issues/343


src/fps/example/Runner.sol
```diff
@@ -0,0 +1,42 @@
+pragma solidity 0.8.15;
```
blmalone, line 1: Now that we have our new `test/` directory, should these files be moved under there?

src/fps/example/runnerConfig.toml
```diff
@@ -0,0 +1,31 @@
+[status]
```
blmalone, line 1: Can we put a comment in this file to explain what it's doing? 

Generally speaking it would be great to get this flow added to the README.md file so it's explained somewhere. Folks will need to be prompted to know to edit this file when they want the monorepo integration tests ran against the state of their code. This is an opt in action.

src/fps/example/Runner.sol
```diff
@@ -0,0 +1,42 @@
...
+        string name;
+        string path;
+        uint256 status;
+    }
+
+    function run() public {
+        string memory runnerConfigFileContents = vm.readFile("src/fps/example/runnerConfig.toml");
+        TasksStatus[] memory tasksStatuses =
+            abi.decode(vm.parseToml(runnerConfigFileContents, ".tasks"), (TasksStatus[]));
+        if (block.chainid == 1) {
```
blmalone, line 20: I think you can do this: 
```suggestion
        if (block.chainid == getChain("Mainnet").chainId) {
```

src/fps/example/Runner.sol
```diff
@@ -0,0 +1,42 @@
...
+            network = "mainnet";
+        } else if (block.chainid == 11155111) {
```
blmalone, line 22: Similarly: 
```suggestion
        } else if (block.chainid == getChain("Sepolia").chainId) {
```

src/fps/example/runnerConfig.toml
```diff
@@ -0,0 +1,31 @@
...
+draft = 0
+ready = 1
+contingency = 2
+signed = 3
+executed = 4
+cancelled = 5
+
+[[tasks]]
+name = "task-00"
+status = 1
```
blmalone, line 11: This feels like an additional flag that we have to manage (change every time we execute etc). Is it possible to figure out the tasks state by looking at the README of a given task? 

What we basically want here is "all non-terminal" tasks. I implemented something to this effect against the old setup, see [here](https://github.com/ethereum-optimism/superchain-ops/pull/420/files#diff-fb15a51c0ce99dc85c2344094d827b878a45fa5140a175a9f7e10ca3a8a1ea3dR17). If we could automatically determine the tasks to run tests against that would be a much better UX.

blmalone:
**Runner.sol**

Assume our directory structure is `tasks/<network>/<task_number_name>` e.g. `tasks/eth/001-gas-update-base`.

Enumerate all tasks under `tasks/` for each network, finding non-terminal tasks. Do this by looking for the keywords in README of these directories. At the end of this execution we should have a mapping of `network -> task directory` where every `task directory` is non-terminal.

**Means**
Keep this directory layout:
```
config.toml -> this should have a new property templateName = "GasConfigTemplate"
README.md
VALIDATIONS.md
```

src/fps/example/runnerConfig.toml
blmalone: Let's remove runnerConfig.toml as discussed offline [related](https://github.com/ethereum-optimism/superchain-ops/pull/505#discussion_r1937947336).

blmalone (approved):
Looks good. I think it's worth mentioning that when we have actual tasks in `tasks/$network` that the `task/mock/example` directories will not be run against the monorepo tests.

PR #503, opened 2025-01-29
setup OPSC Rehearsal 4 

futurealisha:
**Description**

Setup Rehearsal 4 for incoming signers of the OP Security Council

PR #509, opened 2025-01-30
fix: update monorepo submodule

blmalone:
Need to update the monorepo commit to get the changes made in this PR: https://github.com/ethereum-optimism/optimism/pull/14069


mds1:
The Security Council is also doing some rehearsals, which deploy code from the current monorepo submodule version, in https://github.com/ethereum-optimism/superchain-ops/pull/503. I will review that PR today, and make sure this one does not impact it (I don't think it should, but just want to confirm before merge)

mds1 (approved):

PR #527, opened 2025-01-31
Remove Constants File

ElliotFriedman:
**Description**

This PR removes the constants file and all associated dependencies. It instead relies on the updates to supported chains in StdChains. The `lib/forge-std` submodule has been updated to use Metal and Orderly which were recently included.


prateek105 (approved):
lgtm

blmalone (approved):

PR #528, opened 2025-02-01
perf: reduce build time

### LINKED ISSUE #512: Speed up slow superchain-ops CI
CI tends to be slow for superchain-ops. This could be due to many reasons but some low hanging fruit may be to swap public RPC URLs with private ones.

mds1:
Small code changes to allow compiling without via-ir, reduces build time from over 5 minutes to <2s. Via-ir is very slow when doing a lot of struct operations

Also removes extra slashes that were appended to a lot of comments

Partially resolves #512

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -213,14 +213,14 @@ jobs:
             just simulate-council
             just prepare-json
             just simulate-council # simulate again to make sure the json is still valid
-  
+
   just_simulate_ink_respected_game_type:
     docker:
       - image: << pipeline.parameters.ci_builder_image >>
     steps:
       - simulate:
           task: "/eth/ink-002-set-respected-game-type"
-  
+
   just_simulate_eth_029_holocene_system_config_upgrade_and_init_base:
     docker:
       - image: << pipeline.parameters.ci_builder_image >>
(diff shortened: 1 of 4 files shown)


test/task/NestedMultisigTask.t.sol
```diff
@@ -60,17 +60,18 @@ contract NestedMultisigTaskTest is Test {
...
-            /// dataToSign is the data that the EOA owners of the child multisig has to sign to help
-            /// execute the child multisig approval of hashToApproveByChildMultisig
+            // dataToSign is the data that the EOA owners of the child multisig has to sign to help
+            // execute the child multisig approval of hashToApproveByChildMultisig
             bytes memory dataToSign = getNestedDataToSign(childOwnerMultisigs[i]);
-            /// nonce is not decremented by 1 because in task simulation approveHash is called by
-            /// the child multisig which does not increment the nonce
+            // nonce is not decremented by 1 because in task simulation approveHash is called by
+            // the child multisig which does not increment the nonce
+            uint256 nonce = IGnosisSafe(childOwnerMultisigs[i]).nonce();
```
mds1, line 74: @ElliotFriedman @blmalone @prateek105 I pulled out nonce into a variable here to mitigate a stack too deep error. Please confirm this is safe to reuse in this way, i.e. I assume the nonce is expected to be the same for both the `encodeTransactionData` and `getTransactionHash` calls?

.circleci/config.yml
```diff
@@ -298,7 +298,11 @@ workflows:
       # expect the ceremony to work continuously), and remove it once
       # the ceremony is for historical archive only (e.g. the
       # ceremony is done).
-      - just_simulate_sc_rehearsal_1
+      #
+      # We skip rehearsal 1 because we already have completed this rehearsal (https://github.com/ethereum-optimism/superchain-ops/pull/459),
+      # and now it fails with stack too deep after https://github.com/ethereum-optimism/superchain-ops/pull/528.
+      # We wll need to rewrite the rehearsals with the new superchain-ops structure anyway, so this is ok.
+      # - just_simulate_sc_rehearsal_1
```
mds1, line 305: Tracking in https://github.com/ethereum-optimism/superchain-ops/issues/529 and added that as a non-blocker to https://github.com/ethereum-optimism/superchain-ops/issues/396

blmalone (approved):
Looks good!

PR #544, opened 2025-02-05
op-sepolia: Add task to switch back to singlethreaded cannon.

ajsutton:
**Description**

Add task to switch back to singlethreaded cannon.

Updates the set-game-implementation task to handle recent breaking changes in superchain-registry and the monorepo.


tasks/sep/030-revert-mt-cannon/NestedSignFromJson.s.sol
```diff
@@ -0,0 +1,184 @@
...
+import {Vm, VmSafe} from "forge-std/Vm.sol";
+import {GnosisSafe} from "safe-contracts/GnosisSafe.sol";
+import {LibString} from "solady/utils/LibString.sol";
+import "@eth-optimism-bedrock/src/dispute/lib/Types.sol";
+import {DisputeGameFactory} from "@eth-optimism-bedrock/src/dispute/DisputeGameFactory.sol";
+import {FaultDisputeGame} from "@eth-optimism-bedrock/src/dispute/FaultDisputeGame.sol";
+import {PermissionedDisputeGame} from "@eth-optimism-bedrock/src/dispute/PermissionedDisputeGame.sol";
+import {SystemConfig} from "@eth-optimism-bedrock/src/L1/SystemConfig.sol";
+
+contract NestedSignFromJson is OriginalNestedSignFromJson {
```
sebastianst, line 18: Could we use the verification contracts instead, like for many Holocene upgrades?

Ethnical: Agreed with @sebastianst. 
This is easier for us to review and also increase the consistency for the signers + reviewers. 
I would suggest to use this as example: 
https://github.com/ethereum-optimism/superchain-ops/tree/main/tasks/sep/027-holocene-system-config-upgrade-and-init-multi-chain

mbaxter: [This one](https://github.com/ethereum-optimism/superchain-ops/blob/dfdb5698166efb334c8146562397b28c5ac53f42/tasks/sep/026-fp-holocene-upgrade-fix/NestedSignFromJson.s.sol#L13-L13) uses the new `DisputeGameUpgrade` verification contract.

ajsutton: Discussed more in discord - this was generated via the standard template for setting dispute games.  The verification framework assumes there is permissionless and permissioned game types which isn't always the case so doesn't fit well with the template which can handle all chains.  But for this case I've updated the script to use the verification framework.

tasks/sep/030-revert-mt-cannon/.env
```diff
@@ -0,0 +1,9 @@
+ETH_RPC_URL="https://ethereum-sepolia.publicnode.com"
+COUNCIL_SAFE=0xf64bc17485f0B4Ea5F06A96514182FC4cB561977
+FOUNDATION_SAFE=0xDEe57160aAfCF04c34C887B5962D0a69676d3C8B
+SAFE_NONCE=""
+L1_CHAIN_NAME="sepolia"
+L2_CHAIN_NAME="op"
+SYSTEM_CONFIG="0x034edD2A225f7f429A63E0f1D2084B9E0A93b538"
```
mbaxter, line 7: Do we need `SYSTEM_CONFIG` and `PROXY_ADMIN`?  I don't see them used.

tasks/sep/030-revert-mt-cannon/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+This task updates the fault dispute system for op-sepolia: 
+
+* Set implementation for game type 0 to 0xF3CcF0C4b51D42cFe6073F0278c19A8D1900e856 in `DisputeGameFactory` 0x05F9613aDB30026FFd634f38e5C4dFd30a197Fa1: `setImplementation(0, 0xF3CcF0C4b51D42cFe6073F0278c19A8D1900e856)`
+* Set implementation for game type 1 to 0xbbDBdfe37C02439764dE0e41C906e4396B5B3914 in `DisputeGameFactory` 0x05F9613aDB30026FFd634f38e5C4dFd30a197Fa1: `setImplementation(1, 0xbbDBdfe37C02439764dE0e41C906e4396B5B3914)`
+<!--NEXT TASK DESCRIPTION-->
+
+This switches sepolia back to singlethreaded cannon, undoing the upgrade to cannon-mt (025-mt-cannon task)
+
+### State Validations
```
mbaxter, line 15: This README is missing a few sections compared to some other [examples](https://github.com/ethereum-optimism/superchain-ops/blob/d04c75e4793251296310fb29481b80302449fa0e/tasks/sep/026-fp-holocene-upgrade-fix/README.md#L1-L42) like: Pre-deployments, Simulation, Execution.

tasks/sep/030-revert-mt-cannon/VALIDATION.md
```diff
@@ -0,0 +1,154 @@
...
+- All key values match the semantic meaning provided, which can be validated using the storage layout links provided.
+
+## Nested Safe State Overrides and Changes
+
+This task is executed by the nested 2/2 `ProxyAdminOwner` Safe. Refer to the
+[generic nested Safe execution validation document](../../../NESTED-VALIDATION.md)
+for the expected state overrides and changes.
+
+The `approvedHashes` mapping **key** of the `ProxyAdminOwner` that should change during the simulation is
+- Council simulation: `0x7a680028073e956784e910c02dd0a0936604b29cd9f9c71c9d8e568533821e16`
```
mbaxter, line 21: When I simulated for the council, I got: `0xe8f8aa2e65c85624906005cf8077cbc632900bd87fdfb6023df637b4fddfffe3 `

tasks/sep/030-revert-mt-cannon/VALIDATION.md
```diff
@@ -0,0 +1,154 @@
...
+- Foundation simulation: `0xd30a77d4a810ba7768ba1bd52de1cbb869f7b641c57132e46544f044cd7e839a`
```
mbaxter, line 22: I got: `0x0d550ba923712ad13c224c6566f757fbc09207dd05fec85aebd7a66e56e27bd5`

tasks/sep/030-revert-mt-cannon/VALIDATION.md
```diff
@@ -0,0 +1,154 @@
...
+  **Before**: `0x000000000000000000000000bbd576128f71186a0f9ae2f2aab4afb4af2dae17` <br/>
+  **After**: `0x000000000000000000000000bbDBdfe37C02439764dE0e41C906e4396B5B3914` <br/>
+  **Meaning**: Updates the implementation for game type 1. Verify that the new implementation is set using
+  `cast call 0x05F9613aDB30026FFd634f38e5C4dFd30a197Fa1 "gameImpls(uint32)(address)" 1`.
+
+### `0x1Eb2fFc903729a0F03966B917003800b145F56E2` (`ProxyAdminOwner`)
+
+* **Key**: `0x0000000000000000000000000000000000000000000000000000000000000005`
+* **Before**: `0x0000000000000000000000000000000000000000000000000000000000000017`
+* **After**: `0x0000000000000000000000000000000000000000000000000000000000000018`
```
mbaxter, line 61:
```suggestion
* **Before**: `0x0000000000000000000000000000000000000000000000000000000000000011`
* **After**: `0x0000000000000000000000000000000000000000000000000000000000000012`
```
I think you set these in decimal instead of hex? (17 = 0x11)

tasks/sep/fp-recovery/005-set-game-implementation/justfile
```diff
@@ -36,7 +36,7 @@ prep l1='' l2='':
   CONFIG_URL="https://raw.githubusercontent.com/ethereum-optimism/superchain-registry/refs/heads/main/superchain/configs/{{l1}}/{{l2}}.toml"
   CONFIG_TOML=$(curl -s --show-error --fail "${CONFIG_URL}")
   SYSTEM_CONFIG=$(echo "${CONFIG_TOML}" | yq -p toml .addresses.SystemConfigProxy)
-  OWNER_SAFE=$(echo "${CONFIG_TOML}" | yq -p toml .addresses.ProxyAdminOwner)
+  OWNER_SAFE=$(echo "${CONFIG_TOML}" | yq -p toml .roles.ProxyAdminOwner)
```
mbaxter, line 39: I don't see this change in the [superchain-registry submodule](https://github.com/ethereum-optimism/superchain-registry/blob/c08331ab44a3645608c08d8c94f78d9be46c13c9/superchain/configs/sepolia/op.toml#L40)

mbaxter (commented):
Thanks for putting this together!  Left a few comments on some discrepancies I noticed.

tasks/sep/fp-recovery/005-set-game-implementation/justfile
ajsutton: It uses the file via a http request to the superchain-registry, not the submodule to ensure it is always using the latest version. That was part of the feedback on the initial template review.

tasks/sep/030-revert-mt-cannon/README.md
ajsutton: The info on predeployed contracts in included in the validation page.  Not sure we need to repeat instruction on simulating in every task but added.  The details of the execution are already listed in the Objective.

tasks/sep/030-revert-mt-cannon/VALIDATION.md
ajsutton: image

It's currently going from 23 to 24 but what the actual diff is will depend on when it's executed.

ajsutton: Oh this varies depending on council or foundation.  Going to just remove it and say the nonce should update.

mbaxter (approved):

tasks/sep/030-revert-mt-cannon/NestedSignFromJson.s.sol
```diff
@@ -0,0 +1,57 @@
...
+import {NestedSignFromJson as OriginalNestedSignFromJson} from "script/NestedSignFromJson.s.sol";
+import {DisputeGameUpgrade} from "script/verification/DisputeGameUpgrade.s.sol";
+import {CouncilFoundationNestedSign} from "script/verification/CouncilFoundationNestedSign.s.sol";
+import {SuperchainRegistry} from "script/verification/Verification.s.sol";
+
+contract NestedSignFromJson is OriginalNestedSignFromJson, CouncilFoundationNestedSign, DisputeGameUpgrade {
+    constructor()
+    SuperchainRegistry("sepolia", "op", "v1.8.0-rc.4")
+    DisputeGameUpgrade(
+    0x03f89406817db1ed7fd8b31e13300444652cdb0b9c509a674de43483b2f83568, // absolutePrestate
```
Ethnical, line 17: How can precompute this prestate?

tasks/sep/030-revert-mt-cannon/README.md
```diff
@@ -0,0 +1,23 @@
+# ProxyAdminOwner - Set Dispute Game Implementation
```
Ethnical, line 1: I think we should add a link to the version of the contracts used (seems to be the `v1.8.0-rc.4` solidity file `NestedSignFromJson.s.sol`.

tasks/sep/030-revert-mt-cannon/VALIDATION.md
```diff
@@ -0,0 +1,130 @@
...
+
+Mismatch version()(string)
+Was: "1.0.0-beta.7"
+Now: "1.2.1"
+```
+
+In both, there are two changes:
+
+* absolutePrestate() changes from `0x03b7eaa4e3cbce90381921a4b48008f4769871d64f93d113fcadca08ecee503b` to `0x03f89406817db1ed7fd8b31e13300444652cdb0b9c509a674de43483b2f83568`. 
+  These can be verified by comparing to the values in [releases.json](https://github.com/ethereum-optimism/optimism/blob/develop/op-program/prestates/releases.json).
```
Ethnical, line 128: For improving readibility for signers and going faster for reviewers just added the permalink to version `1.4.0`. 
```suggestion
  These can be verified by comparing to the values in [releases.json](https://github.com/ethereum-optimism/optimism/blob/develop/op-program/prestates/releases.json).
  These can be verified by comparing to the values in [releases.json](https://github.com/ethereum-optimism/optimism/blob/be6ca32937c8cd47d08c560dde55bc68142c48cc/op-program/prestates/releases.json#L44).
```

ajsutton: This should be reviewed against the latest version of releases.json - it contains the full history of releases. If for some reason something changed (it shouldn't but if it did) you'd want to review against the latest version and understand if that change affects this update.

tasks/sep/030-revert-mt-cannon/README.md
ajsutton: The contracts used actually come from multiple different releases because our release process doesn't release all contracts together.  The reference to v1.8.0-rc.4 in the superchain registry is a bit misleading but that's how the registry works - it actually defines the expected versions for all contracts at that snapshot even if the contract version is actually from an older release.  I don't think we want to try and explain that level of complexity here - the important thing is actually what VALIDATION.md already covers which is identifying what is actually being changed between the new and old implementation contracts (and the dispute game contracts are not changing - just the MIPS version).

Ethnical:
The simulation match the verification files as expected for both SC an FND. 
1eca6aa37e74ec96c39beadd5ed4c6222ca6c63fab57a5eca3ee682108e06c95

Ethnical (approved):
LGTM!
I would like to know how we can ensure that the deployment of the both DG is correct: 
- [0xF3CcF0C4b51D42cFe6073F0278c19A8D1900e856](https://sepolia.etherscan.io/address/0xF3CcF0C4b51D42cFe6073F0278c19A8D1900e856#code)
- [0xbbDBdfe37C02439764dE0e41C906e4396B5B3914](https://sepolia.etherscan.io/address/0xbbdbdfe37c02439764de0e41c906e4396b5b3914)
- 
CleanShot 2025-02-10 at 12 24 58@2x

I have opened a slack discussion + discord and we should merge this when we resolve this off to make sure this is desired. 

For the CI errors this is related to rate limiting, I have opened an issue here https://github.com/ethereum-optimism/superchain-ops/issues/569 to fix it later.

Issue #557, opened 2025-02-05
Monorepo tests run on a schedule AND merge to main


blmalone:
Following on from this conversation in discord: https://discord.com/channels/1244729134312198194/1327372079380824267/1336824792329883679 

The `monorepo-integration-test` job in ci needs to be executed on a schedule to make sure to catch any errors in simulating tasks that are going to be executed. 

Here's an [example of it in practice](https://github.com/ethereum-optimism/security-tools/pull/190/files).

blmalone:
Closed by: https://github.com/ethereum-optimism/superchain-ops/pull/558

PR #558, opened 2025-02-05
Run CI on main every 4 hours.

blmalone:
Closes: https://github.com/ethereum-optimism/superchain-ops/issues/557


.circleci/config.yml
```diff
@@ -351,6 +351,15 @@ jobs:
...
             (cd src/improvements/script && shellcheck -x *)
 
+scheduled:
+  triggers:
+    - schedule: # Running CI on main every 4 hours
+        cron: "0 */4 * * *"
+        filters:
+          branches:
+            only:
+              - main
```
mds1, line 361: Does this work, where does this get applied to the main workflow? In the monorepo we use the e.g. `build_four_hours` trigger:
- config: https://app.circleci.com/settings/project/github/ethereum-optimism/optimism/triggers?return-to=https://app.circleci.com/pipelines/github/ethereum-optimism/optimism
- applying it: https://github.com/ethereum-optimism/optimism/blob/420be9caa771f7136e65b1bbe47c383892b7b595/.circleci/config.yml#L1791

blmalone: Interesting I haven't used triggers like that in circleci before. Though I have used triggers this way via the yml file itself (security-tools).

mds1: I still don't understand how this gets applied to the main workflow?

blmalone: I've changed the approach to match what we do in the monorepo.

blmalone: https://github.com/ethereum-optimism/superchain-ops/pull/558/commits/2e68695e36a06c70297aaf194228d6b433e13979 and see [here](https://app.circleci.com/settings/project/gh/ethereum-optimism/superchain-ops/triggers?return-to=https%3A%2F%2Fapp.circleci.com%2Fpipelines%2Fgh%2Fethereum-optimism%2Fsuperchain-ops).

mds1 (approved):

.circleci/config.yml
```diff
@@ -353,6 +353,11 @@ jobs:
 
 workflows:
   main:
+    when:
+      or:
+        - equal: [build_four_hours, << pipeline.trigger_source >>]
+        - not:
+            equal: [scheduled_pipeline, << pipeline.trigger_source >>]
```
blmalone, line 360: The `when` condition uses an `or` statement to run the workflow in two scenarios:
1. When the trigger source is 'build_four_hours'
1. When the trigger source is not a scheduled pipeline (which covers all branch pushes and pull requests)

mds1 (approved):

PR #559, opened 2025-02-06
fix: code comment errors

intls:
**Description**

**Tests**

**Additional context**

**Metadata**


mds1:
Thanks, we do not accept PRs that only fix spelling and grammar in the monorepo, and will add a `CONTRIBUTING.md` file to reflect that we follow the same policy here

PR #561, opened 2025-02-06
Scheduled workflow added for clarity.

blmalone:
With the current change, the workflow doesn't run as expected (see screenshot). This change creates a separate workflow that only gets run when the trigger source is the `build_four_hours`. This way it's easier to reason about the workflows in isolation. 

CleanShot 2025-02-05 at 23 50 23@2x

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -352,12 +352,16 @@ jobs:
             (cd src/improvements/script && shellcheck -x *)
 
 workflows:
-  main:
+  monorepo_integration_test_scheduled:
     when:
-      or:
         - equal: [build_four_hours, << pipeline.trigger_source >>]
-        - not:
-            equal: [scheduled_pipeline, << pipeline.trigger_source >>]
+    jobs:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -352,12 +352,16 @@ jobs:
             (cd src/improvements/script && shellcheck -x *)
 
 workflows:
-  main:
+  monorepo_integration_test_scheduled:
```
mds1, line 355: I would suggest running the full main CI every four hours, not just these tests. The rationale being that if something onchain changes and it impacts an ongoing task we can find out

.circleci/config.yml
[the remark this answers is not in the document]
blmalone: Yes thats'a flake.

.circleci/config.yml
blmalone: Okay, I consolidated the workflows so we have one single workflow again.

mds1 (approved):

PR #568, opened 2025-02-09
FIX: replace -j flag with --json in sep/fp-recovery/005/justfile

Wazabie:
**Description**

The -j flag in the sep/fp-recovery/005-set-game-implementation/justfile causes a parsing error (possibly deprecated, misinterpreted or conflicting with another flag). Changing it to --json resolves the issue and correctly retrieves the expected JSON output.

Screenshot 2025-02-09 at 16 56 13


ajsutton (approved):
Ah yes, cast breaks CLI compatibility in some versions.

PR #565, opened 2025-02-07
Fix/reduce registry trust

blmalone:
Building on @mds1's work. I've added address discovery via the L1StandardBridge contract. In the case of the `L1ERC721Bridge` contract, this is not discoverable from the L1StandardBridge and therefore this contract address is taken from the superchain-registry `addresses.json` file. 
Not all chains are on the same version of L1 contracts. For this reason, this PR includes some conditional logic that decides whether to try to retrieve an address or not. For example, all of the dispute game contracts are only discoverable on some chains. 

I was hitting some `StackTooDeep` errors so you'll notice that this PR refactors some logic into functions. I tried to keep these functions as logically named as possible. 

I didn't include the ProtocolVersions contract as this is _not_ in `addresses.json` _and_  not discoverable onchain.

I included 5 chains in the discovery test. The test (`testSuperchainAddressesLoaded`) takes a total of 34 seconds. Which means it's a average of 6.8 seconds per chain discovery. 
```
l2chains = [{name = "Orderly", chainId = 291}, {name = "Metal", chainId = 1750}, {name = "Base", chainId = 8453}, {name = "OP Mainnet", chainId = 10}, {name = "Zora", chainId = 7777777}]
```

A side note, I think that the `tests/task/mock/example/` directory should be removed at some point. The test data isn't descriptive and can lead to confusion.

PR #582, opened 2025-02-11
Discovery bug: try/catch not working for addresses with no code.

blmalone:
Related to: https://github.com/ethereum-optimism/superchain-ops/pull/565

Opting to take the approach of using low-level calls and checking return data to make sure that a contract call occurred as expected.

### DIFF
--- a/src/improvements/AddressRegistry.sol
+++ b/src/improvements/AddressRegistry.sol
@@ -412,10 +412,6 @@ contract AddressRegistry is IAddressRegistry, Test {
         internal
         returns (address)
     {
-        (bool ok, bytes memory data) =
-            address(faultDisputeGame).call(abi.encodeWithSelector(IFetcher.anchorStateRegistry.selector));
-        if (ok && data.length == 32) return abi.decode(data, (address));
-
         return IFetcher(permissionedDisputeGame).anchorStateRegistry();
     }
 
--- a/src/improvements/AddressRegistry.sol
+++ b/src/improvements/AddressRegistry.sol
@@ -410,34 +410,30 @@ contract AddressRegistry is IAddressRegistry, Test {
 
     function getAnchorStateRegistryProxy(address faultDisputeGame, address permissionedDisputeGame)
         internal
-        view
         returns (address)
     {
-        try IFetcher(faultDisputeGame).anchorStateRegistry() returns (address anchorStateRegistryProxy) {
-            return anchorStateRegistryProxy;
-        } catch {
-            return IFetcher(permissionedDisputeGame).anchorStateRegistry();
-        }
+        (bool ok, bytes memory data) =
+            address(faultDisputeGame).call(abi.encodeWithSelector(IFetcher.anchorStateRegistry.selector));
+        if (ok && data.length == 32) return abi.decode(data, (address));
+
+        return IFetcher(permissionedDisputeGame).anchorStateRegistry();
     }
 
     function getDelayedWETHProxy(address faultDisputeGame, address permissionedDisputeGame)
         internal
-        view
         returns (address)
     {
-        try IFetcher(faultDisputeGame).weth() returns (address delayedWethProxy) {
-            return delayedWethProxy;
-        } catch {
-            return IFetcher(permissionedDisputeGame).weth();
-        }
+        (bool ok, bytes memory data) = address(faultDisputeGame).call(abi.encodeWithSelector(IFetcher.weth.selector));
+        if (ok && data.length == 32) return abi.decode(data, (address));
+
+        return IFetcher(permissionedDisputeGame).weth();
     }
 
-    function getMips(address faultDisputeGame, address permissionedDisputeGame) internal view returns (address) {
-        try IFetcher(faultDisputeGame).vm() returns (address mips) {
-            return mips;
-        } catch {
-            return IFetcher(permissionedDisputeGame).vm();
-        }
+    function getMips(address faultDisputeGame, address permissionedDisputeGame) internal returns (address) {
+        (bool ok, bytes memory data) = address(faultDisputeGame).call(abi.encodeWithSelector(IFetcher.vm.selector));
+        if (ok && data.length == 32) return abi.decode(data, (address));
+
+        return IFetcher(permissionedDisputeGame).vm();
     }
 
     function getBatchSubmitter(address systemConfigProxy) internal view returns (address) {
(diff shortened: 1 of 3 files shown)


src/improvements/AddressRegistry.sol
```diff
@@ -410,34 +410,30 @@ contract AddressRegistry is IAddressRegistry, Test {
...
-        try IFetcher(faultDisputeGame).anchorStateRegistry() returns (address anchorStateRegistryProxy) {
-            return anchorStateRegistryProxy;
-        } catch {
-            return IFetcher(permissionedDisputeGame).anchorStateRegistry();
-        }
+        (bool ok, bytes memory data) =
+            address(faultDisputeGame).call(abi.encodeWithSelector(IFetcher.anchorStateRegistry.selector));
+        if (ok && data.length == 32) return abi.decode(data, (address));
+
+        return IFetcher(permissionedDisputeGame).anchorStateRegistry();
```
mds1, line 419: There's an invariant that ALL chains on any fault proof version will have a permissioned dispute game. And the fault dispute game is only present for chains on permissionless proofs
```suggestion
        return IFetcher(permissionedDisputeGame).anchorStateRegistry();
```

src/improvements/AddressRegistry.sol
```diff
@@ -410,34 +410,30 @@ contract AddressRegistry is IAddressRegistry, Test {
...
         internal
-        view
         returns (address)
     {
-        try IFetcher(faultDisputeGame).weth() returns (address delayedWethProxy) {
-            return delayedWethProxy;
-        } catch {
-            return IFetcher(permissionedDisputeGame).weth();
-        }
+        (bool ok, bytes memory data) = address(faultDisputeGame).call(abi.encodeWithSelector(IFetcher.weth.selector));
```
mds1, line 426: We can change these all to staticcalls and keep the `view` modifier
```suggestion
        (bool ok, bytes memory data) = address(faultDisputeGame).staticcall(abi.encodeWithSelector(IFetcher.weth.selector));
```

blmalone: https://github.com/ethereum-optimism/superchain-ops/pull/582/commits/9fea16d01342df5fb6b77b9218b62376ebd61b2c

src/improvements/AddressRegistry.sol
```diff
@@ -413,31 +413,27 @@ contract AddressRegistry is IAddressRegistry, Test {
...
-        try IFetcher(faultDisputeGame).weth() returns (address delayedWethProxy) {
-            return delayedWethProxy;
-        } catch {
-            return IFetcher(permissionedDisputeGame).weth();
-        }
+        (bool ok, bytes memory data) =
+            address(faultDisputeGame).staticcall(abi.encodeWithSelector(IFetcher.weth.selector));
+        if (ok && data.length == 32) return abi.decode(data, (address));
+
+        return IFetcher(permissionedDisputeGame).weth();
```
mds1, line 428: This code block is inaccurate since there is one delayed WETH proxy per game type. So the FDG has it's own delayed weth, and the PDG has its own delayed weth. I would change this method to take a single dispute game input, then call it once or twice (i.e. once for the PDG, once for the FDG if it's present)

src/improvements/AddressRegistry.sol
```diff
@@ -413,31 +413,27 @@ contract AddressRegistry is IAddressRegistry, Test {
...
-            return mips;
-        } catch {
-            return IFetcher(permissionedDisputeGame).vm();
-        }
+        (bool ok, bytes memory data) =
+            address(faultDisputeGame).staticcall(abi.encodeWithSelector(IFetcher.vm.selector));
+        if (ok && data.length == 32) return abi.decode(data, (address));
+
+        return IFetcher(permissionedDisputeGame).vm();
     }
```
mds1, line 437: Similar to the ASR method, no reason to check the FDG and take that as an input here

src/improvements/AddressRegistry.sol
blmalone: Does this mean there should be two 'DelayedWETHProxy' addresses saved to the AddressRegistry is a chain has both FDG and PDG? Or should we only save one? If FDG is present then save it's DelayedWETHProxy otherwise default to saving the PDG DelayedWETHProxy.

src/improvements/AddressRegistry.sol
blmalone: https://github.com/ethereum-optimism/superchain-ops/pull/582/commits/d706fec9ae0caf08656b793d2b1d3d96b297ed8b

src/improvements/AddressRegistry.sol
mds1: > there should be two 'DelayedWETHProxy' addresses saved to the AddressRegistry is a chain has both FDG and PDG

Exactly. Actually, I would probably just stop the lookup at the dispute game addresses, save those, and a task can fetch the delayed WETH addresses if needed. No need to store the WETH addresses. This will make it more maintainable, because eventually we will add a third game type (asterisc)

blmalone: Would another approach be to look at the current game type on the OptimismPortalProxy via `resepectedGameType()`? https://etherscan.io/address/0xbEb5Fc579115071764c7423A4f12eDde41f106Ed#readProxyContract

Once you have that, you can decide which delayed weth proxy to read (GameType = 0, read FDG, GameType = 1, read PDG).

mds1: Not in the generic case. For example if you are changing the implementation of the delayedWETH contract, you need to check the new impls for both DelayedWETHs

mds1: We don't know what a task will do, so we can't make assumptions about which delayedWETH it will want

blmalone: https://github.com/ethereum-optimism/superchain-ops/pull/582/commits/a695a7bb747b372f465e763892df68edc7987569

src/improvements/AddressRegistry.sol
```diff
@@ -408,36 +407,18 @@ contract AddressRegistry is IAddressRegistry, Test {
...
-    {
-        try IFetcher(faultDisputeGame).weth() returns (address delayedWethProxy) {
-            return delayedWethProxy;
-        } catch {
-            return IFetcher(permissionedDisputeGame).weth();
-        }
+    function getDelayedWETHProxy(address disputeGame) internal view returns (address) {
+        (bool ok, bytes memory data) = address(disputeGame).staticcall(abi.encodeWithSelector(IFetcher.weth.selector));
+        if (ok && data.length == 32) return abi.decode(data, (address));
+        else return address(0);
```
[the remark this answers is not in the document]
blmalone: https://github.com/ethereum-optimism/superchain-ops/pull/582/commits/62a516a33a4015aa3173c9f9a7e8f38062da7897

mds1 (approved):

Issue #571, opened 2025-02-10
We must delegate call to OPCM from superchain-ops


blmalone:
This involves using @maurelian's new MultiCall3DelegateCall contract. 

As part of the superchain-ops improvements, we must provide the ability to swap in different MultiCall3 contracts based on whether the task is using OPCM or not.

maurelian:
FWIW the contract itself should be here on Sepolia and Mainnet: 0x95b259eae68ba96edB128eF853fFbDffe47D2Db0

blmalone:
There are other functions on OPCM that we will want to delegate call to in the future. 

Two types of calls for superchain-ops 2.0:
- Loop over each chain and make calls
- Delegate call to our onchain OPCM (opcm loops - technically you can look by having an array of 1)

Maybe get rid of looping in OPCM. This might be more consistent with what we already have.

maurelian:
We don't _have_ to use `multicall3delegatecall` but we needed a way to make a `delegatecall` from the Safe while still leveraging all of the existing infrastructure in the superchain-ops repo infrastructure for signing and reading the input.json tx bundles. 

However we fulfill those reqs is fine with me.

blmalone:
Closed by: https://github.com/ethereum-optimism/superchain-ops/pull/606

PR #599, opened 2025-02-13
Add OPCM upgrade template

prat-gpt:
**Description**
- Extend Multisig task and create opcm base task
- Add opcm upgrade template
- template uses a dummy OPCM contract on sepolia

**Tests**
- Add test to check the complete flow: data signing by child multisig EOAs, hash approval by child multisig, final execution by parent multisig.

**Metadata**
Related issue: https://github.com/ethereum-optimism/superchain-ops/issues/571


src/improvements/tasks/OPCMBaseTask.sol
```diff
@@ -0,0 +1,120 @@
...
+            gasPrice: 0,
+            gasToken: address(0),
+            refundReceiver: address(0),
+            _nonce: _getNonce(safe)
+        });
+    }
+
+    /// @notice execute post-task checks.
+    /// read states that are expected to have changed during the simulate step.
+    /// removes the storage write checks
```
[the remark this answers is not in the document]
prat-gpt: @mds1 mentioned in the standup that opcm will update states of multiple contracts and for OPCM tasks it can be ignored.

src/improvements/tasks/MultisigTask.sol
```diff
@@ -119,6 +119,8 @@ abstract contract MultisigTask is Test, Script, ITask {
     /// @notice The address of the child multisig for this task
     address public childMultisig;
 
+    address public targetMulticall;
```
blmalone, line 122: Like the rest of the variables, can you comment on this.

src/improvements/tasks/MultisigTask.sol
```diff
@@ -233,6 +237,9 @@ abstract contract MultisigTask is Test, Script, ITask {
 
         _setIsNestedSafe();
 
+        vm.label(address(addresses), "Addresses");
+        vm.label(address(this), "MultisigTask");
```
blmalone, line 241: I'm not sure, do you want these labels in?

test/tasks/mock/example/task-04/config.toml
```diff
@@ -0,0 +1,8 @@
+# this is the file used to determine the network configuration
```
blmalone, line 1: Can we move this into `test/tasks/mock/configs`?

src/improvements/tasks/OPCMBaseTask.sol
```diff
@@ -0,0 +1,120 @@
...
+
+        /// generate calldata
+        data = abi.encodeWithSignature("aggregate3((address,bool,bytes)[])", calls);
+    }
+
+    /// @notice get the data to sign by EOA
+    /// @param safe The address of the safe
+    /// @param data The calldata to be executed
+    /// @return The data to sign
+    function getDataToSign(address safe, bytes memory data) public view override returns (bytes memory) {
```
blmalone, line 62: You can actually get rid of this `getDataToSign` function entirely if you create a function in `MultisigTask.sol` like so: 

```solidity
    function getMulticallAddress(address safe) public view virtual returns (address) {
        require(safe != address(0), "safe address cannot be zero");
        return targetMulticall;
    }
```

and override it inside `OPCMBaseTask.sol` like so (deleting `_getMulticallAddress`): 

```solidity
    /// @notice get the multicall address for the given safe
    /// if the safe is the parent multisig, return the delegatecall multicall address
    /// otherwise if the safe is a child multisig, return the regular multicall address
    /// @param safe The address of the safe
    /// @return The address of the multicall
    function getMulticallAddress(address safe) public view override returns (address) {
        return (safe == multisig) ? MULTICALL3_DELEGATECALL_ADDRESS : MULTICALL3_ADDRESS;
    }
```

Granted the `address` variable isn't used in `MultisigTask.sol` (other than in the require statement), but it means less code duplication.

blmalone (commented):
still need to go through the rest of this.

src/improvements/tasks/MultisigTask.sol
```diff
@@ -119,6 +119,8 @@ abstract contract MultisigTask is Test, Script, ITask {
     /// @notice The address of the child multisig for this task
     address public childMultisig;
```
blmalone, line 120: I was wondering, because we are already using the term `childMultisig`, can we rename the state variable in this file to be `parentMultisig`? I think we will get some readability gains in functions like: 

```solidity
    function _getNonce(address safe) internal view returns (uint256) {
        return (safe == multisig) ? nonce : IGnosisSafe(safe).nonce();
    }
```

would become:

```solidity
    function _getNonce(address currentSafe) internal view returns (uint256) {
        return (currentSafe == parentMultisig) ? nonce : IGnosisSafe(safe).nonce();
    }
```

src/improvements/tasks/MultisigTask.sol
```diff
@@ -798,6 +793,18 @@ abstract contract MultisigTask is Test, Script, ITask {
...
+        /// this is to avoid counting subcalls as actions
+        /// cannot set to constant as in tests, the depth will be constant + 1
+        /// for the same account access
+        uint256 minDepth = accountAccesses[0].depth;
+        for (uint256 i = 1; i < accountAccesses.length; i++) {
+            uint256 currentDepth = accountAccesses[i].depth;
+            if (currentDepth < minDepth) {
+                minDepth = currentDepth;
+            }
+        }
```
mds1, line 806: I don't really understand this. In tests, we have an extra top level call, so `minDepth` in a test is initialized to `n+1` where `n` is what `minDepth` is initialized to in a task sim/execution? But why do we set `if (currentDepth < minDepth) minDepth = currentDepth`, i.e. how can the depth decrease if we are starting iteration from `accountAccesses[0]` which presumably has the smallest depth since it's the top level call? (whatever the answer is lets make sure the code comments clarify this)

src/improvements/tasks/MultisigTask.sol
```diff
@@ -813,7 +820,13 @@ abstract contract MultisigTask is Test, Script, ITask {
...
                 accountAccesses[i].account != address(addresses) && accountAccesses[i].account != address(vm)
                     && accountAccesses[i].accessor != address(addresses)
-                    && accountAccesses[i].kind == VmSafe.AccountAccessKind.Call && accountAccesses[i].accessor == multisig
+                    && (
+                        accountAccesses[i].kind == VmSafe.AccountAccessKind.Call
+                            || (
+                                accountAccesses[i].kind == VmSafe.AccountAccessKind.DelegateCall
+                                    && accountAccesses[i].depth == minDepth
+                            )
+                    ) && accountAccesses[i].accessor == multisig
```
mds1, line 829: Can we make this `if` condition more readable by pulling checks into individual named bools and boiling this down to just `if (myFinalBool)`? This is pretty hard to parse

src/improvements/tasks/OPCMBaseTask.sol
```diff
@@ -0,0 +1,120 @@
...
+        /// get task actions
+        (address[] memory targets,, bytes[] memory arguments) = getTaskActions();
+
+        /// create calls array with targets and arguments
+        Call3[] memory calls = new Call3[](targets.length);
+
+        for (uint256 i; i < calls.length; i++) {
+            require(targets[i] != address(0), "Invalid target for multisig");
+            calls[i] = Call3({target: targets[i], allowFailure: false, callData: arguments[i]});
+        }
```
mds1, line 52: For now, do we have an invariant that `targets` has a length of 1, and the target will always be the `OPCM` address? I know this might change in the future with onchain validation, but for now it would be nice if the `address public constant OPCM` from `OPCMUpgradeTemplate.sol` is a virtual, unimplemented `opcm()` function in this contract to ensure the new OPCM template redefines that, and then in this method we can `requri` that `targets` and `calls` have a length of 1, and that the target is `opcm()`

src/improvements/tasks/OPCMBaseTask.sol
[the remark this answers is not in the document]
mds1: Hmm how about we keep it in for now, and in a future PR we can consider making it optional? This way we can see how tedious it is to actually enumerate where we expect storage changes in an OPCM task before deciding to remove this check

src/improvements/tasks/OPCMBaseTask.sol
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    /// @notice get the multicall address for the given safe
+    /// if the safe is the parent multisig, return the delegatecall multicall address
+    /// otherwise if the safe is a child multisig, return the regular multicall address
+    /// @param safe The address of the safe
+    /// @return The address of the multicall
+    function _getMulticallAddress(address safe) internal view returns (address) {
+        return (safe == multisig) ? MULTICALL3_DELEGATECALL_ADDRESS : MULTICALL3_ADDRESS;
+    }
```
mds1, line 97: Can we rename `multisig` to something like `rootMultisig`? I see the natspec for `multisig` is `The address of the multisig for this task` but a task can have 3 involved multisigs in the nested case

src/improvements/template/OPCMUpgradeTemplate.sol
```diff
@@ -0,0 +1,62 @@
...
+contract OPCMUpgradeTemplate is OPCMBaseTask {
+    address public constant OPCM = 0x5BC817c7C3F1A8dCAA01d229Cbdeed9624C80E09;
+
+    /// @notice Struct to store gas limits to be set for a specific L2 chain ID
+    /// @param chainId The ID of the L2 chain
+    /// @param prestate The prestate to be set for the l2 chain
+    struct OPCMUpgrade {
+        uint256 chainId;
+        bytes32 prestate;
+    }
```
mds1, line 16: We should so something like this to make it obvious and self-documenting where the things are that you will have to change. Another example is changing `address public constant OPCM = 0x5BC817c7C3F1A8dCAA01d229Cbdeed9624C80E09` to the `opcm()` function mentioned in a prior comment with a default implementation in this template of `require(false, "this should return the OPCM address based on `block.chainid`")` or something
```suggestion
    /// @notice Struct to store the per-chain inputs for the `OPCM.upgrade()` method.
    /// THIS IS AN EXAMPLE STRUCT for the case where the only per-chain input is the absolute prestate.
    /// YOU MAY NEED TO MODIFY THIS for your specific OPCM instance.
    /// @param chainId The ID of the L2 chain
    /// @param prestate The prestate to be set for the l2 chain
    struct OPCMUpgrade {
        uint256 chainId;
        bytes32 prestate;
    }
```

src/improvements/template/OPCMUpgradeTemplate.sol
```diff
@@ -0,0 +1,62 @@
...
+
+    /// @notice Mapping of l2 chain IDs to their respective prestates
+    mapping(uint256 => bytes32) public opcmUpgrades;
+
+    /// @notice Returns the safe address string identifier
+    /// @return The string "ProxyAdminOwner"
+    function safeAddressString() public pure override returns (string memory) {
+        return "ProxyAdminOwner";
+    }
```
mds1, line 25: This will be the same for ll OPCM tasks so we can move this into `OPCMBaseTask`

src/improvements/template/OPCMUpgradeTemplate.sol
```diff
@@ -0,0 +1,62 @@
...
+        }
+        console.log("address this", address(this));
+        vm.label(OPCM, "OPCM");
+
+        (bool success,) =
+            OPCM.delegatecall(abi.encodeWithSignature("upgrade((address,address,bytes32)[])", opcmConfigs));
+        require(success, "OPCMUpgrateTemplate: failed to upgrade OPCM");
+    }
+
+    function _validate(uint256 chainId) internal view override {}
```
mds1, line 61: In general we should do things like this as the defaults, to prevent devs from forgetting to implement this
```suggestion
    function _validate(uint256 chainId) internal view override {
      require(false, "You must implement the _validate function");
    }
```

test/tasks/MultisigTask.t.sol
```diff
@@ -26,11 +26,14 @@ contract MultisigTaskUnitTest is Test {
     bytes32 public constant MULTISIG_SLOT = bytes32(uint256(36));
 
     /// @notice storage slot for the addresses contract
-    bytes32 public constant MOCK_TARGET_SLOT = bytes32(uint256(51));
+    bytes32 public constant MOCK_TARGET_SLOT = bytes32(uint256(52));
```
mds1, line 29: maintainability suggestion: instead of hardcoding slot numbers we can use forge-std's `stdStorage` library to overwrite slots based on their getter method (also, the natspec here is wrong)

see stdStorage examples here (contract to the readme, packed slots are now supported): https://github.com/foundry-rs/forge-std?tab=readme-ov-file#example-usage-1

test/tasks/NestedMultisigTask.t.sol
```diff
@@ -218,6 +218,94 @@ contract NestedMultisigTaskTest is Test {
...
+        MultiSigOwner[] memory newOwners = new MultiSigOwner[](9);
+        (newOwners[0].walletAddress, newOwners[0].privateKey) = makeAddrAndKey("Owner0");
+        (newOwners[1].walletAddress, newOwners[1].privateKey) = makeAddrAndKey("Owner1");
+        (newOwners[2].walletAddress, newOwners[2].privateKey) = makeAddrAndKey("Owner2");
+        (newOwners[3].walletAddress, newOwners[3].privateKey) = makeAddrAndKey("Owner3");
+        (newOwners[4].walletAddress, newOwners[4].privateKey) = makeAddrAndKey("Owner4");
+        (newOwners[5].walletAddress, newOwners[5].privateKey) = makeAddrAndKey("Owner5");
+        (newOwners[6].walletAddress, newOwners[6].privateKey) = makeAddrAndKey("Owner6");
+        (newOwners[7].walletAddress, newOwners[7].privateKey) = makeAddrAndKey("Owner7");
+        (newOwners[8].walletAddress, newOwners[8].privateKey) = makeAddrAndKey("Owner8");
```
mds1, line 247: We should consider using `lib/optimism/packages/contracts-bedrock/test/safe-tools/SafeTestTools.sol` for cleaner test UX

test/tasks/NestedMultisigTask.t.sol
```diff
@@ -218,6 +218,94 @@ contract NestedMultisigTaskTest is Test {
...
+            multisigTask = new OPCMUpgradeTemplate();
+            multisigTask.approveFromChildMultisig(opcmTaskConfigFilePath, childMultisig, packedSignaturesChild);
+        }
+
+        /// no offchain signatures for the parent multisig
+        bytes memory packedSignaturesParent;
+
+        /// execute the task
+        multisigTask = new OPCMUpgradeTemplate();
+        multisigTask.executeRun(opcmTaskConfigFilePath, packedSignaturesParent);
```
mds1, line 306: I'm not too familiar with how we are doing tests, what is this test actually asserting? I see a `.simulateRun` and `.executeRun` call, but no assertions that e.g. those had equivalent state diffs, or that that execution did what we wanted. The code comments seem to explain what we are doing, but not the "why" behind it

src/improvements/tasks/OPCMBaseTask.sol
blmalone: +1 mentioned here: https://github.com/ethereum-optimism/superchain-ops/pull/599#discussion_r1955283286

src/improvements/tasks/MultisigTask.sol
prat-gpt: I also think that these labels will only be helpful and we can keep these.

src/improvements/tasks/MultisigTask.sol
prat-gpt: > I don't really understand this. In tests, we have an extra top level call, so minDepth in a test is initialized to n+1 where n is what minDepth is initialized to in a task sim/execution?

Yes, right.

> But why do we set if (currentDepth < minDepth) minDepth = currentDepth, i.e. how can the depth decrease if we are starting iteration from accountAccesses[0] which presumably has the smallest depth since it's the top level call? (whatever the answer is lets make sure the code comments clarify this)

I was not sure whether account accesses are in order so just to be sure I added the loop. I just re read the [documentation](https://book.getfoundry.sh/cheatcodes/stop-and-return-state-diff#description) and found that account accesses are in order for sure.
`The ordering of AccountAccess records reflect the EVM execution order of their associated operations.`.
So I have updated this logic to get the depth from accountAccesses[0] and also updated code comments.

src/improvements/tasks/OPCMBaseTask.sol
prat-gpt: I think this invariant is a default check and is better suited for `validate` function so I have added this invariant by overriding `validate`. Also have added the `opcm()` virtual unimplemented function.

src/improvements/template/OPCMUpgradeTemplate.sol
prat-gpt: `_validate` is already a unimplemented virtual function is the `MultisigTask` which means the template has to implement this function. I use a dummy OPCM contract, there is nothing to validate per chain for this opcm and hence I added this empty function. If I add your suggestion in the template the example opcm task will revert. But I get the overall idea that you want to convey here, will keep this in mind that we must force the devs to override a few functions.

test/tasks/MultisigTask.t.sol
prat-gpt: This will require upgrading `forge-std`. with current version I am unable to get the slot. I think this can be done is a separate PR.
Fixed the natspec error.

test/tasks/NestedMultisigTask.t.sol
prat-gpt: I would like to tackle this in a separate PR as well.

test/tasks/NestedMultisigTask.t.sol
prat-gpt: In this test `.sumlateRun` generates the data to sign for each of the child multisig EOA owners. Child safe owner EOAs are updated and the data to sign is signed by their private keys. The approve hash transaction for each child multisig is broadcasted using `approveFromChildMultisig` function and finally `executeRun` broadcasts the final opcm delegate call successfully. It is asserted that the call was a success. This proves that the data to sign generated by `simulateRun` is correct. 

There are no assertions because the dummy OPCM that I am using does not do any state updates. If you take a look at `testNestedExecuteWithSignatures` test, there are assertions at the end.

Added comment on top of the test explaining the why behind the test.

src/improvements/tasks/MultisigTask.sol
```diff
@@ -798,6 +793,18 @@ abstract contract MultisigTask is Test, Script, ITask {
 
         _processStateDiffChanges(accountAccesses);
 
+        /// get the minimum depth of the calls, we only care about the top level calls
+        /// this is to avoid counting subcalls as actions
+        /// cannot set to constant as in tests, the depth will be constant + 1
+        /// for the same account access
```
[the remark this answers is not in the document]
prat-gpt: no need of this after the updated PR.

test/tasks/MultisigTask.t.sol
prat-gpt: Added issue: https://github.com/ethereum-optimism/superchain-ops/issues/605

test/tasks/NestedMultisigTask.t.sol
prat-gpt: Added issue: https://github.com/ethereum-optimism/superchain-ops/issues/605

src/improvements/tasks/MultisigTask.sol
blmalone: Sounds good, resolving.

src/improvements/tasks/OPCMBaseTask.sol
```diff
@@ -0,0 +1,103 @@
...
+
+    /// @notice set the multicall address
+    /// overrides MultisigTask to set the multicall address to the delegatecall multicall address
+    function _setMulticallAddress() internal override {
+        multicallTarget = MULTICALL3_DELEGATECALL_ADDRESS;
+    }
+
+    /// @notice overrides to do nothing per chain
+    /// all the chains are handled in a single call to OPCM contract
+    function _buildPerChain(uint256 chainId) internal override {}
```
blmalone, line 102: Maybe we intentionally revert in here if devs try to invoke this function. For this version, we know this function will not be used with OPCM.

src/improvements/tasks/OPCMBaseTask.sol
```diff
@@ -0,0 +1,103 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import {Test} from "forge-std/Test.sol";
+
+import {Signatures} from "@base-contracts/script/universal/Signatures.sol";
+import {IGnosisSafe, Enum} from "@base-contracts/script/universal/IGnosisSafe.sol";
+
+import {MultisigTask} from "src/improvements/tasks/MultisigTask.sol";
+import {AddressRegistry as Addresses} from "src/improvements/AddressRegistry.sol";
```
blmalone, line 10: This is unused.

src/improvements/tasks/OPCMBaseTask.sol
```diff
@@ -0,0 +1,103 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import {Test} from "forge-std/Test.sol";
+
+import {Signatures} from "@base-contracts/script/universal/Signatures.sol";
+import {IGnosisSafe, Enum} from "@base-contracts/script/universal/IGnosisSafe.sol";
```
blmalone, line 7: This is unsed

src/improvements/tasks/OPCMBaseTask.sol
```diff
@@ -0,0 +1,103 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import {Test} from "forge-std/Test.sol";
+
+import {Signatures} from "@base-contracts/script/universal/Signatures.sol";
```
blmalone, line 6: This is unused.

src/improvements/tasks/OPCMBaseTask.sol
```diff
@@ -0,0 +1,103 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import {Test} from "forge-std/Test.sol";
```
blmalone, line 4: This is unused.

src/improvements/tasks/OPCMBaseTask.sol
blmalone: I'll add this in.

src/improvements/template/OPCMUpgradeTemplate.sol
```diff
@@ -0,0 +1,77 @@
...
+            abi.decode(vm.parseToml(vm.readFile(taskConfigFilePath), ".opcmUpgrades.opcmPrestates"), (OPCMUpgrade[]));
+
+        for (uint256 i = 0; i < opcmUpgrade.length; i++) {
+            opcmUpgrades[opcmUpgrade[i].chainId] = opcmUpgrade[i].prestate;
+        }
+    }
+
+    /// @notice build the task action for all l2chains in the task
+    /// in a single call to the OPCM.upgrade() function.
+    function build() public override buildModifier {
```
blmalone, line 54: Shouldn't we be overriding `_buildSingle` in this template and not build?

blmalone: So it looks like this:

```
    function _buildSingle() internal override {
        Addresses.ChainInfo[] memory chains = addresses.getChains();
        OpChainConfig[] memory opcmConfigs = new OpChainConfig[](chains.length);

            opcmConfigs[i] = OpChainConfig({
                systemConfigProxy: addresses.getAddress("SystemConfigProxy", chains[i].chainId),
                proxyAdmin: addresses.getAddress("ProxyAdmin", chains[i].chainId),
                absolutePrestate: opcmUpgrades[chains[i].chainId]
            });
        }
        console.log("address this", address(this));
        vm.label(opcm(), "OPCM");

        (bool success,) =
            opcm().delegatecall(abi.encodeWithSignature("upgrade((address,address,bytes32)[])", opcmConfigs));
        require(success, "OPCMUpgrateTemplate: failed to upgrade OPCM");
    }
    ```

File: src/improvements/template/OPCMUpgradeTemplate.sol [hunk 24]
```diff
@@ -0,0 +1,77 @@
...
+        OpChainConfig[] memory opcmConfigs = new OpChainConfig[](chains.length);
+
+        for (uint256 i = 0; i < chains.length; i++) {
+            opcmConfigs[i] = OpChainConfig({
+                systemConfigProxy: addresses.getAddress("SystemConfigProxy", chains[i].chainId),
+                proxyAdmin: addresses.getAddress("ProxyAdmin", chains[i].chainId),
+                absolutePrestate: opcmUpgrades[chains[i].chainId]
+            });
+        }
+        console.log("address this", address(this));
```
blmalone, line 65: Remove.

src/improvements/tasks/MultisigTask.sol
```diff
@@ -793,26 +810,42 @@ abstract contract MultisigTask is Test, Script, ITask {
...
                 }
             }
 
-            /// only care about top level calls from the multisig,
-            /// static calls are ignored,
             /// calls to and from Addresses and the vm contract are ignored
-            /// ignore calls to vm in the build function
+            bool accountIsNotAddressesOrVm =
+                accountAccesses[i].account != address(addresses) && accountAccesses[i].account != address(vm);
+            bool accessorIsNotAddresses = accountAccesses[i].accessor != address(addresses);
```
mds1, line 838: nit for a future PR: rename the `addresses` var to `addrRegistry` everywhere, the vaguely named `addresses` keeps confusing me lol

src/improvements/tasks/MultisigTask.sol
```diff
@@ -793,26 +810,42 @@ abstract contract MultisigTask is Test, Script, ITask {
...
+                && accountAccesses[i].depth == topLevelDepth;
+            /// only record actions from the parent multisig
+            bool accessorIsParentMultisig = accountAccesses[i].accessor == parentMultisig;
+
             if (
-                accountAccesses[i].account != address(addresses) && accountAccesses[i].account != address(vm)
-                    && accountAccesses[i].accessor != address(addresses)
-                    && accountAccesses[i].kind == VmSafe.AccountAccessKind.Call && accountAccesses[i].accessor == multisig
+                accountIsNotAddressesOrVm && accessorIsNotAddresses && (isCall || isTopLevelDelegateCall)
+                    && accessorIsParentMultisig
```
mds1, line 848: I think the `Action` struct will need to be updated to also specify whether the parent multisig is executing a call or delegatecall, and the description correspondingly updated?

src/improvements/template/OPCMUpgradeTemplate.sol
mds1: > If I add your suggestion in the template the example opcm task will revert

Here you are specifically referring to execution of the `test/tasks/mock/configs/ExampleOPCMUpgradeTemplate.toml` config file right? If so, that config file points to a template name of `OPCMUpgradeTemplate`. But in reality, `OPCMUpgradeTemplate` is a _template for templates_, and not a template itself, right? So we would never want to execute that template directly?

If that's correct, the `ExampleOPCMUpgradeTemplate.toml` should be updated to point to `templateName = "TestOPCMUpgradeVxyz"`, where `TestOPCMUpgradeVxyz is OPCMUpgradeTemplate` and implements the methods that must be implemented to avoid reverting

test/tasks/NestedMultisigTask.t.sol
mds1: Got it, thank you. Another type of test we should have if we don't have it already, is config files of real templates. For example, we can have a `ProtocolVersionsUpdate` template, and a test config file specifies some random version strings to upgrade too. The `ProtocolVersionsUpdate` solidity template has validations, and CI runs the test config files for these real templates. This gives us confidence that our templates are always working and pass the validations

src/improvements/tasks/MultisigTask.sol
blmalone: update in new PR that builds on this one at commit: https://github.com/ethereum-optimism/superchain-ops/pull/606/commits/672353b54374b0aa5605825ad4f8d6d074ddfe0e

src/improvements/tasks/MultisigTask.sol
blmalone: update in new PR that builds on this one at commit: https://github.com/ethereum-optimism/superchain-ops/pull/606/commits/62456eb46ff582e38d906a743fcca1e4ec3c8540

src/improvements/tasks/MultisigTask.sol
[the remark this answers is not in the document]
blmalone: Is this okay to be resolved @ElliotFriedman - I'm working through open comments.

src/improvements/tasks/MultisigTask.sol
```diff
@@ -764,6 +754,11 @@ abstract contract MultisigTask is Test, Script, ITask {
         );
     }
 
+    /// @notice set the multicall address
```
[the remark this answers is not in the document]
blmalone: I see this was implemented so resolving comment.

src/improvements/tasks/MultisigTask.sol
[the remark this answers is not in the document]
blmalone: resolving

src/improvements/tasks/OPCMBaseTask.sol
```diff
@@ -0,0 +1,120 @@
...
+    /// @param safe The address of the safe
+    /// @return The address of the multicall
+    function _getMulticallAddress(address safe) internal view returns (address) {
+        return (safe == multisig) ? MULTICALL3_DELEGATECALL_ADDRESS : MULTICALL3_ADDRESS;
+    }
+
+    /// @notice overrides to start pranking the
+    /// multisig with delegatecall flag set to true
+    function _startBuild() internal override {
+        vm.startPrank(multisig, true);
```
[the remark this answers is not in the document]
blmalone: this was implemented so resolving this thread.

src/improvements/template/OPCMUpgradeTemplate.sol
blmalone: update in new PR that builds on this one at commit: https://github.com/ethereum-optimism/superchain-ops/pull/606/commits/ad4c2411600b4c5daa045713beb7868b07c43752

test/tasks/NestedMultisigTask.t.sol
blmalone: Gotcha, I've capture this in an issue as a follow up task: https://github.com/ethereum-optimism/superchain-ops/issues/610

src/improvements/template/OPCMUpgradeTemplate.sol
blmalone: Also fixed what I think are most of the natspec violations here: https://github.com/ethereum-optimism/superchain-ops/pull/606/commits/50cd82ba5d5d68c822c13e70138ca0f334ece9ec

src/improvements/tasks/MultisigTask.sol
blmalone: quick follow up here to no rename imports.

blmalone:
Closing as https://github.com/ethereum-optimism/superchain-ops/pull/606 is finishing out PR comments. All commits have been cherry picked on new branch.

Issue #596, opened 2025-02-13
Safety checks after performing a delegate call


blmalone:
A new feature in superchain-ops 2.0 is enabling the parent multisig to delegatecall to a contract. Our initial use case involves OPCM. Instead of using [Multicall3.sol](https://github.com/mds1/multicall/blob/main/src/Multicall3.sol) for batching actions, we are switching to [Multicall3Delegatecall.sol](https://github.com/ethereum-optimism/superchain-ops/blob/d38a5d835ebf93f81be788004266cf59a205bbf4/src/Multicall3Delegatecall.sol), which performs a delegatecall to OPCM.

**_Why use delegatecall?_**
The primary reason is to prevent any state changes in OPCM when executing functions. Allowing state modifications in OPCM could lead to non-deterministic behavior for chains integrating with OPCM in the future.

**_Security precautions to detect unwanted behavior early_**
By using delegatecall from the upgrade-controller, we introduce a risk: state changes on the upgrade-controller contract itself, which m**ust be avoided at all costs**.

To mitigate this, I propose that superchain-ops’ simulation step includes a post-execution check to verify that no state changes occurred in both the upgrade-controller contract and OPCM.

maurelian:
Agree we definitely need to enforce this property. FWIW we have always been delegatecalling to Multicall3, so this is not entirely new, but it does add more risk since the OPCM logic is more complex and will change overtime. 

I think this could be fairly easily done in [`checkStateDiff()`](https://github.com/ethereum-optimism/superchain-ops/blob/95d16a2d05019dad3449d2ec2f5ee3180e7b9e44/src/JsonTxBuilderBase.sol#L59), since the only diff should be the nonce.

PR #602, opened 2025-02-13
Add checkStateDiff Function

ElliotFriedman:
This PR adds a checkStateDiff function which takes an array of all state diffs for a given task execution.

This is the first part of adding safety checks for OPCM upgrades https://github.com/ethereum-optimism/superchain-ops/issues/596


blmalone (approved):

src/improvements/tasks/MultisigTask.sol
```diff
@@ -840,8 +840,16 @@ abstract contract MultisigTask is Test, Script, ITask {
         }
     }
 
+    /// @notice helper function that can be overridden by template contracts to
+    /// check the state changes applied by the task. This function can check
+    /// that only the nonce changed in the parent multisig when executing a task
+    /// by checking the slot and address where the slot changed.
+    function checkStateDiff(VmSafe.AccountAccess[] memory) internal view virtual {}
```
mds1, line 847: We should avoid empty default implementations, as they make it too easy to forget to implement safety-critical functions like this (need to use `require(false, msg)` instead of `revert(msg)` to prevent solidity from warning about unreachable code in child contracts
```suggestion
    function checkStateDiff(VmSafe.AccountAccess[] memory) internal view virtual {
      require(false, "checkStateDiff must be implemented");
    }
```

maurelian: Even if for some unexpected reason it makes sense for templates to leave this function empty, it's good to include this revert, so that the template will be forced to implement the empty function body and be reviewed thusly.

PR #603, opened 2025-02-13
Justfile Improvements

ElliotFriedman:
Changes to justfiles:

- Automatically find the template name from the config.toml file
- Remove owner safe as an env var as it is no longer needed with the new tooling
- Remove need to pass RPC URL to justfile as this is automatically detected based on the directory the config.toml file is in

Commands to test:

```
cd src/improvements/tasks/example/eth/000-initial-example-task
just --justfile ../../../../../improvements/single.just sign
```


src/improvements/get-rpc-url.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
```
blmalone, line 1: Can this file be moved into `src/improvements/script`?

src/improvements/single.just
```diff
@@ -1,39 +1,40 @@
...
+export TASK_PATH := invocation_directory()
+export SCRIPT_PATH := justfile_directory()
+
+# Get RPC URL based on task path
+rpc_cmd := SCRIPT_PATH + "/script/get-rpc-url.sh \"" + TASK_PATH + "\""
+export rpcUrl := shell(rpc_cmd)
+
+# Fetch the template name from the config.toml file
+cmd := "sed -n 's/^[[:space:]]*templateName[[:space:]]*=[[:space:]]*\\\"\\([^\\\"]*\\)\\\".*/\\1/p' " + TASK_PATH + "/config.toml"
+export SCRIPT_NAME := shell(cmd)
```
mds1, line 14: We already have [yq](https://github.com/mikefarah/yq) as a dependency, which we can use to parse the config file in a way that's more readable then sed

src/improvements/single.just
```diff
@@ -1,39 +1,40 @@
...
   echo "Using script ${script}"
-  echo "Safe address: ${ownerSafe}"
-  echo "Getting signer address..."
-
-  if [ ! -z "$SIMULATE_WITHOUT_LEDGER" ]; then
-    signer=$(cast call ${ownerSafe} "getOwners()(address[])" -r ${rpcUrl} | grep -oE '0x[a-fA-F0-9]{40}' | head -n1)
-  else
-    signer=$(cast wallet address --ledger --mnemonic-derivation-path "m/44'/60'/{{hdPath}}'/0/0")
-  fi
-  echo "Simulating with: ${signer}"
```
mds1, line 24: Why was this deleted? THese logs helps signers know they have the correct wallet address connected

src/improvements/single.just
```diff
@@ -1,39 +1,40 @@
...
   echo ""
 
   forge build
   forge script ${script} \
     --rpc-url ${rpcUrl} \
     --sig "simulateRun(string)" ${config} \
-    --sender ${signer}
```
mds1, line 31: Same question, why was this deleted?

blmalone:
Output from my most recent run at commit: c45a976cd99cf0933a9211490bddba0bdc8ed6c0
CleanShot 2025-02-13 at 19 37 00@2x

src/improvements/single.just
mds1: simulate is used before we have all the signatures, signers will use it to simulate as their wallet and review the state diff before deciding to actually sign — did we remove that use case?

src/improvements/single.just
mds1: Hm, this is in `single.just` so there are no child multisigs?

src/improvements/nested.just
```diff
@@ -1,38 +1,51 @@
...
   if [ "{{whichSafe}}" == "foundation" ]; then
-    safe="{{foundationSafe}}"
-  fi
-  if [ "{{whichSafe}}" == "council" ]; then
-    safe="{{councilSafe}}"
-  fi
-  if [ "{{whichSafe}}" == "chain-governor" ]; then
-    if [ -z "{{chainGovernorSafe}}" ]; then
-      echo "Error: CHAIN_GOVERNOR_SAFE is not set for chain-governor." >&2
+    if [ -z "$FOUNDATION_SAFE" ]; then
```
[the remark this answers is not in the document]
mds1: Got it, I would suggest reverting back to the old UX since it's more user-friendly. We can maintain a global mapping from `nested root safe -> name -> child safe address`, and just pas the `just .../nested.just {council|foundation|chainGovernor}` still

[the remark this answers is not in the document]
blmalone: Is it possible to have the global mapping include the role labels (`foundation`, `council` and `chain-governor`)? 

For example, the mapping for OP Mainnet Upgrade Controller Nested Safe might looks like:
```
ParentMultisig: 
0x5a0Aae59D09fccBdDb6C6CcEB07B7279367C3d2A

ChildMultisigs:
[
0x847B5c174615B1B7fDF770882256e2D3E95b9D92 -> foundation
0xc2819DC788505Aac350142A7A707BF9D03E3Bd03 -> council
]
```

If a user passes through: `foundation`, the script could:
1. Look for the parent multisig of the task
2. Grab all child multsig owners on the parent multisig
3. From this list of owners decide which multisig should be used for `foundation` (reference a mapping we keep locally)

blmalone (commented):
can confirm signing is working for me.

src/improvements/single.just
mds1: So in the `single` flow we still use `approveHash`? If so I believe that is different from our current flow (cc @blmalone to confirm) so if we are changing that let's be really clear about the diff, it's consequences, and the rationale for it somewhere

src/improvements/nested.just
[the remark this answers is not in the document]
mds1: On the standup today we said to avoid increasing scope we'll keep the env var file around for now just to specify the foudation, council, and chain gov addresses, this way the UX stays the same (i.e. being able to specify `just simulate coucil` — is that correct?

mds1:
@ElliotFriedman Super helpful break down, thank you! One comment would be to also add the Domain hash and Message hash as individual lines on the `sign` terminal logs, since that's how ledger displays it (even though it's the same info as the full data to sign string)

mds1 (approved):

src/improvements/single.just
blmalone: This seems to be resolved.

src/improvements/nested.just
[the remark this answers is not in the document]
blmalone: This seems to have been settled. I think we can resolve this comment now.

src/improvements/nested.just
```diff
@@ -1,38 +1,51 @@
...
-export scriptName := env_var('SCRIPT_NAME')
-
-# Accounts
-export councilSafe := env_var("COUNCIL_SAFE")
-export foundationSafe := env_var("FOUNDATION_SAFE")
-export chainGovernorSafe := env_var_or_default('CHAIN_GOVERNOR_SAFE', '')
-export randomPersonEoa := "0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045"
+randomPersonEoa := "0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045"
 
 simulate whichSafe hdPath='0':
```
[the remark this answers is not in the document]
blmalone: Awesome, I think we can resolve this thread now. I can confirm logs match.

src/improvements/nested.just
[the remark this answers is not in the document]
blmalone: I'm going to close this thread. Please DM if it shouldn't be closed. Lets follow up in a separate PR.

src/improvements/single.just
blmalone: I'm going to close this thread. Please DM if it shouldn't be closed. Lets follow up in a separate PR.

blmalone (approved):

PR #607, opened 2025-02-14
feat: autogenerate decoded state diffs for validations

mds1:
Autogenerates decoded state diffs to the terminal, requires the change in https://github.com/ethereum-optimism/optimism/pull/14378 which is why the submodule is updated. Functionality works but doesn't decode mappings yet

This will also be easy to port to the new superchain-ops approach


tasks/sep/zora-002-fp-upgrade/NestedSignFromJson.s.sol
```diff
@@ -116,6 +118,8 @@ contract NestedSignFromJson is OriginalNestedSignFromJson, CouncilFoundationNest
         checkDisputeGameFactory();
         checkPermissionedDisputeGame();
         console.log("All assertions passed!");
+
+        StateDiffDecoder.decode(accesses);
```
mds1, line 122: This file shows example usage, this feature is not used by default

src/improvements/tasks/StateDiffDecoder.sol
```diff
@@ -0,0 +1,536 @@
...
+                    oldValue = toAddress(_oldValue, offset);
+                    newValue = toAddress(_newValue, offset);
+                } else if (kind.contains("uint")) {
+                    oldValue = toUint(_oldValue, offset);
+                    newValue = toUint(_newValue, offset);
+                }
+
+                string memory label = layout[i]._label;
+                return DecodedSlot({kind: kind, oldValue: oldValue, newValue: newValue, summary: label, detail: ""});
+            }
```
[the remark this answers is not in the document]
mds1: This does not support mappings yet, that will come later

src/improvements/tasks/StateDiffDecoder.sol
```diff
@@ -0,0 +1,536 @@
...
+        }
+        if (_slot == GAS_PAYING_TOKEN_SYMBOL_SLOT) {
+            return DecodedSlot({
+                kind: "string",
+                oldValue: vm.toString(_oldValue),
+                newValue: vm.toString(_newValue),
+                summary: "Gas paying token symbol",
+                detail: "Unstructured storage slot for the symbol of the gas paying token."
+            });
+        }
```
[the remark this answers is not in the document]
mds1: That feels like a coupling of concerns, this file is purely decoding of state diffs, and we have existing approaches to handle validating the correctness of storage writes

src/improvements/tasks/StateDiffDecoder.sol
```diff
@@ -0,0 +1,536 @@
...
+                        contractName = "OptimismPortal2";
+                    }
+                    // We make a call to see if the contract is a Safe.
+                    if (isGnosisSafe(target)) {
+                        contractName = string.concat(contractName, " (GnosisSafe)");
+                    }
+
+                    return (vm.parseUint(chainId), contractName);
+                }
+            }
```
mds1, line 478: @ElliotFriedman @blmalone @prat-gpt when we implement this decoded logging into the new version, we can replace this block with a lookup in the address registry using the `getAddressInfo` method that @prat-gpt added

src/libraries/StateDiffDecoder.sol
```diff
@@ -0,0 +1,536 @@
...
+    struct JsonStorageLayout {
+        string _bytes;
+        string _label;
+        uint256 _offset;
+        string _slot;
+        string _type;
+    }
+
+    // forgefmt: disable-start
+    bytes32 internal constant ERC1967_IMPL_SLOT = 0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc;
```
blmalone, line 57: We've defined this same constant in Constants.sol in `contracts-bedrock/src/libraries`. Not sure if you want to import from there or not.

src/libraries/StateDiffDecoder.sol
```diff
@@ -0,0 +1,536 @@
...
+        // from the monorepo at that tag.
+        string memory basePath = "/lib/optimism/packages/contracts-bedrock/snapshots/storageLayout/";
+        string memory artifactName = _contractName.endsWith("(GnosisSafe)") ? "GnosisSafe" : _contractName;
+        string memory path = string.concat(vm.projectRoot(), basePath, artifactName, ".json");
+
+        string memory storageLayout;
+        try vm.readFile(path) returns (string memory result) {
+            storageLayout = result;
+        } catch {
+            console.log("\x1B[33m[WARN]\x1B[0m Failed to read storage layout file at %s", path);
```
blmalone, line 406: I see three of these logs when I run: 

```
cd superchain-ops/tasks/sep/zora-002-fp-upgrade
SIMULATE_WITHOUT_LEDGER=1 just --dotenv-path $(pwd)/.env --justfile ../../../nested.just simulate foundation
```

They're being logged when we try to read the storage layout of the GnosisSafe. Is there any reason why we don't have the GnosisSafe storage layout at: `snapshots/storageLayout/`? Or how do we add it to the directory so this error doesn't happen?

blmalone (approved):
Looks good to me! Just left two non-blocking comments.

src/libraries/StateDiffDecoder.sol
mds1: will make this change in a follow up PR to avoid the need for re-approval here

src/libraries/StateDiffDecoder.sol
mds1: You may need to update your monorepo dep version with `forge install`, see https://github.com/ethereum-optimism/optimism/pull/14378

src/libraries/StateDiffDecoder.sol
mds1: actually just going to leave this and avoid the `Constants.sol` import/dependency, since this is standardized

Issue #608, opened 2025-02-14
Implement checkStateDiff Function in Example Templates


ElliotFriedman:
Currently, the GasConfig, OwnerTransfer, SetGameType and DisputeGameUpgrade template have an empty `checkStateDiff` function implemented.

At some point we should revisit these tasks and implement this function.

PR #609, opened 2025-02-14
Make checkStateDiff Revert

ElliotFriedman:
Make the checkStateDiff function revert by default in MultisigTask, which means all new templates must have at the very least an empty implementation.

Added TODO's to the code to come back and fill out the checkStateDiff function in later PR's.

An issue has been created for filling in these checkStateDiff functions at a later point in time https://github.com/ethereum-optimism/superchain-ops/issues/608

This PR created an unreachable code warning due to checkStateDiff reverting. In order to work around this and stop CI failures, two flags were added to the forge build command in CI:

```
# this code ignores unreachable code warnings
 --ignored-error-codes=5740

# this code ignores contract code size over the EVM bytecode limit
--ignored-error-codes=5574
```

Foundry error code source: https://github.com/foundry-rs/foundry/blob/51b75c83045a963a48f9cd8d765e0f2e775acb3e/crates/config/src/error.rs#L244C13-L255

The compiler was throwing spdx license warnings, this was fixed by adding spdx identifiers to all template files with an MIT license.

Other MultisigTask changes:
- `validate` function now takes an array of AccountAccesses as a parameter, consolidating the location of the validation of the state diffs
- `simulateRun` function now takes signatures as a second parameter. This parameter is optional and can be empty
- `simulate` function now takes signatures as a parameter. This parameter is optional and can be empty
- `simulate` function now returns an array of AccountAccesses which is then passed to validate
- tests do not use `executeRun` function because that function performs a broadcast, which wipes out all of the mocked state, causing test failures
- If a task developer fails to add an address whose storage is touched in a task, foundry will crash with an opaque error message when the task is run from the tests. I could not pin this down, but it is likely an issue with interactions of foundry test and `startStateDiffRecording` functions. This issue did not manifest or replicate in tasks with logic errors when running the same task with the `forge script` command.
- `_taskStorageWrites` automatically adds the safe address too as this address has its storage changed in the simulate and execute functions.


src/improvements/tasks/MultisigTask.sol
```diff
@@ -893,7 +921,9 @@ abstract contract MultisigTask is Test, Script, ITask {
     /// check the state changes applied by the task. This function can check
     /// that only the nonce changed in the parent multisig when executing a task
     /// by checking the slot and address where the slot changed.
-    function checkStateDiff(VmSafe.AccountAccess[] memory) internal view virtual {}
+    function checkStateDiff(VmSafe.AccountAccess[] memory) internal view virtual {
+        revert("checkStateDiff must be implemented");
```
mds1, line 925: This tricks solc into think subsequent code is reachable since it does not statically analyze `false`. This will let you avoid the need to suppress warnings, which is safer
 `foundry.toml` to ensure
```suggestion
        require(false, "checkStateDiff must be implemented");
``1

File: .circleci/config.yml [hunk 2]
```diff
@@ -274,7 +274,7 @@ jobs:
           command: |
             just install
             forge --version
-            forge build --deny-warnings
+            forge build --deny-warnings --ignored-error-codes=5740 --ignored-error-codes=5574
```
mds1, line 277: The bytecode limit only applies to src contracts that we intended to deploy, it does not apply to script or test contracts. We should instead make sure the corresponding contracts are marked as scripts or tests accordingly, which I believe forge does by checking for the `*.t.sol` and `*.s.sol` extensions. All of our templates and tasks should probably have `*.s.sol` since they are invoked via `forge script`?

src/improvements/tasks/MultisigTask.sol
mds1, line 925: Maybe we are still miscommunicating here. The default implementation of this method should be identical to what it is in the legacy superchain-ops, and it should run on every task (every template) by default. This PR requires every task template to reimplement this method https://github.com/ethereum-optimism/superchain-ops/blob/9f1cb1aa2f8bf3938289a50e2ff5317dd22c773f/src/JsonTxBuilderBase.sol#L59

The new `validate` signature and where we call `checkStateDiff` LGTM now, it's just this default implementation that we need to fix. I will push a commit to this PR so it's clear what I mean

test/tasks/Regression.t.sol
```diff
@@ -16,7 +16,7 @@ contract RegressionTest is Test {
...
         vm.createSelectFork("mainnet", 21724199);
         MultisigTask multisigTask = new GasConfigTemplate();
-        multisigTask.simulateRun(taskConfigFilePath);
+        multisigTask.simulateRun(taskConfigFilePath, "");
```
mds1, line 19: We should just have an overload of `simulateRun` that does not take a bytes arg, and internally that method just calls `simulateRun(taskConfigFilePath, "")`. This would be cleaner and more clear than empty strings and `0x` everywhere

.circleci/config.yml
mds1, line 1: > If a task developer fails to add an address whose storage is touched in a task, foundry will crash with an opaque error message when the task is run from the tests. I could not pin this down, but it is likely an issue with interactions of foundry test and `startStateDiffRecording` functions. This issue did not manifest or replicate in tasks with logic errors when running the same task with the `forge script` command.

What's the error message? And are you able to consistently reproduce this? I am curious if this is a bug in our code that we need to fix, or a bug in foundry that we need to open an issue for

src/improvements/tasks/MultisigTask.sol
```diff
@@ -153,28 +153,40 @@ abstract contract MultisigTask is Test, Script, ITask {
...
+    {
         /// sets safe to the safe specified by the current template from addresses.json
         _taskSetup(taskConfigFilePath);
 
         /// now execute task actions
         build();
-        simulate();
-        validate();
+        VmSafe.AccountAccess[] memory accountAccesses = simulate(signatures);
+        validate(accountAccesses);
```
mds1, line 167: Nice, this is much cleaner than the old stateful approach, and very clear 👌

src/improvements/template/DisputeGameUpgradeTemplate.sol
```diff
@@ -1,6 +1,9 @@
+// SPDX-License-Identifier: MIT
 pragma solidity 0.8.15;
 
 import {IDisputeGameFactory, IDisputeGame} from "@eth-optimism-bedrock/interfaces/dispute/IDisputeGameFactory.sol";
+import {VmSafe, Vm} from "forge-std/Vm.sol";
```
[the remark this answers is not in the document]
blmalone: The AddressRegistry in this file is also an unused import.

src/improvements/tasks/ITask.sol
```diff
@@ -28,13 +33,13 @@ abstract contract ITask {
     /// @notice actually simulates the task.
```
blmalone, line 33: What does this comment mean? I'm not following the Timelock Controller reference or what is meant by _actually_. In comparison to what?

src/improvements/tasks/ITask.sol
```diff
@@ -17,7 +19,10 @@ abstract contract ITask {
 
     /// @notice runs all task steps
     /// 1. builds, simulates, validates and then prints the task
```
blmalone, line 21:
```suggestion
    /// builds, simulates, validates and then prints the task
```

src/improvements/tasks/MultisigTask.sol
```diff
@@ -439,6 +476,9 @@ abstract contract MultisigTask is Test, Script, ITask {
         for (uint256 i = 0; i < chains.length; i++) {
             _validate(chains[i].chainId);
         }
+
+        /// reverts by default and must be overridden by the child templates
```
blmalone, line 480: Is this true? I see `checkStateDiff` in this function that is implemented and doesn't seem to revert.

src/improvements/template/GasConfigTemplate.sol
```diff
@@ -1,6 +1,8 @@
+// SPDX-License-Identifier: MIT
 pragma solidity 0.8.15;
 
 import {SystemConfig} from "@eth-optimism-bedrock/src/L1/SystemConfig.sol";
+import {VmSafe, Vm} from "forge-std/Vm.sol";
```
blmalone, line 5: this is unused

src/improvements/template/GasConfigTemplate.sol
```diff
@@ -1,6 +1,8 @@
...
 
 import {MultisigTask} from "src/improvements/tasks/MultisigTask.sol";
 import {AddressRegistry} from "src/improvements/AddressRegistry.sol";
```
blmalone, line 8: this is unused

src/improvements/template/SetGameTypeTemplate.sol
```diff
@@ -1,9 +1,11 @@
+// SPDX-License-Identifier: MIT
 pragma solidity 0.8.15;
 
 import {
     DeputyGuardianModule, IOptimismPortal2, GameType
 } from "@eth-optimism-bedrock/src/safe/DeputyGuardianModule.sol";
 import {LibGameType} from "@eth-optimism-bedrock/src/dispute/lib/LibUDT.sol";
+import {VmSafe, Vm} from "forge-std/Vm.sol";
```
blmalone, line 8: this is unused

src/improvements/template/SetGameTypeTemplate.sol
```diff
@@ -1,9 +1,11 @@
...
 
 import {MultisigTask} from "src/improvements/tasks/MultisigTask.sol";
 import {AddressRegistry} from "src/improvements/AddressRegistry.sol";
```
blmalone, line 11: this is unused

src/improvements/template/TestOPCMUpgradeVxyz.sol
```diff
@@ -1,4 +1,7 @@
-pragma solidity ^0.8.0;
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import {VmSafe} from "forge-std/Vm.sol";
```
blmalone, line 4: this is unused

src/improvements/template/TransferOwnerTemplate.sol
```diff
@@ -1,6 +1,8 @@
+// SPDX-License-Identifier: MIT
 pragma solidity 0.8.15;
 
 import {ProxyAdmin} from "@eth-optimism-bedrock/src/universal/ProxyAdmin.sol";
+import {VmSafe, Vm} from "forge-std/Vm.sol";
```
blmalone, line 5: this is unused

blmalone: As well as AddressRegistry.

test/tasks/mock/template/IncorrectGasConfigTemplate2.sol
```diff
@@ -1,3 +1,4 @@
+// SPDX-License-Identifier: MIT
 pragma solidity 0.8.15;
 
 import {GasConfigTemplate} from "src/improvements/template/GasConfigTemplate.sol";
```
blmalone, line 4: AddressRegistry is an unused import

test/tasks/mock/template/IncorrectGasConfigTemplate1.sol
```diff
@@ -1,3 +1,4 @@
+// SPDX-License-Identifier: MIT
 pragma solidity 0.8.15;
 
 import {SystemConfig} from "@eth-optimism-bedrock/src/L1/SystemConfig.sol";
```
blmalone, line 4: Unused import + AddressRegistry is also unused.

src/improvements/tasks/MultisigTask.sol
mds1: Thank you! Let's make sure to add tests for this new functionality too

src/improvements/tasks/MultisigTask.sol
```diff
@@ -725,6 +761,9 @@ abstract contract MultisigTask is Test, Script, ITask {
         }
     }
 
+    /// @notice overridden in templates
+    function getCodeExceptions() internal view virtual returns (address[] memory) {}
```
mds1, line 765: Let's avoid no-op default implementations
```suggestion
    /// @notice overridden in templates
    function getCodeExceptions() internal view virtual returns (address[] memory);
```

src/improvements/tasks/ITask.sol
blmalone: Ok resolving.

src/improvements/tasks/ITask.sol
blmalone: Ok resolving.

src/improvements/tasks/MultisigTask.sol
```diff
@@ -153,28 +153,47 @@ abstract contract MultisigTask is Test, Script, ITask {
...
     /// @notice Executes the task with the given configuration file path and signatures.
     /// Sets the address registry, initializes and executes the task the single multisig
     /// as well as the nested multisig.
     /// @param taskConfigFilePath The path to the task configuration file.
     /// @param signatures The signatures to execute the task.
-    function executeRun(string memory taskConfigFilePath, bytes memory signatures) public {
-        _taskSetup(taskConfigFilePath);
-        /// now execute task actions
-        build();
+    function executeRun(string memory taskConfigFilePath, bytes memory signatures)
```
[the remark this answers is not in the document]
mds1: In general, the current `MultisigTask.sol` relies heavily on using global storage to pass input and outputs between methods. This effectively means we are stuck with `MultisigTask` as a big monolithic contract, and makes it hard to refactor things into modular libraries or unit test components. So my concern was that relying on storage too much will make things less maintainable, extensible, and flexible

The alternative is a more functional approach, which is why we now return account accesses. We can then, for example, implement a library that parses account accesses to do things [like this](https://github.com/solidity-labs-io/superchain-ops/pull/33/files#diff-b4c4de2f3fe743624e8cacac84e0a7e1d19bb54d696e1a5978e6cb643816dae0R66-R70). Potential implementation of such a library:

```solidity
library AccountAccessParser {
    function getWritesFor(VmSafe.AccountAccesses[] memory accesses, account who)
        internal
        returns (VmSafe.StorageAccesses[] memory)
    {
        // implement logic to find all storage writes for account `who`
    }
}

abstract contract OPCMBaseTask {
    using AccountAccessParser for VmSafe.AccountAccess[];

    function validate(VmSafe.AccountAccess[] memory accesses) public override {
        VmSafe.StorageAccesses[] memory storageWrites = accesses.getWritesFor(parentMultisig);
        require(storageWrites.length == 1, "OPCMBaseTask: only nonce should be updated on upgrade controller multisig");
    }
}
```

Technically this code will be less efficient because `getWritesFor` would loop over all accesses every time its called, but unless that function becomes a bottleneck (which it won't, because RPC calls will be the bottleneck), than that doesn't matter

mds1 (dismissed):

src/improvements/tasks/MultisigTask.sol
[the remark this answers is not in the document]
prat-gpt: Got it. Thanks.

prat-gpt:
Also closes issue https://github.com/ethereum-optimism/superchain-ops/issues/596 with merge of my PR https://github.com/solidity-labs-io/superchain-ops/pull/33 in this PR.

src/improvements/tasks/OPCMBaseTask.sol
```diff
@@ -57,10 +59,15 @@ abstract contract OPCMBaseTask is MultisigTask {
...
+    function validate(VmSafe.AccountAccess[] memory accesses) public override {
         (address[] memory targets,,) = getTaskActions();
         require(targets.length == 1 && targets[0] == opcm(), "OPCMBaseTask: only OPCM is allowed as target");
-        super.validate();
+        super.validate(accesses);
+        require(
+            _stateInfos[parentMultisig].length == 1,
+            "OPCMBaseTask: only nonce should be updated on upgrade controller multisig"
+        );
+        require(_stateInfos[opcm()].length == 0, "OPCMBaseTask: Storage writes are not allowed on OPCM");
```
mds1, line 70: Once https://github.com/ethereum-optimism/superchain-ops/pull/616 is merged we should use that library instead of relying on the stateful read of `_stateInfos` here

mds1 (approved):

PR #611, opened 2025-02-14
Adding Simulation of Stacked Tasks

Ethnical:
This PR simulated multiples tasks in a row within the same anvil fork.
Really useful for ensuring multiples upgrade in the same batch and avoid error that would make all the signers resign again. 

https://github.com/user-attachments/assets/a0ba8bf4-d5cd-48fc-96ba-ffe53154343f

To execute the script just need to do: 
```
 ./script/utils/sim-sequence.sh eth "021 022 base-003 ink-001"
```

and to kill anvil if needed because stuck. 
```
ps aux | grep anvil | grep -v grep | awk '{print $2}' | xargs kill; ./script/utils/sim-sequence.sh eth "021 022 base-003 ink-001"
```
## TESTS
Tests made are covering all the necessary case we need using this:
```
#   - expected to pass: ./script/utils/sim-sequence.sh eth "021 022 base-003 ink-001" ✅
#   - expected to pass: ./script/utils/sim-sequence.sh eth "021 022 base-003" ✅
#   - expected to fail: ./script/utils/sim-sequence.sh eth "021 base-003 ink-001 022" ✅
#   - expected to fail: ./script/utils/sim-sequence.sh eth "021 base-003 ink-001" ✅
#   - expected to fail: ./script/utils/sim-sequence.sh eth "021 025" ✅
#   - expected to fail: create a dir called 021-tmp and run ./script/utils/sim-sequence.sh eth "021 022 base-003 ink-001" ✅
```


tasks/eth/ink-001-permissionless-proofs/NestedSignFromJson.s.sol
```diff
@@ -9,7 +9,6 @@ import {stdToml} from "forge-std/StdToml.sol";
 import {Vm, VmSafe} from "forge-std/Vm.sol";
 import {GnosisSafe} from "safe-contracts/GnosisSafe.sol";
 import {LibString} from "solady/utils/LibString.sol";
-import {Types} from "@eth-optimism-bedrock/scripts/Types.sol";
```
Ethnical, line 12: I have to remove this because this is copy paste from previous tasks and causing issue since the lib the file is not here anymore

.circleci/config.yml
```diff
@@ -264,6 +264,23 @@ jobs:
...
+        type: string
+      block_number:
+        type: string
+    docker:
+      - image: <<pipeline.parameters.default_docker_image>>
+    steps:
+      - utils/checkout-with-mise
+      - run:
+          name: "Run simulation sequence for <<parameters.prefix>> tasks: <<parameters.tasks>>"
+          command: export FOUNDRY_DISABLE_NIGHTLY_WARNING=0; git submodule update; just clean; just install && echo "yes" | bash ./script/utils/sim-sequence.sh <<parameters.prefix>> "<<parameters.tasks>>" "<<parameters.block_number>>"
```
mds1, line 282: We now use mise and foundry v1 in this repo, so we should be able to remove `export FOUNDRY_DISABLE_NIGHTLY_WARNING=0`

.circleci/config.yml
```diff
@@ -264,6 +264,23 @@ jobs:
       - simulate_nested:
           task: "/sep/zora-002-fp-upgrade"
 
+  simulate_sequence:
+    description: "Runs a sequence of simulations"
+    parameters:
+      prefix:
+        type: string
```
mds1, line 271: Can we call this `network` or `network-directory` to be more clear?

.circleci/config.yml
```diff
@@ -264,6 +264,23 @@ jobs:
...
+      tasks:
+        type: string
+      block_number:
+        type: string
```
mds1, line 275: is `latest` supported, so we can continually run it against the latest block for in-progress tasks?

nested.just
```diff
@@ -12,6 +12,11 @@ export ownerSafe := env_var('OWNER_SAFE')
 export chainGovernorSafe := env_var_or_default('CHAIN_GOVERNOR_SAFE', '')
 export randomPersonEoa := "0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045"
 
+export PRIVATE_KEY_OWNER := "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80"
+export PRIVATE_KEY_EXECUTOR :="0x59c6995e998f97a5a0044966f0945389dc9e86dae88c7a8412f4603b6b78690d"
```
mds1, line 16: nit
```suggestion
# These are two of the anvil default private keys, they are publicly known
export PRIVATE_KEY_OWNER := "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80"
export PRIVATE_KEY_EXECUTOR :="0x59c6995e998f97a5a0044966f0945389dc9e86dae88c7a8412f4603b6b78690d"
```

nested.just
```diff
@@ -94,6 +99,124 @@ sign whichSafe hdPath='0':
...
+
+  if [[ -z "$hash" ]]; then
+      echo "Hash not found in the file."
+      exit 10
+  else
+      echo "Extracted Hash: $hash"
+  fi
+
+  ## 2. Approval of this hash in the SC and FND respectively.
+  cast send ${safe} "approveHash(bytes32)" ${hash} --rpc-url ${ANVIL_LOCALHOST_RPC} --private-key {{PRIVATE_KEY_OWNER}} 2>&1 /dev/null
```
mds1, line 182: Does this also support chain governor, for the 3/3 (e.g. Unichain) case? I think yes based on the code, but comments indicate otherwise, so just confirming

nested.just
```diff
@@ -94,6 +99,124 @@ sign whichSafe hdPath='0':
...
+
+  ## 3. Execute the Approval permissionless on the L1PAO will revert here if there is an error with the nonce.
+  forge script ${script} \
+    --rpc-url ${ANVIL_LOCALHOST_RPC} \
+    --broadcast \
+    --private-key {{PRIVATE_KEY_EXECUTOR}} \
+    --sig "approveJson(string,address,bytes)" \
+    ${bundlePath} \
+    ${safe} \
+    11111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111 2>&1 /dev/null
```
mds1, line 192: What is the string of 1's? Also the comment on L184 confuses me a bit, can we rephrase that to clarify?

nested.just
```diff
@@ -151,6 +273,8 @@ execute hdPath='0':
     --sig "runJson(string)" \
     ${bundlePath}
 
+
+
```
mds1, line 277: delete?
```suggestion
```

nested.just
```diff
@@ -94,6 +99,124 @@ sign whichSafe hdPath='0':
...
+  
+  forge build
+  outputforge=$(forge script ${script} \
+    --rpc-url ${ANVIL_LOCALHOST_RPC} \
+    --private-key {{PRIVATE_KEY_OWNER}} \
+    --sig "signJson(string,address)" \
+    ${bundlePath} \
+    ${safe})
+
+  echo "Outputforge: $outputforge" > /tmp/outputforge.txt
```
mds1, line 169: Would suggest using `mktmp` instead of hardcoding `/tmp`, since on macOS the `mktmp` uses a dir in `/var`

script/utils/sim-sequence.sh
```diff
@@ -0,0 +1,308 @@
+#!/bin/bash
+
+
+LOGFILE="/tmp/sim-sequence.log"
```
mds1, line 4: Same comment on `mktmp`, applies to a few places in this file

script/utils/sim-sequence.sh
```diff
@@ -0,0 +1,308 @@
...
+## Nonce Values
+MAX_NONCE_ERROR=9999999
+FUS_BEFORE=$MAX_NONCE_ERROR
+FOS_BEFORE=$MAX_NONCE_ERROR
+SC_BEFORE=$MAX_NONCE_ERROR
+L1PAO_BEFORE=$MAX_NONCE_ERROR
```
mds1, line 10: What do we initialize all nonces to 9999999?

script/utils/sim-sequence.sh
```diff
@@ -0,0 +1,308 @@
...
+#   ./script/utils/sim-sequence.sh eth "021 022 base-003 ink-001"
+
+# TODO start an anvil fork and run all simulations against the same node.
+# TODO verify the following test cases, push a commit verifying each one has the expected result in CI:
+#   - expected to pass: ./script/utils/sim-sequence.sh eth "021 022 base-003 ink-001"
+#   - expected to pass: ./script/utils/sim-sequence.sh eth "021 022 base-003"
+#   - expected to fail: ./script/utils/sim-sequence.sh eth "021 base-003 ink-001 022"
+#   - expected to fail: ./script/utils/sim-sequence.sh eth "021 base-003 ink-001"
+#   - expected to fail: ./script/utils/sim-sequence.sh eth "021 025"
+#   - expected to fail: create a dir called 021-tmp and run ./script/utils/sim-sequence.sh eth "021 022 base-003 ink-001"
```
mds1, line 68: Are these still TODOs?

script/utils/sim-sequence.sh
```diff
@@ -0,0 +1,308 @@
...
+set -euo pipefail
```
mds1, line 69: nit, can this line be right after the shebang? don't usually see it this low — or was this intentional for some reason?

script/utils/sim-sequence.sh
```diff
@@ -0,0 +1,308 @@
...
+  fi
+
+  # Kill the anvil fork at the end if it was started by this script
+  if $DESTROY_ANIVIL_AFTER_EXECUTION; then
+    echo "Anvil is still open. To kill it, please use the command below:"
+    echo "ps aux | grep anvil | grep -v grep | awk '{print \$3}' | xargs kill"
+    #TODO: In the future, we should kill the anvil fork here?
+  fi
+}
+createFork() {
```
mds1, line 111: nit
```suggestion
}

createFork() {
```

script/utils/sim-sequence.sh
```diff
@@ -0,0 +1,308 @@
...
+
+  if [[ -f "$backup_file" ]]; then
+    mv "$backup_file" "$simulation_file"
+  fi
+
+  # Kill the anvil fork at the end if it was started by this script
+  if $DESTROY_ANIVIL_AFTER_EXECUTION; then
+    echo "Anvil is still open. To kill it, please use the command below:"
+    echo "ps aux | grep anvil | grep -v grep | awk '{print \$3}' | xargs kill"
+    #TODO: In the future, we should kill the anvil fork here?
```
mds1, line 108: Still a TODO it seem? Feel free to create tracking issues for any of these future improvements

script/utils/sim-sequence.sh
```diff
@@ -0,0 +1,308 @@
...
+    if [[ -n "$block_number" ]]; then
+      anvil -f $RPC_URL --fork-block-number $block_number >> /tmp/anvil.logs &
+    else
+      anvil -f $RPC_URL >> /tmp/anvil.logs &
+    fi
+    sleep 5
+  fi
+}
+
+NonceDisplayModified(){
```
mds1, line 128: Can be done in a follow up PR, but would be nice to support the chain governor for 3/3 case here

mds1:
This is really awesome, great work!

.circleci/config.yml
Ethnical: Got it! 
I did this because I had warnings logs that was ugly into CI simulation. 
will do that now and see.

.circleci/config.yml
Ethnical: since `block_number` is optional, not setting it would be resulting to simulate against the latest block.

script/utils/sim-sequence.sh
Ethnical: will comment but because a safe can have the nonce=0 for example in sepolia or with a new safe. 
if the nonce is 0, we can't really know if this an error because the RPC is dead or retrived incorrect data.
but the if the nonce `9999999` is returned we no for sure that no value has been overwritten.

script/utils/sim-sequence.sh
Ethnical: Correct! will do that ;)

script/utils/sim-sequence.sh
Ethnical: Done here -> https://github.com/ethereum-optimism/superchain-ops/issues/646

script/utils/sim-sequence.sh
Ethnical: open a tracking issue here -> https://github.com/ethereum-optimism/superchain-ops/issues/647

nested.just
Ethnical: This is a dummy signature this is required since superchain-ops doesn't allow empty signature for certain action I replaced with {FAKE_SIG}
in commit 1d2ba459a79483845bc2523c601a1e5265d47bce

script/utils/sim-sequence.sh
Ethnical: Correct removed everything now

script/utils/sim-sequence.sh
Ethnical: applied in commit 05ac656de8969a7cdae7c05b879592f0f8da4914

nested.just
Ethnical: Thanks applied in 05ac656de8969a7cdae7c05b879592f0f8da4914

nested.just
Ethnical: this is supporting any safes. However, for the 3/3 will require a bit of rework for now since we need to approve his `approvedHash` also.

.circleci/config.yml
mds1: Ok perfect, thank you! So for tests we should specify a block number like you did, and in practice when we go to use this we will want to use latest

nested.just
mds1: Got it, and then this comment:

> Execute the Approval permissionless on the L1PAO will revert here if there is an error with the nonce.

Can we rephrase that? I think it is saying "Execute the approveHash on the L1PAO. This will revert if there is an error with the nonce." — the "permissionless" part is just throwing me off 😅

Ethnical: Ohh, got it! This is insisting on the fact that we execute approval from an external EOA `PRIVATE_KEY_EXECUTOR` that is not the owner of the safe!
Otherwise, we got through an error that we had to debug and found with kelvin later last week link on slack [here](https://oplabs-pbc.slack.com/archives/C07QHCG58AK/p1739450684479809?thread_ts=1739301675.627029&cid=C07QHCG58AK).
I will make this clearer here!

script/utils/sim-sequence.sh
Ethnical: Whoops forgot to remove this comment correct!

nested.just
Ethnical: fixed in this commit: 1b5f047e7755775581a646cd008eb447b8f902bb

mds1: thank you!

mds1 (approved):

PR #616, opened 2025-02-17
refactor: add `AccountAccessParser` library

mds1:
Merges the existing state processing libraries into a single `AccountAccessParser.sol` library, with corresponding tests. This is not yet used in `MultisigTask.sol` because it is stateful, so the `print()` method does not have access to the `accountAccesses` array. Once that method does have access, we delete the corresponding code and simply call `accountAccesses.decodeAndPrint()`

Example usage:
```solidity
contract MyContract {
    using AccountAccessParser for VmSafe.AccountAccess[];

    function myFunction(VmSafe.AccountAccess[] memory accountAccesses) public {
        // Decode all state changes and ETH/ERC20 transfers and print to the terminal.
        accountAccesses.decodeAndPrint();

        // Get all decoded data without printing.
        (
            AccountAccessParser.DecodedTransfer[] memory transfers,
            AccountAccessParser.DecodedStateDiff[] memory diffs
        ) = accountAccesses.decode();

        // Get the state diff for a given account.
        StateDiff[] memory diffs = accountAccesses.getStateDiffFor(myContract);

        // Get an array of all unique accounts that had state changes.
        address[] memory accountsWithStateChanges = accountAccesses.getUniqueWrites();
    }
}
```


src/libraries/AccountAccessParser.sol
```diff
@@ -77,86 +115,207 @@ library StateDiffDecoder {
...
+    // ==============================================================
+
+    /// @notice Convenience function that wraps decode and print together.
+    function decodeAndPrint(VmSafe.AccountAccess[] memory _accesses) internal view {
+        (DecodedTransfer[] memory transfers, DecodedStateDiff[] memory stateDiffs) = decode(_accesses);
+        print(transfers, stateDiffs);
+    }
+
+    /// @notice Decodes the provided AccountAccess array into decoded transfers and state diffs.
+    function decode(VmSafe.AccountAccess[] memory _accountAccesses)
```
mds1, line 140: note to self: make sure to handle the `reverted` field properly for both account and storage accesses

src/libraries/AccountAccessParser.sol
```diff
@@ -35,6 +64,13 @@ library StateDiffDecoder {
         DecodedSlot decoded;
     }
 
+    struct DecodedTransfer {
+        address from;
+        address to;
+        uint256 value;
+        address tokenAddress;
```
maurelian, line 71: So, this is an erc20 transfer? Should that be reflected in the struct name? 

Is the idea that. We would add other structs for other high level interaction types?

mds1: It can also an be an ETH transfer if the token address is the `0xeee` address, which is the more likely use case for us (some FP recovery runbooks have ETH transfers)

Definitely open to adding structs for other high level interactions — any particular ones you'd want to add?

maurelian: I'm still just trying to tease out the motivation. AFAI recall, most of our ops tasks would have no eth or erc20 transfers?

I would imagine we'd want to add new logic to identify and return different decodings by recognizing addresses from the superchain registry and then mapping slots to var names based on the storage layout?

mds1: > AFAI recall, most of our ops tasks would have no eth or erc20 transfers?

You're right that most tasks won't use the token transfer decoding, though ETH transfers do happen as part of FP [incident response](https://github.com/ethereum-optimism/optimism/blob/074c2231c3e2f501833bb519e5102ed009c99349/packages/contracts-bedrock/src/dispute/DelayedWETH.sol#L118). Maybe we start doing some `Governor` related tasks here that move OP tokens, so figured I'd include the ERC20 decoding as well since it's easy to add.

> I would imagine we'd want to add new logic to identify and return different decodings by recognizing addresses from the superchain registry and then mapping slots to var names based on the storage layout?

This logic is present! See here: https://github.com/ethereum-optimism/superchain-ops/blob/924fc50797b1e659009fa5881e6015fab8772f7d/src/libraries/AccountAccessParser.sol#L387-L400

And we get contract name from an address here: https://github.com/ethereum-optimism/superchain-ops/blob/924fc50797b1e659009fa5881e6015fab8772f7d/src/libraries/AccountAccessParser.sol#L642-L694

> I'm still just trying to tease out the motivation. 

The main motivation for this is here: https://github.com/ethereum-optimism/superchain-ops/pull/609#discussion_r1958638853. Basically `MultisigTask.sol` is a monolith and IMO because it relies on storage to pass things around, it's hard to test and follow, so I wanted to modularize it. Pulling out all state diff functionality is the most obvious improvement there, so that's what this PR does

maurelian: OK, I see it now.  I'll leave it for others to approve, but this is all very awesome.

blmalone (approved):
this is awesome, great UX

PR #618, opened 2025-02-18
feat: minimal changes for foundry 1.0.0

alcueca:
**Description**

Foundry 1.0.0 is out, and this repo is an easier place to test the upgrade than the monorepo.

To allow the existing tests to pass, it was chosen to allow `expectRevert` to catch reverts from internal functions, instead of the more convoluted route of creating a helper contract. From [here](https://book.getfoundry.sh/guides/v1.0-migration#expect-revert-cheatcode-disabled-on-internal-calls-by-default).

**Tests**

Compared to `nightly-5d16800a64e5357fbb2493e4cae061756d145981` there are no significant improvements on build (3.92s for 1.0.0 vs 3.88s for 0.3.1) or test time (21.35s vs 21.59s).

### DIFF
--- a/test/libraries/BytecodeComparison.t.sol
+++ b/test/libraries/BytecodeComparison.t.sol
@@ -15,6 +15,25 @@ contract MockContract {
     }
 }
 
+/// @notice Library to expose the internal functions, to avoid tests stopping after hitting an expected revert.
+/// https://book.getfoundry.sh/cheatcodes/expect-revert
+library BytecodeComparisonHarness {
+
+    function compare(address _contractA, address _contractB, BytecodeComparison.Diff[] memory _allowed)
+        external view
+        returns (bool)
+    {
+        return BytecodeComparison.compare(_contractA, _contractB, _allowed);
+    }
+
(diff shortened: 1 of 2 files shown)


test/libraries/BytecodeComparison.t.sol
```diff
@@ -32,6 +32,7 @@ contract BytecodeComparison_compare_Test is Test {
     }
 
     /// @notice Test that different contracts with different bytecode lengths revert.
+    /// forge-config: default.allow_internal_expect_revert = true
```
mds1, line 35: I do think we should use the convoluted route because it's safer — if you use this approach, the test stops executing after the first `expectRevert` catches an internal (top-level) revert (at least, it did before v1, I have not checked if that changed)

You can see an example of how I normally do the proper revert test here: https://github.com/ethereum-optimism/superchain-ops/blob/main/test/ComputeSafeOwnerSwapHash.t.sol

blmalone (approved):

mds1 (approved):
Thank you!

PR #619, opened 2025-02-18
Create new template cleanup

blmalone:
Revert on default template that is created via `just new template`. Also, removing a bunch of unused imports.

### DIFF
--- a/src/improvements/script/create-template.sh
+++ b/src/improvements/script/create-template.sh
@@ -17,7 +17,6 @@ create_template() {
 // SPDX-License-Identifier: MIT
 pragma solidity 0.8.15;
 import {MultisigTask} from "src/improvements/tasks/MultisigTask.sol";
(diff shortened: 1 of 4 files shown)


src/improvements/script/create-template.sh
```diff
@@ -35,35 +34,35 @@ contract ${contract_name} is MultisigTask {
     /// @notice Returns the safe address string identifier.
     /// @return A string identifier.
     function safeAddressString() public pure override returns (string memory) {
-        // TODO: Return the actual safe address string identifier as defined in
+        require(false, "TODO: Return the actual safe address string identifier as defined in Superchain-Registry's addresses.json.");
         /// Superchain-Registry's addresses.json.
```
mds1, line 38: nit
```suggestion
        // Superchain-Registry's addresses.json.
```

mds1 (approved):
We should maintain this solidity template in a `*.template.sol` file so it's easier to maintain then in a shell script

PR #620, opened 2025-02-18
Extract empty template into its own  .template.sol file

blmalone:
Instead of doing string interpolation and not getting compiler errors. We can add the Solidity code into its own file and reference that way.


src/improvements/template/EmptyTemplate.template.sol
```diff
@@ -0,0 +1,55 @@
...
+        // return storageWrites;
+    }
+
+    /// @notice Sets up the template using configuration data from a file.
+    /// @param taskConfigFilePath The path to the configuration file.
+    function _templateSetup(string memory taskConfigFilePath) internal override {
+        require(false, "TODO: Implement the logic to parse the configuration file and populate the `taskConfig` mapping.");
+    }
+
+    /// @notice Implement one of _buildPerChain() or _buildSingle()
```
blmalone, line 40: Removed _build() in favor of _buildSingle and _buildPerChain

mds1 (dismissed):

mds1 (approved):

PR #622, opened 2025-02-19
Multisig Cleanup

ElliotFriedman:
This PR rounds out rough edges and makes MultisigTask more legible and play nicely with the scripting which broadcasts transactions.

- add helper function `execTransaction` to avoid code duplication calling execTransaction on gnosis safe
- inline `_setIsNestedSafe` logic
- print functions only gets data to sign on Gnosis Safe owners
- `executeRun` now only calls execute instead of `simulateRun`
- calls to approve child multisig hash now are broadcast if running as a forge script and the `--broadcast` flag is present

### DIFF
--- a/src/improvements/AddressRegistry.sol
+++ b/src/improvements/AddressRegistry.sol
@@ -62,6 +62,15 @@ contract AddressRegistry is Test {
         ChainInfo chainInfo;
     }
 
+    /// @notice Structure used to read in the following hardcoded addresses:
+    ///    - Foundation Upgrade Safe
+    ///    - Foundation Operation Safe
+    ///    - Security Council
+    struct HardcodedAddress {
+        address addr;
+        string identifier;
+    }
+
     /// @notice Maps an identifier and l2 instance chain ID to a stored address entry.
     /// All addresses will live on the same chain.
     mapping(string => mapping(uint256 => RegistryEntry)) private registry;
@@ -74,6 +83,16 @@ contract AddressRegistry is Test {
 
     /// @notice Array of supported chains and their configurations
     ChainInfo[] public chains;
+
+    /// @notice loads hardcoded foundation and security council addresses from
+    /// addresses.toml file
+    function _loadHardcodedAddresses(string memory chainKey, ChainInfo memory chain) internal {
+        HardcodedAddress[] memory hardcodedAddresses =
+            abi.decode(vm.parseToml(vm.readFile("./src/improvements/addresses.toml"), chainKey), (HardcodedAddress[]));
+        for (uint256 i = 0; i < hardcodedAddresses.length; i++) {
+            saveAddress(hardcodedAddresses[i].identifier, chain, hardcodedAddresses[i].addr);
+        }
+    }
 
     /// @notice Initializes the contract by loading addresses from TOML files
     /// and configuring the supported L2 chains.
@@ -111,6 +130,10 @@ contract AddressRegistry is Test {
             supportedL2ChainIds[chains[i].chainId] = true;
 
             _processAddresses(chains[i], chainAddressesContent);
+
+            _loadHardcodedAddresses(
+                block.chainid == getChain("mainnet").chainId ? ".mainnetAddresses" : ".testnetAddresses", chains[i]
+            );
         }
     }
 
--- a/src/improvements/addresses.toml
+++ b/src/improvements/addresses.toml
@@ -1,0 +1,24 @@
+[[mainnetAddresses]]
+identifier = "FoundationUpgradeSafe"
+addr = "0x847B5c174615B1B7fDF770882256e2D3E95b9D92"
+
+[[mainnetAddresses]]
+identifier = "FoundationOperationSafe"
+addr = "0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A"
+
+[[mainnetAddresses]]
+identifier = "SecurityCouncil"
+addr = "0xc2819DC788505Aac350142A7A707BF9D03E3Bd03"
+
+
+[[testnetAddresses]]
+identifier = "FoundationUpgradeSafe"
+addr = "0xDEe57160aAfCF04c34C887B5962D0a69676d3C8B"
+
+[[testnetAddresses]]
+identifier = "FoundationOperationSafe"
+addr = "0x837DE453AD5F21E89771e3c06239d8236c0EFd5E"
+
+[[testnetAddresses]]
+identifier = "SecurityCouncil"
+addr = "0xf64bc17485f0B4Ea5F06A96514182FC4cB561977"
--- a/test/registry/TestnetAddressRegistry.t.sol
+++ b/test/registry/TestnetAddressRegistry.t.sol
@@ -4,33 +4,24 @@ pragma solidity 0.8.15;
 import {Test} from "forge-std/Test.sol";
 import {AddressRegistry} from "src/improvements/AddressRegistry.sol";
 
-contract MainnetAddressRegistryTest is Test {
+contract TestnetAddressRegistryTest is Test {
     AddressRegistry private addresses;
 
-    uint256 public metalChainId;
-    uint256 public baseChainId;
-    uint256 public opMainnetChainId;
-    uint256 public zoraChainId;
-    uint256 public modeChainId;
+    uint256 public opSepoliaChainId;
+    uint256 public metalSepoliaChainId;
 
     function setUp() public {
-        string memory networkConfigFilePath = "test/tasks/mock/configs/DiscoverChainAddressesConfig.toml";
+        string memory networkConfigFilePath = "test/tasks/mock/configs/DiscoverChainAddressesTestnetConfig.toml";
 
-        vm.createSelectFork("mainnet");
+        vm.createSelectFork("sepolia");
 
         addresses = new AddressRegistry(networkConfigFilePath);
-        metalChainId = getChain("metal").chainId;
-        baseChainId = getChain("base").chainId;
-        opMainnetChainId = getChain("optimism").chainId;
-        zoraChainId = getChain("zora").chainId;
-        modeChainId = getChain("mode").chainId;
+        opSepoliaChainId = getChain("optimism_sepolia").chainId;
+        metalSepoliaChainId = getChain("metal_sepolia").chainId;
     }
 
     function testContractState() public view {
-        assertTrue(addresses.supportedL2ChainIds(metalChainId), "Metal chain ID not supported");
-        assertTrue(addresses.supportedL2ChainIds(baseChainId), "Base chain ID not supported");
-        assertTrue(addresses.supportedL2ChainIds(opMainnetChainId), "OP Mainnet chain ID not supported");
-        assertTrue(addresses.supportedL2ChainIds(zoraChainId), "Zora chain ID not supported");
+        assertTrue(addresses.supportedL2ChainIds(opSepoliaChainId), "Op Sepolia chain ID not supported");
     }
 
     function testSuperchainAddressesLoaded() public view {
@@ -180,17 +171,17 @@ contract MainnetAddressRegistryTest is Test {
 
             assertEq(
                 addresses.getAddress("FoundationUpgradeSafe", chainId),
-                0x847B5c174615B1B7fDF770882256e2D3E95b9D92,
+                0xDEe57160aAfCF04c34C887B5962D0a69676d3C8B,
                 "FoundationUpgradeSafe not properly loaded"
             );
             assertEq(
                 addresses.getAddress("FoundationOperationSafe", chainId),
-                0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A,
+                0x837DE453AD5F21E89771e3c06239d8236c0EFd5E,
                 "FoundationOperationSafe not properly loaded"
             );
             assertEq(
                 addresses.getAddress("SecurityCouncil", chainId),
-                0xc2819DC788505Aac350142A7A707BF9D03E3Bd03,
+                0xf64bc17485f0B4Ea5F06A96514182FC4cB561977,
                 "SecurityCouncil not properly loaded"
             );
         }
@@ -203,7 +194,7 @@ contract MainnetAddressRegistryTest is Test {
 
     function testGetNonExistentAddressFails() public {
         vm.expectRevert("Address not found");
-        addresses.getAddress("NON_EXISTENT_ADDRESS", opMainnetChainId);
+        addresses.getAddress("NON_EXISTENT_ADDRESS", opSepoliaChainId);
     }
 
     function testGetNonExistentAddressInfoFails() public {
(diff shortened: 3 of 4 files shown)


src/improvements/tasks/MultisigTask.sol
```diff
@@ -187,9 +187,20 @@ abstract contract MultisigTask is Test, Script {
...
+        build();
+
+        /// now execute task actions
+        VmSafe.AccountAccess[] memory accountAccesses = execute(signatures);
+
+        /// validate all state transitions
+        validate(accountAccesses);
+
+        /// print out results of execution
+        print();
```
mds1, line 203: What is the motivation for this change? I think it's ok, just want to understand

src/improvements/tasks/MultisigTask.sol
```diff
@@ -205,6 +216,11 @@ abstract contract MultisigTask is Test, Script {
         _taskSetup(taskConfigFilePath);
         build();
         approve(_childMultisig, signatures);
+        console.log(
+            "--------- Successfully %s Child Multisig %s Approval ---------",
+            vm.isContext(VmSafe.ForgeContext.ScriptBroadcast) ? "Broadcasted" : "Simulated",
```
mds1, line 221: I believe `ScriptResume` is also needed here?
```suggestion
            vm.isContext(VmSafe.ForgeContext.ScriptBroadcast || VmSafe.ForgeContext.ScriptResume) ? "Broadcasted" : "Simulated",
```

src/improvements/tasks/MultisigTask.sol
```diff
@@ -254,7 +270,18 @@ abstract contract MultisigTask is Test, Script {
...
+            uint256 quorum = IGnosisSafe(parentMultisig).getThreshold();
+            uint256 eoaOwnerCount;
+            for (uint256 i = 0; i < owners.length; i++) {
+                if (owners[i].code.length == 0) {
+                    eoaOwnerCount++;
+                }
+            }
+
+            isNestedSafe = eoaOwnerCount >= quorum ? false : true;
+        }
```
mds1, line 284: Let's retain the old logic here of "if any owner on the safe is an EOA, it is not nested". This definition is a bit arbitrary and doesn't really benefit us

src/improvements/tasks/MultisigTask.sol
```diff
@@ -392,33 +415,51 @@ abstract contract MultisigTask is Test, Script {
...
+        address target,
+        uint256 value,
+        bytes memory data,
+        Enum.Operation operationType,
+        bytes memory signatures
+    ) internal {
+        (bool success) = IGnosisSafe(multisig).execTransaction(
+            target, value, data, operationType, 0, 0, 0, address(0), payable(address(0)), signatures
         );
         require(success, "MultisigTask: execute failed");
```
mds1, line 464: We should propagate the revert reason, which for a safe is typically a string, which is why the cast from bytes to string is ok
```suggestion
        (bool success, bytes memory returnData) = IGnosisSafe(multisig).execTransaction(
            target, value, data, operationType, 0, 0, 0, address(0), payable(address(0)), signatures
        );
        require(success, string.concat("MultisigTask: execute failed with error", string(returnData)));
```

src/improvements/tasks/MultisigTask.sol
```diff
@@ -392,33 +415,51 @@ abstract contract MultisigTask is Test, Script {
...
-            approveCalldata,
-            Enum.Operation.DelegateCall,
-            0,
-            0,
-            0,
-            address(0),
-            payable(address(0)),
-            signatures
-        );
+        if (vm.isContext(VmSafe.ForgeContext.ScriptBroadcast)) {
```
mds1, line 418: Similar comment about needing `ScriptResume` here — we should extract this check into:

```solidity
function isBroadcastContext() internal view returns (bool) {
  return vm.isContext(VmSafe.ForgeContext.ScriptBroadcast) || vm.isContext(VmSafe.ForgeContext.ScriptResume);
}

File: src/improvements/tasks/MultisigTask.sol [hunk 6]
```diff
@@ -392,33 +415,51 @@ abstract contract MultisigTask is Test, Script {
...
+            /// only broadcast if we are running forge script with the --broadcast flag
```
mds1, line 419: nit, can't use natspec in a function so should only be 2 slashes
```suggestion
            // only broadcast if we are running forge script with the --broadcast flag
```

src/improvements/tasks/MultisigTask.sol
```diff
@@ -392,33 +415,51 @@ abstract contract MultisigTask is Test, Script {
...
 
     /// @notice Executes the task with the given signatures.
     /// @param signatures The signatures to execute the task.
-    function execute(bytes memory signatures) public {
+    function execute(bytes memory signatures) public returns (VmSafe.AccountAccess[] memory) {
         bytes memory data = getCalldata();
         bytes32 hash = getHash();
 
-        signatures = Signatures.prepareSignatures(parentMultisig, hash, signatures);
+        /// if no signatures are attached, create them
```
mds1, line 432: same natspec comment, can be done in a follow up PR to fix all these. Natspec is only supported for contracts, interfaces, errors, events, functions, but not for arbitrary code blocks: https://docs.soliditylang.org/en/latest/natspec-format.html
```suggestion
        // if no signatures are attached, create them
```

src/improvements/tasks/MultisigTask.sol
```diff
@@ -392,33 +415,51 @@ abstract contract MultisigTask is Test, Script {
...
+        if (signatures.length == 0) {
+            signatures = prepareSignatures(parentMultisig, hash);
+        } else {
+            /// otherwise we assume the signatures are already ordered
+            Signatures.sortUniqueSignatures(
+                parentMultisig, signatures, hash, IGnosisSafe(parentMultisig).getThreshold(), signatures.length
+            );
+        }
```
mds1, line 440: Can you clarify the if/else block here in the code comments? I don't understand why "creating signatures" means? Also the `else` block says "we assume the signatures are already ordered" but then proceeds to sort the signatures anyway

src/improvements/tasks/MultisigTask.sol
```diff
@@ -392,33 +415,51 @@ abstract contract MultisigTask is Test, Script {
...
+
+        vm.startStateDiffRecording();
+        if (vm.isContext(VmSafe.ForgeContext.ScriptBroadcast)) {
+            /// only broadcast if we are running forge script with the --broadcast flag
+            vm.broadcast();
```
mds1, line 445: Seems this `vm.broadcast()` call with the corresponding `if (isBroadcastContext())` block and state diff recording should be moved inline to the `execTransaction` method? This way:
- the broadcast invocation is right next to the call being broadcast, so if `execTransaction` gets modified for whatever reason we don't accidentally broadcast a second transaction
- we deduplcate all the `isBroadcastContext()` checks
- `execTransaction` now returns the account accesses and is a nicely encapsulated method handling all execution considerations

src/improvements/tasks/MultisigTask.sol
```diff
@@ -612,14 +653,17 @@ abstract contract MultisigTask is Test, Script {
     function printNestedDataToSign() public view {
         bytes memory callData = generateApproveMulticallData();
 
-        if (childMultisig != address(0)) {
+        if (childMultisig != address(0) && childMultisig.code.length != 0) {
```
mds1, line 656: Should this be replaced with `if (isNestedSafe)`? If not, what's the difference between that bool and this check? Code comments explaining the "why" behind logic would be helpful here

src/improvements/tasks/MultisigTask.sol
```diff
@@ -612,14 +653,17 @@ abstract contract MultisigTask is Test, Script {
...
             console.log("vvvvvvvv");
             console.logBytes(getDataToSign(childMultisig, callData));
             console.log("^^^^^^^^\n");
         } else {
             for (uint256 i; i < startingOwners.length; i++) {
+                if (startingOwners[i].code.length == 0) {
+                    continue;
+                }
                 console.log("Nested multisig: %s", getAddressLabel(startingOwners[i]));
                 console.logBytes(getDataToSign(startingOwners[i], callData));
```
mds1, line 668: I don't understand what this block's logic well either, is this printing data to sign for every owner on the safe?

src/improvements/tasks/MultisigTask.sol
```diff
@@ -630,11 +674,16 @@ abstract contract MultisigTask is Test, Script {
     function printNestedHashToApprove() public view {
         bytes memory callData = generateApproveMulticallData();
 
-        if (childMultisig != address(0)) {
+        if (childMultisig != address(0) && childMultisig.code.length != 0) {
```
mds1, line 677: Similar comments for this method. I would just create a single `printNestedData()` method to hold all of:
```solidity
if (isNestedSafe) {
    console.log("\n\n------------------ Nested Multisig EOAs Data to Sign ------------------");
    printNestedDataToSign();
    console.log("\n\n------------------ Nested Multisig EOAs Hash to Approve ------------------");
    printNestedHashToApprove();
```
that logic. Because both `printNestedDataToSign` and `printNestedHashToApprove` are only called once anyway, so this way we can clearly document all the logs and logic for the nested case.

`printDataToSign` and `printHashToApprove` are also both only used once, so we can again bundle the logs and logic for the single case in the `else` block

test/tasks/NestedMultisigTask.t.sol
```diff
@@ -301,7 +319,17 @@ contract NestedMultisigTaskTest is Test {
...
+        /// snapshot before running the task so we can roll back to this pre-state
+        uint256 newSnapshot = vm.snapshot();
+
         multisigTask.simulateRun(opcmTaskConfigFilePath);
+        bytes32 taskHash = multisigTask.getHash();
+
+        /// now run the executeRun flow
+        vm.revertTo(newSnapshot);
+
+        multisigTask.executeRun(opcmTaskConfigFilePath, prepareSignatures(multisig, taskHash));
```
mds1, line 332: What does snapshot and reverting do for us here? It seems we aren't asserting that anything is equal between the `simulateRun` and `executeRun` flows?

prat-gpt (approved):
lgtm!

src/improvements/tasks/MultisigTask.sol
mds1: Got it. Are each of these bugs caught by one of the tests added in this PR? If not let's make sure to add that test before merging

src/improvements/tasks/MultisigTask.sol
mds1: I realized I misread here, this is not a low-level call so there is no return data to propagate

src/improvements/tasks/MultisigTask.sol
```diff
@@ -392,34 +414,72 @@ abstract contract MultisigTask is Test, Script {
...
+        bool success = false;
+
+        try IGnosisSafe(multisig).execTransaction(
+            target, value, data, operationType, 0, 0, 0, address(0), payable(address(0)), signatures
+        ) returns (bool execStatus) {
+            success = execStatus;
+        } catch (bytes memory err) {
+            console.log("Error executing multisig transaction");
+            console.logBytes(err);
+        }
```
mds1, line 481: I'm pretty sure this extra `catch` block can never be entered because this is a high level call that only returns a bool, right? So we should switch back to the original approach you had, sorry. See https://github.com/ethereum-optimism/superchain-ops/pull/622#discussion_r1961775502

src/improvements/tasks/MultisigTask.sol
mds1: > if `isNestedSafe` is true, that does not imply that `childMultisig` address is non zero. if `childMultisig` address is non zero, that does not imply anything about whether or not this is a nested safe, as a child multisig can have other Gnosis Safe signers.

Why do we care if a child multisig can have other safe signers? More generically, if took the implementer of this code running tests to realize that `childMultisig != address(0)` is not equivalent to `isNestedSafe`, the code is probably confusing and we should make sure to add code comments or helper methods to make things more clear :)

src/improvements/tasks/MultisigTask.sol
mds1: Is this a change from the old UX then, i.e. why does it not only print calldata for the child safe we are currently simulating or executing as (i.e. council, foundation, or chainGovernor)?

test/tasks/NestedMultisigTask.t.sol
mds1: > We are testing the `executeRun` method by running it.

So we are just testing that it does not revert? Do we have stronger assertions somewhere, e.g. that the `accountAccesses` returned by `simulateRun` are equivalent to those returned by `executeRun` (with the exception of potential differences resulting from e.g. state overrides)

src/improvements/tasks/MultisigTask.sol
prat-gpt: No, it is not a change from the old UX. If you take a look at the new `nested.just` file. for `simulate` and `sign` recipes `signFromChildMultisig` is called. In this function the child multisig address is passed and so only calldata for the passed child safe is printed.
We are printing data to sign for every owner in case when `simulateRun` is called for a nested multisig and the child multisig is not known.

src/improvements/tasks/MultisigTask.sol
```diff
@@ -630,11 +717,19 @@ abstract contract MultisigTask is Test, Script {
     function printNestedHashToApprove() public view {
         bytes memory callData = generateApproveMulticallData();
 
+        // this branch means the function `signFromChildMultisig` is being called
```
blmalone, line 720: I think we will have to revisit the general architecture of MultisigTask.sol. This comment is good in the meantime, but if we move to a more functional style and away from having to keep track of global state variable changes, I think we will gain a lot more readability.

blmalone (approved):

src/improvements/tasks/MultisigTask.sol
blmalone: Going to resolve this comment. Based on Elliots testing above.

src/improvements/tasks/MultisigTask.sol
blmalone: Okay, as I said in another comment. I think this is fine for now but having to have comments like:

> this branch means function `signFromChildMultisig` is not being called

Usually suggests that the code is hard to follow. I think the readability of this file is key. We should have a followup task for an architecture refactor. Resolving this comment for now.

test/tasks/NestedMultisigTask.t.sol
mds1: Created https://github.com/ethereum-optimism/superchain-ops/issues/637 and moved to the bottom of https://github.com/ethereum-optimism/superchain-ops/issues/396, @blmalone feel free to reorganize where that new issue lives in the tracker

src/improvements/tasks/MultisigTask.sol
mds1: Tracking in https://github.com/ethereum-optimism/superchain-ops/issues/636, so resolving this

blmalone (approved):

src/improvements/tasks/MultisigTask.sol
mds1: got it, that makes sense. thanks!

Issue #621, opened 2025-02-18
Rewrite first sepolia execution task as template


blmalone:
Rewrite task as template: https://github.com/ethereum-optimism/superchain-ops/tree/main/tasks/sep/029-deputy-pause-module

PR #624, opened 2025-02-19
Enable Module Template

ElliotFriedman:
Create a new template to enable modules.

Not currently ready for review.

Should close https://github.com/ethereum-optimism/superchain-ops/issues/621


src/improvements/template/EnableDeputyPauseModuleTemplate.sol
```diff
@@ -0,0 +1,188 @@
...
+
+        string[] memory storageWrites;
+
+        if (CHAIN_ID == MAINNET_CHAIN_ID) {
+            storageWrites = new string[](2);
+            storageWrites[1] = "LivenessGuard";
+        } else {
+            // otherwise, we only need to write to the Safe
+            storageWrites = new string[](1);
+        }
```
mds1, line 72: Sepolia also has the Liveness Guard
```suggestion
        if (CHAIN_ID == getChain("mainnet").chainId || getChain("sepolia").chainId) {
            storageWrites = new string[](2);
            storageWrites[1] = "LivenessGuard";
        } else {
            // otherwise, we only need to write to the Safe
            storageWrites = new string[](1);
        }
```

src/improvements/template/EnableDeputyPauseModuleTemplate.sol
```diff
@@ -0,0 +1,188 @@
...
+
+        storageWrites[0] = _SAFE_ADDRESS;
```
mds1, line 74: readability nit, seems weird to write array index 0 after 1, I'd move this line to be right after we declare `string[] memory storageWrites;`

src/improvements/template/EnableDeputyPauseModuleTemplate.sol
```diff
@@ -0,0 +1,188 @@
...
+
+        return storageWrites;
+    }
+
+    /// @notice Sets up the template with module configuration from a TOML file
+    /// @param taskConfigFilePath Path to the TOML configuration file
+    function _templateSetup(string memory taskConfigFilePath) internal override {
+        string memory file = vm.readFile(taskConfigFilePath);
+        newModule = vm.parseTomlAddress(file, ".newModule");
+        assertNotEq(newModule.code.length, 0, "new module must have code");
```
mds1, line 84: We can strength the pre-check assertions here to make sure the deuputy pause module is configured correctly, instead of just checking code:

```solidity
require(
  IDeputyPauseModule(IDeputyPauseModule(newModule).foundationSafe() == addrRegistry.getAddress(safeAddressString),
  "foundation safe mismatch"
);
require(
  IDeputyPauseModule(IDeputyPauseModule(newModule).superchainConfig() == addrRegistry.getAddress("SuperchainConfig"),
  "SuperchainConfig mismatch"
);
```

src/improvements/template/EnableDeputyPauseModuleTemplate.sol
```diff
@@ -0,0 +1,188 @@
...
+
+        // only allow one chain to be modified at a time with this template
+        AddressRegistry.ChainInfo[] memory _chains =
+            abi.decode(vm.parseToml(vm.readFile(taskConfigFilePath), ".l2chains"), (AddressRegistry.ChainInfo[]));
```
mds1, line 88: This is a "superchain level" task, i.e. we are applying it directly to a safe, not to protocol contracts. So I don't think this is the right framing or config file structure (i.e. there are no L2 chains to specify here)

src/improvements/template/EnableDeputyPauseModuleTemplate.sol
mds1: Ok, I see you have these in the validate method. Putting them here technically makes more sense to me because we are verifying the module is configured correctly before installing it, but that doesn't matter much in this case

src/improvements/template/EnableDeputyPauseModuleTemplate.sol
```diff
@@ -0,0 +1,188 @@
...
+                            sentinelModuleValue,
+                            bytes32(uint256(uint160(newModule))),
+                            "sentinel does not point to new module"
+                        );
+
+                        moduleWriteFound = true;
+                    }
+                }
+            }
+        }
```
mds1, line 176: A lot of this can replaced with the account access parser library! Specifically it's `getStateDiffFor` method. Also, this should logic should just be in `_validate` — `checkStateDiff` is intended to have a default implementation that is never changed, but I see `_validate` does not yet take account accesses as an input

src/improvements/template/EnableDeputyPauseModuleTemplate.sol
```diff
@@ -0,0 +1,188 @@
...
+        assertTrue(moduleWriteFound, "Module write not found");
+    }
+}
+
+interface IDeputyGuardianModuleFetcher {
+    function deputyGuardian() external view returns (address);
+    function foundationSafe() external view returns (address);
+    function superchainConfig() external view returns (address);
+    function version() external view returns (string memory);
+}
```
mds1, line 188: We can delete this and just import the actual interface from `lib/optimism/packages/contracts-bedrock/interfaces/safe/IDeputyPauseModule.sol`

test/tasks/Regression.t.sol
```diff
@@ -110,6 +111,24 @@ contract RegressionTest is Test {
         assertEq(keccak256(bytes(dataToSign)), keccak256(bytes(expectedDataToSign)));
     }
 
+    function testRegressionCallDataMatches_EnableDeputyPauseModuleTemplate() public {
```
mds1, line 114: Is the developer UX that part of template creation requires adding a corresponding regression test? That seems reasonable to me if so, but just want to confirm

src/improvements/template/EnableDeputyPauseModuleTemplate.sol
mds1: Ah right the DPM goes on a foundation safe, which does not have the liveness guard on both mainnet and sepolia. So I think we can delete this altogether?

src/improvements/template/EnableDeputyPauseModuleTemplate.sol
mds1: yea the alternative is setting the length before the if based on the conditional you already have, but doesn't matter, so resolving

src/improvements/template/EnableDeputyPauseModuleTemplate.sol
mds1: So the guidance for devs is: If generating a generic task that doesn't apply so a specific chain, just use the L1 chain ID? Does that have any consequences or side effects? Let's make sure to include this in the docs

src/improvements/template/EnableDeputyPauseModuleTemplate.sol
mds1: > should `_validate` take an account accesses parameter?

Yes! That is how you validate arbitrary aspects of the returned state diff. `validate` UX should analogous to our current `postCheck`, and `checkStateDiff` remains unchanged (i.e. devs never need to think about `checkStateDiff`, its always run, automatically)

test/tasks/Regression.t.sol
mds1: Sanity checking we are working on "template creation docs" that explain this? It also looks like these tests duplicate tons of code, we probably can make it a lot easier to implement and maintain these tests with some helper functions

src/improvements/tasks/MultisigTask.sol
```diff
@@ -927,7 +927,7 @@ abstract contract MultisigTask is Test, Script {
     /// check the state changes applied by the task. This function can check
     /// that only the nonce changed in the parent multisig when executing a task
     /// by checking the slot and address where the slot changed.
-    function checkStateDiff(VmSafe.AccountAccess[] memory accountAccesses) internal view virtual {
+    function checkStateDiff(VmSafe.AccountAccess[] memory accountAccesses) internal virtual {
```
blmalone, line 930: Can this function not remain view? I've made the change locally and it seems to work.

src/improvements/tasks/MultisigTask.sol
```diff
@@ -144,7 +144,7 @@ abstract contract MultisigTask is Test, Script {
 
     /// @notice abstract function to be implemented by the inheriting contract
     /// specifies the addresses that must have their storage written to
-    function _taskStorageWrites() internal pure virtual returns (string[] memory);
+    function _taskStorageWrites() internal virtual returns (string[] memory);
```
blmalone, line 147: can we keep this pure?

src/improvements/template/EnableDeputyPauseModuleTemplate.sol
blmalone: +1 to this, having l2chains included for this is a counter intuitive flow. We should also have the same flow for OPCM tasks as this is also a "superchain level" task.

src/improvements/tasks/MultisigTask.sol
blmalone: I don't think we want this to be virtual either. Templates won't be overriding this functionality. It should be run all the time regardless.

test/tasks/mock/configs/EnableDeputyPauseModuleTemplate.toml
```diff
@@ -0,0 +1,8 @@
+# this is the file used to determine the network configuration
+
+l2chains = [{name = "OP Testnet", chainId = 11155420}]
```
blmalone, line 3: Can we add an example task under `src/improvements/tasks/example/sep`? My PR will pick this up later and simulate it on every CI run.

src/improvements/tasks/MultisigTask.sol
mds1: It is currently virtual in the legacy code, but agreed it needs to be

src/improvements/template/EnableDeputyPauseModuleTemplate.sol
blmalone: I'm fine with it being a separate PR.

src/improvements/template/EnableDeputyPauseModuleTemplate.sol
```diff
@@ -0,0 +1,145 @@
...
+    /// @param chainId The ID of the L2 chain
+    function _buildPerChain(uint256 chainId) internal override {}
+
+    /// @notice Builds the action for enabling the module in the Safe
+    function _buildSingle() internal override {
+        ModuleManager(parentMultisig).enableModule(newModule);
+    }
+
+    /// @notice Validates that the module was enabled correctly
+    /// @param chainId The ID of the L2 chain to validate
```
prat-gpt, line 77: natspec nit

src/improvements/template/EnableDeputyPauseModuleTemplate.sol
blmalone: Going to resolve this issue as it is being tracked in: https://github.com/ethereum-optimism/superchain-ops/issues/396

src/improvements/template/EnableDeputyPauseModuleTemplate.sol
blmalone: Going to resolve this thread. I think @mds1 raises a good point, having the ability to do prevalidations should be possible. I'm opening an issue for this to look into it further.

blmalone: Made this issue: https://github.com/ethereum-optimism/superchain-ops/issues/654

src/improvements/template/EnableDeputyPauseModuleTemplate.sol
blmalone: Going to resolve this thread as the commit above looks like it addresses Solomons comment.

test/tasks/Regression.t.sol
blmalone: The commit above doesn't include a refactor of `Regression.t.sol` with helper functions to avoid code duplication. I've raised an issue for this to be tackled later: https://github.com/ethereum-optimism/superchain-ops/issues/655

blmalone: Resolving.

src/improvements/template/EnableDeputyPauseModuleTemplate.sol
blmalone: Resolving.

test/tasks/mock/MockMultisigTask.sol
```diff
@@ -48,7 +48,7 @@ contract MockMultisigTask is MultisigTask {
         }
```
blmalone, line 48: The build is failing here, we need this import: `import {VmSafe} from "forge-std/Vm.sol";`

blmalone (commented):
build is failing

src/improvements/template/EnableDeputyPauseModuleTemplate.sol
```diff
@@ -0,0 +1,145 @@
...
+        );
+
+        bytes32 moduleSlot = keccak256(abi.encode(newModule, MODULE_MAPPING_STORAGE_OFFSET));
+        bytes32 sentinelSlot = keccak256(abi.encode(SENTINEL_MODULE, MODULE_MAPPING_STORAGE_OFFSET));
+
+        bool moduleWriteFound;
+
+        address[] memory uniqueWrites = accountAccess.getUniqueWrites();
+        assertEq(uniqueWrites.length, 1, "should only write to foundation ops safe");
+        assertEq(uniqueWrites[0], parentMultisig, "should only write to foundation ops safe address");
```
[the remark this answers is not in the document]
blmalone: `uniqueWrites[0]` is `0xfb1bffC9d739B8D520DaF37dF666da4C687191EA` which is the same as the GnosisSafeL2 predeploy: https://github.com/safe-global/safe-smart-account/blob/a1e7f4a763952f5b116a8ab4c7361b95de2083c3/CHANGELOG.md?plain=1#L444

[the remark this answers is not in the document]
blmalone: Closing as this was fixed in: https://github.com/ethereum-optimism/superchain-ops/pull/656

blmalone (approved):

PR #629, opened 2025-02-19
fix: using internal rpc urls for forge test

blmalone:
We have failing CI pipelines in superchain-ops: https://app.circleci.com/pipelines/github/ethereum-optimism/superchain-ops 

All pipelines that are failing fail at the `forge test`, `simulate` or `simulate-nested` step. 

I've moved the internal addresses to the `foundry.toml` file. The thought process here is that we can eventually remove references to them in the circleci yml file. In the meantime, we will rely on foundry profiles to make sure the correct rpc's are used.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,8 +32,8 @@ commands:
         type: string
     steps:
       - utils/checkout-with-mise
-      - set-simulation-eth-rpc-url
-        task: "<< parameters.task >>"
+      - set-simulation-eth-rpc-url:
+          task: "<< parameters.task >>"
       - run:
           name: "simulate << parameters.task >>"
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,6 +34,8 @@ commands:
       - utils/checkout-with-mise
       - set-simulation-eth-rpc-url:
           task: "<< parameters.task >>"
+          l1_mainnet_rpc_url: "<< pipeline.parameters.l1_mainnet_rpc_url >>"
+          l1_sepolia_rpc_url: "<< pipeline.parameters.l1_sepolia_rpc_url >>"
       - run:
           name: "simulate << parameters.task >>"
           command: |
@@ -53,6 +55,8 @@ commands:
       - utils/checkout-with-mise
       - set-simulation-eth-rpc-url:
           task: "<< parameters.task >>"
+          l1_mainnet_rpc_url: "<< pipeline.parameters.l1_mainnet_rpc_url >>"
+          l1_sepolia_rpc_url: "<< pipeline.parameters.l1_sepolia_rpc_url >>"
       - run:
           name: "simulate nested << parameters.task >>"
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -311,7 +311,7 @@ jobs:
           name: forge test
           command: |
             forge --version
-            forge test -vvv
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -34,6 +34,8 @@ commands:
       - utils/checkout-with-mise
       - run:
           name: "simulate << parameters.task >>"
+          environment:
+            ETH_RPC_URL: << pipeline.parameters.l1_mainnet_rpc_url >>
```
mds1, line 38: I think this is not right, because we might be simulating on sepolia?

.circleci/config.yml
```diff
@@ -51,6 +53,8 @@ commands:
       - utils/checkout-with-mise
       - run:
           name: "simulate nested << parameters.task >>"
+          environment:
+            ETH_RPC_URL: << pipeline.parameters.l1_mainnet_rpc_url >>
```
mds1, line 57: Same

.circleci/config.yml
```diff
@@ -295,6 +299,8 @@ jobs:
       - utils/checkout-with-mise
       - run:
           name: forge test
+          environment:
+            ETH_RPC_URL: << pipeline.parameters.l1_mainnet_rpc_url >>
```
mds1, line 303: Same, we have forge tests that fork from both mainnet and sepolia

blmalone: Can you point me to the place where we fork from sepolia?

.circleci/config.yml
blmalone: https://github.com/ethereum-optimism/superchain-ops/pull/629/commits/e00e4dfa97eb42201cfdad64472349677246a048

.circleci/config.yml
blmalone: https://github.com/ethereum-optimism/superchain-ops/pull/629/commits/e00e4dfa97eb42201cfdad64472349677246a048

Ethnical:
I am wondering since we need to use the `circleci_ip_ranges: true` in almost of the function etc maybe we can factorized it in a way to put in a location that make all the necessary execution with the flag?

blmalone:
@Ethnical Right now I'm following the pattern that we have in the monorepo. I think handing out access to internal nodes on a case by case basis for each job is preferable. I'm not sure about a cleaner way to do this.

.circleci/config.yml
mds1: https://github.com/ethereum-optimism/superchain-ops/blob/877329654a1e3d731b58524ddd7a0f1eedf2abe1/test/tasks/NestedMultisigTask.t.sol#L223

.circleci/config.yml
```diff
@@ -251,15 +254,21 @@ jobs:
...
     steps:
       - simulate:
           task: "/eth/ink-002-set-respected-game-type"
 
   just_simulate_zora_002_fp_upgrade:
+    circleci_ip_ranges: true
     docker:
       - image: <<pipeline.parameters.default_docker_image>>
+    environment:
+      ETH_RPC_URL: << pipeline.parameters.l1_sepolia_rpc_url >>
```
mds1, line 271: Do we need this parameter, or can we infer this from the task path in the `simulate` and `simulate-nested` commands and set it there? Or we can add `network` as a another parameter that way we just need to specify "mainnet" or "sepolia" instead of an env var

mds1 (approved):

PR #635, opened 2025-02-20
Add readonly public repo token context

raffaele-oplabs:
**Description**

When using mise in the repo it reach unauthenticated requests to GitHub rate limit. To overcome the issue we added a readonly GITHUB_TOKEN that will be used for the requests. 
```
WARN  GitHub rate limit exceeded. Resets at 2025-02-20 04:52:17
WARN  GitHub API returned a 403 Forbidden error. This likely means you have exceeded the rate limit.
WARN  GITHUB_TOKEN is not set. This means mise is making unauthenticated requests to GitHub which have a lower rate limit.
To increase the rate limit, set the GITHUB_TOKEN environment variable to a GitHub personal access token.
Create a token at https://github.com/settings/tokens and set it as GITHUB_TOKEN in your environment.
You do not need to give this token any scopes.
```

**Tests**

**Additional context**

**Metadata**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -366,29 +366,53 @@ workflows:
         - equal: [webhook, << pipeline.trigger_source >>]
         - equal: ["build_four_hours", <<pipeline.schedule.name>>]
     jobs:
-      - print_versions
+      - print_versions:
+          context:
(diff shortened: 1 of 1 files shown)


JosepBove (approved):
Checked and talked in slack

PR #640, opened 2025-02-20
Actions/Childmultisig no longer global variables

blmalone:
This PR is a first step to refactoring MultisigTask to improve readability. Having `actions` as a global state variable makes following the code tricky. Moving towards passing around the actions allows us to easily see the flow of data and what data each function requires. 

Similarly for the `childMultisig` global variable.


foundry.toml
```diff
@@ -32,4 +32,4 @@ sepolia = "https://ci-sepolia-l1-archive.optimism.io" # Must have 'circleci_ip_r
 [profile.default.rpc_endpoints]
 localhost = "http://127.0.0.1:8545"
 mainnet = "https://ethereum.publicnode.com"
-sepolia = "https://ethereum-sepolia-rpc.publicnode.com"
+sepolia = "https://gateway.tenderly.co/public/sepolia"
```
blmalone, line 35: this is an archive node.

blmalone:
* **#641** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/superchain-ops/641?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#640** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/superchain-ops/640?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/ethereum-optimism/superchain-ops/640?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

foundry.toml
blmalone: `testRegressionCallDataMatches_OPCMUpgradeVxyz` in `Regression.t.sol` requires an archive node with the default profile.

test/tasks/MultisigTask.t.sol
```diff
@@ -100,48 +106,6 @@ contract MultisigTaskUnitTest is Test {
         task.build();
     }
 
-    function testSimulateFailsHashMismatch() public {
```
blmalone, line 103: Didn't have a simple way to reproduce this test case.

src/improvements/tasks/MultisigTask.sol
```diff
@@ -924,85 +918,101 @@ abstract contract MultisigTask is Test, Script {
...
-                        value: accountAccesses[i].value,
-                        target: accountAccesses[i].account,
-                        arguments: accountAccesses[i].data,
-                        operation: operation,
-                        description: description
-                    })
-                );
-            }
+    /// @dev Returns true if the given account access should be recorded as an action.
+    function _isValidAccess(VmSafe.AccountAccess memory access, uint256 topLevelDepth) internal view returns (bool) {
```
mds1, line 973: nit, based on the natspec, maybe
```suggestion
    /// @dev Returns true if the given account access should be recorded as an action.
    function _isValidAction(VmSafe.AccountAccess memory access, uint256 topLevelDepth) internal view returns (bool) {
```

src/improvements/tasks/MultisigTask.sol
```diff
@@ -924,85 +918,101 @@ abstract contract MultisigTask is Test, Script {
...
+                validActions[index] = Action({
+                    value: accesses[i].value,
+                    target: accesses[i].account,
+                    arguments: accesses[i].data,
+                    operation: op,
+                    description: desc
+                });
+                index++;
+            }
+        }
```
mds1, line 967: Looks like this is just a refactor, but no logic changes? This seems like a critical piece of logic, so because all tests pass I think we are confident in this change, since changing the actions would change the hash we sign, therefore a test would fail—is that correct?

test/tasks/MultisigTask.t.sol
mds1: You could use `vm.mockCall` so the result of the `.getTransactionHash()` returns the wrong hash

mds1 (dismissed):
This is great, nice work!

src/improvements/tasks/MultisigTask.sol
```diff
@@ -730,33 +724,33 @@ abstract contract MultisigTask is Test, Script {
...
+        );
         Simulation.logSimulationLink({_to: parentMultisig, _data: txData, _from: msg.sender, _overrides: overrides});
     }
 
     /// @notice get the hash for this safe transaction
-    /// can only be called after the build function, otherwise it reverts
-    function getHash() public view returns (bytes32) {
-        bytes memory data = getCalldata();
-        return keccak256(getDataToSign(parentMultisig, data));
+    function getHash(bytes memory callData, address safe) public view returns (bytes32) {
```
prat-gpt, line 746: getHash always uses `getMulticall3Calldata()` and parent multisig, I think we can avoid passing these two parameters to it.

prat-gpt (approved):
Added a comment regarding `getHash()` function. Rest looks good to me.

src/improvements/tasks/MultisigTask.sol
blmalone: I'm intentionally splitting them out so that it's clearer to see the exactly flow. That way the `getMulticall3Calldata()` function isn't hidden behind `getHash()`.

src/improvements/tasks/MultisigTask.sol
blmalone: Good call out, I went line by line through the conditional logic from the previous impl: 

```solidity
            // // calls to and from AddressRegistry and the vm contract are ignored
            // bool accountIsNotAddressRegistryOrVm =
            //     accountAccesses[i].account != address(addrRegistry) && accountAccesses[i].account != address(vm);
            // bool accessorIsNotAddressRegistry = accountAccesses[i].accessor != address(addrRegistry);
            // // only care about calls or top leveldelegate calls from the multisig, static calls are ignored
            // bool isCall = accountAccesses[i].kind == VmSafe.AccountAccessKind.Call;
            // bool isTopLevelDelegateCall = accountAccesses[i].kind == VmSafe.AccountAccessKind.DelegateCall
            //     && accountAccesses[i].depth == topLevelDepth;
            // // only record actions from the parent multisig
            // bool accessorIsParentMultisig = accountAccesses[i].accessor == parentMultisig;

            // if (
            //     accountIsNotAddressRegistryOrVm && accessorIsNotAddressRegistry && (isCall || isTopLevelDelegateCall)
            //         && accessorIsParentMultisig
            // ) {
```


and confirmed the same logic is reimplemented.

test/tasks/MultisigTask.t.sol
blmalone: I added this test back in. 
https://github.com/ethereum-optimism/superchain-ops/pull/640/commits/c0be570db05327b8422fe7c536c5c4dca85c2088

There was some tight variable packing going on between the `multicallAddress` and `_buildStarted` state variables. 

It meant I was getting: `0x00Ca11bdE05977B3631167028862be2a173976cA` instead of  `0xcA11bde05977b3631167028862bE2a173976CA11`. The first byte was the `_buildStarted` bool. I've changed `_buildStarted` to be a uint256.

src/improvements/tasks/MultisigTask.sol
```diff
@@ -578,18 +559,31 @@ abstract contract MultisigTask is Test, Script {
...
+        actions = _endBuild();
+
+        _buildStarted = false;
+
+        return actions;
     }
 
     /// @notice print task description, actions, transfers, state changes and EOAs datas to sign
-    function print() public view {
+    function print(Action[] memory actions, address optionalChildMultisig) public view {
```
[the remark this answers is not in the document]
blmalone: Addressed in: https://github.com/ethereum-optimism/superchain-ops/pull/672

src/improvements/tasks/MultisigTask.sol
```diff
@@ -564,18 +550,33 @@ abstract contract MultisigTask is Test, Script {
...
 
         AddressRegistry.ChainInfo[] memory chains = addrRegistry.getChains();
 
         for (uint256 i = 0; i < chains.length; i++) {
             _buildPerChain(chains[i].chainId);
         }
+
+        actions = _endBuild();
+
+        _buildStarted = uint256(0);
```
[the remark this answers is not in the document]
blmalone: https://github.com/ethereum-optimism/superchain-ops/pull/640/commits/710b3def3c7cc76eb4d907ba3811377bebc097e5

src/improvements/tasks/MultisigTask.sol
```diff
@@ -564,18 +550,33 @@ abstract contract MultisigTask is Test, Script {
     /// @dev contract calls must be perfomed in plain solidity.
     ///      overriden requires using buildModifier modifier to leverage
     ///      foundry snapshot and state diff recording to populate the actions array.
-    function build() public buildModifier {
+    function build() public returns (Action[] memory actions) {
+        require(parentMultisig != address(0), "Must set address registry for multisig address to be set");
+
+        console.log("_buildStarted");
+        console.log(_buildStarted);
```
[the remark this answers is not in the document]
blmalone: https://github.com/ethereum-optimism/superchain-ops/pull/640/commits/710b3def3c7cc76eb4d907ba3811377bebc097e5

test/tasks/Regression.t.sol
```diff
@@ -1,8 +1,7 @@
 pragma solidity 0.8.15;
 
-import {console} from "forge-std/console.sol";
 import {Test} from "forge-std/Test.sol";
-
+import {VmSafe} from "forge-std/Vm.sol";
```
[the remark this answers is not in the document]
blmalone: https://github.com/ethereum-optimism/superchain-ops/pull/640/commits/710b3def3c7cc76eb4d907ba3811377bebc097e5

test/tasks/MultisigTask.t.sol
```diff
@@ -4,6 +4,7 @@ pragma solidity 0.8.15;
 import {IMulticall3} from "forge-std/interfaces/IMulticall3.sol";
 import {VmSafe} from "forge-std/Vm.sol";
 import {Test} from "forge-std/Test.sol";
+import {console} from "forge-std/console.sol";
```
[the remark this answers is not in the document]
blmalone: https://github.com/ethereum-optimism/superchain-ops/pull/640/commits/710b3def3c7cc76eb4d907ba3811377bebc097e5

src/improvements/tasks/MultisigTask.sol
```diff
@@ -191,16 +175,16 @@ abstract contract MultisigTask is Test, Script {
...
-        build();
+        Action[] memory actions = build();
 
         // now execute task actions
-        VmSafe.AccountAccess[] memory accountAccesses = execute(signatures);
+        VmSafe.AccountAccess[] memory accountAccesses = execute(signatures, actions);
 
         // validate all state transitions
-        validate(accountAccesses);
+        validate(accountAccesses, actions);
```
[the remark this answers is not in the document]
blmalone: The validate function in OPCMBaseTask does use actions though.

blmalone:
@ElliotFriedman I've simulated all the example tasks individually and they're all working.

src/improvements/tasks/MultisigTask.sol
blmalone: Resolving.

test/tasks/MultisigTask.t.sol
blmalone: Resolving.

src/improvements/tasks/MultisigTask.sol
mds1, line 1: > Have you tried running the new task type from the command line? That would be good to know that changes in the `_isValidAction` function around call depth are working as expected.

@ElliotFriedman what did you mean by this? It sounds like we are missing some tests if we have to run this from the CLI to verify?

src/improvements/tasks/MultisigTask.sol
```diff
@@ -112,32 +108,12 @@ abstract contract MultisigTask is Test, Script {
     TaskConfig public config;
 
     /// @notice flag to determine if the task is being simulated
-    bool private _buildStarted;
-
-    /// @notice The address of the child multisig for this task
-    /// this state variable will only be set when function
-    /// `signFromChildMultisig` is called.
-    address public childMultisig;
+    uint256 private _buildStarted;
```
mds1, line 111: Wy is this now a uint?

mds1 (approved):

src/improvements/tasks/MultisigTask.sol
blmalone: I mentioned the reason here in this thread: https://github.com/ethereum-optimism/superchain-ops/pull/640#discussion_r1965805909

src/improvements/tasks/MultisigTask.sol
blmalone: The `simulate-all-templates` command that runs in CI will run simulate on all the templates the same way a signer would. We would need to make sure that approve, sign and execute all work. Right now we don't have tests that call these commands the same way a signer would. Only unit tests.

mds1: So if the `simulate-all-templates` runs in CI, does that accomplish what elliot was asking, since if we had a bug presumably actions would be incorrect and validations in those would fail?

blmalone: Yes I think so. I think maybe we still have a blind spot in end to end testing which is making sure _all_ commands in both `single.just` and `nested.just` woke as expected. `simulate-all-templates` covers the simulate commands for both `single.just` and `nested.just`.

mds1 (approved):

PR #645, opened 2025-02-21
Move foundry version from stable to v1.0.0

JosepBove:
We want to pin our release to foundry v1.0.0, the stable tag will change on the future once they release new versions.

The tag exists here: https://github.com/foundry-rs/foundry/releases/tag/v1.0.0


alcueca:
Good point

alcueca (approved):
Good point.

PR #652, opened 2025-02-21
fix: fix wildcard in find and redundant check in while loop

barajeel:
**Description**  
Fixed a potential issue with `find` misinterpreting `*.env` by escaping it as `\*.env`. This prevents unexpected behavior in certain shells.  

Also removed the redundant `|| [[ -n "$line" ]]` check in `while read`, as `read -r` correctly handles the last line even without a trailing newline.  

**Tests**  
Tested the `find` command with and without escaping to confirm proper behavior across different shells.  
Verified that `while read -r` correctly processes the last line without the extra condition.  

**Additional context**  
These fixes improve script reliability and prevent edge-case failures in certain environments.

### DIFF
--- a/script/utils/check-nonce-overrides.sh
+++ b/script/utils/check-nonce-overrides.sh
@@ -3,14 +3,14 @@
 search_dir="./tasks"
 
 # Find all .env files in the given directory and its subdirectories
-env_files=$(find "$search_dir" -type f -name "*.env")
+env_files=$(find "$search_dir" -type f -name '*.env')
 
 # Process each .env file
 for file in $env_files; do
     echo "Checking file: $file"
 
     # Process the file line by line
-    while IFS= read -r line || [[ -n "$line" ]]; do
+    while IFS= read -r line; do
         # Check if the line starts with SAFE_NONCE_
         if [[ $line == SAFE_NONCE_* ]]; then
             # Extract the variable name (part before =)


script/utils/check-nonce-overrides.sh
```diff
@@ -3,14 +3,14 @@
 search_dir="./tasks"
 
 # Find all .env files in the given directory and its subdirectories
-env_files=$(find "$search_dir" -type f -name "*.env")
+env_files=$(find "$search_dir" -type f -name '*.env')
```
Ethnical, line 7: Let's add also a check to make sure env_files return is not empty otherwise return an error.
```suggestion
# Check if any .env files were found.
if [ -z "$env_files" ]; then
    echo "Error no .env files found in $search_dir"
    exit 0
fi

```

script/utils/check-nonce-overrides.sh
```diff
@@ -3,14 +3,14 @@
 search_dir="./tasks"
 
 # Find all .env files in the given directory and its subdirectories
-env_files=$(find "$search_dir" -type f -name "*.env")
```
Ethnical, line 6: tried locally I don't see any problem with it :)

script/utils/check-nonce-overrides.sh
```diff
@@ -3,14 +3,14 @@
...
+env_files=$(find "$search_dir" -type f -name '*.env')
 
 # Process each .env file
 for file in $env_files; do
     echo "Checking file: $file"
 
     # Process the file line by line
-    while IFS= read -r line || [[ -n "$line" ]]; do
```
Ethnical, line 13: with this current fix we skip some line. 
I prefer to keep the previous version here.
I have run few tests and we miss approx 10 lines and 2 SAFE_NONCES line.
cf the line here: 
image

Ethnical:
Thanks for the PR!
I have open one suggestion + one comment to revert one change on the check of the line not empty.
but the rest looks nice we can merge after this 🚀

Ethnical:
I will close for now since we don't have any answer from the contributor.

PR #657, opened 2025-02-21
Unichain SystemConfig Update

winnsterx:
**Description**

**Tests**

**Additional context**

**Metadata**


tasks/eth/031-uni-holocene-systemconfig-update/NestedSignFromJson.s.sol
```diff
@@ -0,0 +1,122 @@
...
+    //         "-",
+    //         l1ChainName
+    //     );
+    //     console.log(
+    //         "with SystemConfigProxy @",
+    //         sysCfgUpgrade.systemConfigAddress()
+    //     );
+    //     addAllowedStorageAccess(sysCfgUpgrade.systemConfigAddress());
+    //     addCodeExceptions(sysCfgUpgrade.getCodeExceptions());
+    // }
```
geoknee, line 86: This should be removed if it is not being used.

tasks/eth/031-uni-holocene-systemconfig-update/NestedSignFromJson.s.sol
```diff
@@ -0,0 +1,122 @@
...
+    string constant l2ChainName = "unichain";
+    string constant release = "v1.8.0-rc.4";
+    // string[1] l2ChainNames = ["unichain"];
+
+    bytes32 constant ABSOLUTE_PRESTATE =
+        0x0336751a224445089ba5456c8028376a0faf2bafa81d35f43fab8730258cdf37;
+    address constant FAULT_DISPUTE_GAME =
+        0x08f0F8F4E792d21E16289dB7a80759323C446F61;
+    address constant PERMISSIONED_DISPUTE_GAME =
+        0x2B9fD545CcFC6611E5F1e3bb52840010aA64C5C6;
```
geoknee, line 30: I don't think these are used, could be removed.

tasks/eth/031-uni-holocene-systemconfig-update/README.md
```diff
@@ -0,0 +1,64 @@
+# Holocene Hardfork Upgrade - `SystemConfig`
+
+Status: DRAFT, NOT READY TO SIGN
+
+## Objective
+
+Upgrades the **Unichain Mainnet** Fault Proof PDG to a new challenger and upgrades the System Config to the Holocene version.
```
geoknee, line 7: Can we please double check this. From the state diff, I don;'t think we are are doing anything with the fault proofs system with this task.

tasks/eth/031-uni-holocene-systemconfig-update/input.json
```diff
@@ -0,0 +1,105 @@
...
...0000000000000000000000000000000000000000000000000dbba00000000000000000000000002f60a5184c63ca94f82a27100643dbabe4f3f7fd0000000000000000000000000000000000000000000000000000000001c9c380000000000000000000000000833c6f278474a78658af91ae8edc926fe33a230e0000000000000000000000000000000000000000000000000000000001312d00000000000000000000000000000000000000000000000000000000000000000a0000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000000000000003b9aca0000000000000000000000000000000000000000000000000000000000000f424000000000000000000000000000000000ffffffffffffffffffffffffffffffff000000000000000000000000ff000000000000000000000000000000000001300000000000000000000000009a3d64e386c18cb1d6d5179a9596a4b5736e98a6000000000000000000000000d04d0d87e0bd4d2e50286760a3ef323fea6849cf00000000000000000000000081014f44b0a345033bb2b3b21c7a1a308b35feea0000000000000000000000002f12d621a16e2d3285929c9996f478508951dfe40000000000000000000000000bd48f6b86a26d3a217d0fa6ffe2b491b956a7a2000000000000000000000000a2b597eaeacb6f627e088cbead319e934ed5edad000000000000000000000000eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee00000000000000000000000000000000000000000000000000000000",
```
geoknee, line 74: I believe the validation is failing because this tx bundles a change to the `blobBaseFeeScalar` (it was implicitly 0, it is changing to 900000). The verification system expects the `blobBaseFeeScalar` to remain unchanged.

We either need to extend the verification to permit this case, or regenerate the tx bundle so that `blobBaseFeeScalar` is set to 0 when the contract initializes.

tasks/eth/031-uni-holocene-systemconfig-update/VALIDATION.md
```diff
@@ -0,0 +1,108 @@
...
+above.
+
+## State Changes
+
+### `0xc407398d063f942feBbcC6F80a156b47F3f1BDA6` (`SystemConfigProxy`)
+
+- **Key**: `0x0000000000000000000000000000000000000000000000000000000000000066`
+  **Before**: `0x00000000000000000000000000000000000000000000000000000000000dbba0`
+  **After**: `0x010000000000000000000000000000000000000000000000000dbba0000007d0`
+  **Meaning**: Updates the `SystemConfig` proxy implementation.
```
geoknee, line 84: This is actually the scalar slot. See e.g. https://github.com/ethereum-optimism/superchain-ops/blob/ea0cadfbbc49c94d13202017c9ff2111d55cdd65/tasks/eth/028-holocene-system-config-upgrade-and-init-multi-chain/VALIDATION.md?plain=1#L67-L70.

The `dbba0` bit being nonzero is what is causing the validation to fail.

geoknee (dismissed):

tasks/eth/031-uni-holocene-systemconfig-update/input.json
sebastianst: we could add an optional override via an env var, a pattern we used for other optional overrides.

sbvegan: I do think it was intentional, but we're waiting for confirmation from @trianglesphere who is OOO until Monday. If it is intentional, I prefer we take the route that seb is suggesting.

.circleci/config.yml
```diff
@@ -283,7 +283,14 @@ jobs:
...
-          
+
+  just_simulate_031_uni_holocene_systemconfig_update:
+    docker:
+      - image: <<pipeline.parameters.default_docker_image>>
+    steps:
+      - simulate_nested:
+          task: "/eth/031-uni-holocene-systemconfig-update"
+
   simulate_sequence:
```
Ethnical, line 294: We also need to add the task 031 in the sim sequence **AFTER** the PR ( https://github.com/ethereum-optimism/superchain-ops/pull/683 ) is merged. 
We will have to merge and then add the 031 after the 030 to simulate in sequence.

Ethnical: This one has been merged so we can add into the ci. 
I will make the suggestion

tasks/eth/031-uni-holocene-systemconfig-update/input.json
winnsterx: This is confirmed to be an intended behaviour to correct their current L1 contract's blobBaseFeeScalar. https://uniscan.xyz/address/0x420000000000000000000000000000000000000f#readProxyContract 

Could we go ahead with overriding as @sebastianst suggested if isnt an issue? @sbvegan @geoknee @sebastianst @mds1

sebastianst: To implement an override in the `SystemConfig` generic verification script, you can read from an env var, if present, to configure the optional override. E.g. [here](https://github.com/ethereum-optimism/superchain-ops/blob/f3f2a089bc5c6e007526f38f27c264b2638bce1d/script/verification/Verification.s.sol#L188) we implemented optional overrides for the MIPS address and version.
Or [here](https://github.com/ethereum-optimism/superchain-ops/blob/f3f2a089bc5c6e007526f38f27c264b2638bce1d/script/verification/DisputeGameUpgrade.s.sol#L111) for optionally allowing the delayed weth to change (we should actually have implemented it better with a specific address, not only a bool...).
I'll see if I can push a quick example.

sebastianst: I see that this task actually already configures 3 overrides in the `.env` file for the dispute game check script.

winnsterx: Could we skip these checks based on the chain names (unichain/mainnet) in the verification script?

sebastianst: I pushed a commit that contains the override mechanism and makes the tests pass.

.circleci/config.yml
```diff
@@ -449,27 +463,20 @@ workflows:
...
-          context:
-            - circleci-repo-readonly-authenticated-github-token
 
+      # Simulation sequence for Superchain Version Sync (at the 27th of February 2025). The next task would be 31 but we need to wait until the task is 31 is merged.
+      # tasks: "030 031"
       - simulate_sequence:
           name: simulate_sequence_eth
           network: "eth"
-          tasks: "ink-003"
+          tasks: "unichain-001"
```
Ethnical, line 479: here, this is the whole sequence you need to simulate so first ink and then uni task like this: 
```suggestion
          tasks: "ink-003 unichain-001"
```

lib/base-contracts
Ethnical, line 1: Looks strange to me to upgrade the `lib` here. 
usually this is not desired. 
if yes why do we need to upgrade the `lib` ?

.circleci/config.yml
```diff
@@ -265,6 +265,20 @@ jobs:
               --justfile ../../../nested.just \
               simulate council
 
+  just_simulate_ink_sep_holocene_system_config_upgrade:
```
Ethnical, line 268: Why do we add this?

lib/base-contracts
Ethnical: Fixed it now.

script/verification/HoloceneSystemConfigUpgrade.s.sol
```diff
@@ -64,6 +64,7 @@ contract HoloceneSystemConfigUpgrade is SuperchainRegistry, VerificationBase {
     {
         systemConfigAddress = proxies.SystemConfig;
         sysCfg = ISystemConfig(proxies.SystemConfig);
+        addAllowedStorageAccess(systemConfigAddress); //TODO: check if necessary here
```
Ethnical, line 67: @winnsterx I tried and this line seems to not be necessary could you explain me more why we add this here?

tasks/eth/unichain-001-holocene-systemconfig-update/VALIDATION.md
```diff
@@ -0,0 +1,76 @@
...
+### `0xc2819DC788505Aac350142A7A707BF9D03E3Bd03` (Council Safe)
+- **Key:** `0x0000000000000000000000000000000000000000000000000000000000000005` <br/>
+  **Value:** `0x0000000000000000000000000000000000000000000000000000000000000014` <br/>
+  **Meaning:** Override the nonce value of the `Security Council` by increasing from 18 to 19.
+
+
+### `0x847B5c174615B1B7fDF770882256e2D3E95b9D92` (Foundation Upgrade Safe)
+- **Key:** `0x0000000000000000000000000000000000000000000000000000000000000005` <br/>
+  **Value:** `0x0000000000000000000000000000000000000000000000000000000000000012` <br/>
+  **Meaning:** Override the nonce value of the Foundation Upgrade Safe by increasing from 16 to 17.
```
Ethnical, line 38: Override are incorrect now should 17 -> 18

Ethnical:
This matches the validation file we can soon merge this when all the discussion are resolve :) 
eb95cdb375da1b52ea2a1d6f0e02251fdd64651fcab568425625553b3d0437bf

Ethnical (approved):
Perfect!

PR #670, opened 2025-02-25
Fix/nested just safes

blmalone:
Building on https://github.com/ethereum-optimism/superchain-ops/pull/662 from @prat-gpt. Shellcheck wasn't working.


alcueca:
The change looks good, but how did we miss this with the testing?

alcueca (approved):

PR #680, opened 2025-02-26
030 fix: updating input and validation to reflect 40M gas limit

sbvegan:
**Description**

**Tests**

**Additional context**

**Metadata**


maurelian (approved):
Checks are passing, meaning there are no longer unexpected config changes.

I've also decoded both the storage value and input data changes to confirm that they now set the gas to 40MM.

PR #681, opened 2025-02-26
fix: fixed bug in simulation of tasks.

blmalone:
I noticed the isNestedSafe forge script wasn't parsing the return value correctly. It's now working as expected. I also clarified the purpose of the `example` directory with a simple readme. Moved a check into root simulateRun function for full coverage.

### DIFF
--- a/src/improvements/justfile
+++ b/src/improvements/justfile
@@ -110,31 +110,16 @@ simulate-all-templates:
     #!/usr/bin/env bash
     set -euo pipefail
     root_dir=$(git rev-parse --show-toplevel)
-    nested_just_file="${root_dir}/src/improvements/nested.just"
-    single_just_file="${root_dir}/src/improvements/single.just"
     forge build
 
     simulation_count=0
-    
     for task in ${root_dir}/src/improvements/tasks/example/*/*; do
-        echo ""
         if [ -d "$task" ]; then
-            rpcUrl=$(${root_dir}/src/improvements/script/get-rpc-url.sh $task)
-            echo "Task: $task"
-            is_nested=$(forge script ${root_dir}/src/improvements/tasks/TaskRunner.sol --sig "isNestedTask(string)" "$task/config.toml" --rpc-url $rpcUrl -vvvv)
-            if [ "$is_nested" = "true" ]; then
-                (cd $task && SIMULATE_WITHOUT_LEDGER=1 just --dotenv-path $(pwd)/.env --justfile $nested_just_file simulate foundation)
-                echo "Task $task is nested"
-            else
-                (cd $task && SIMULATE_WITHOUT_LEDGER=1 just --dotenv-path $(pwd)/.env --justfile $single_just_file simulate)
-                echo "Task $task is not nested"
-            fi
+            ${root_dir}/src/improvements/script/simulate-task.sh $task
             simulation_count=$((simulation_count + 1))
         fi
     done
-
-    echo "\n$simulation_count simulations run."
-
+    echo "$simulation_count simulations run."
     template_count=$(find "${root_dir}/src/improvements/template" -type f ! -name "*.template.sol" | wc -l)
 
     if [ "$simulation_count" -ne "$template_count" ]; then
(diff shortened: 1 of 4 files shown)


src/improvements/tasks/MultisigTask.sol
```diff
@@ -143,6 +143,10 @@ abstract contract MultisigTask is Test, Script {
         validate(accountAccesses, actions);
         print(actions, accountAccesses, optionalChildMultisig, true);
 
+        if (optionalChildMultisig != address(0)) {
```
blmalone, line 146: I think this logic is better here incase a nested safe ever uses a different entrypoint that `signFromChildMultisig` to simulate.

src/improvements/justfile
```diff
@@ -110,31 +110,16 @@ simulate-all-templates:
...
     forge build
 
     simulation_count=0
-    
     for task in ${root_dir}/src/improvements/tasks/example/*/*; do
-        echo ""
         if [ -d "$task" ]; then
-            rpcUrl=$(${root_dir}/src/improvements/script/get-rpc-url.sh $task)
-            echo "Task: $task"
-            is_nested=$(forge script ${root_dir}/src/improvements/tasks/TaskRunner.sol --sig "isNestedTask(string)" "$task/config.toml" --rpc-url $rpcUrl -vvvv)
```
blmalone, line 124: This was wrong, I wasn't capturing the return type correctly.

src/improvements/tasks/example/README.md
```diff
@@ -0,0 +1 @@
+This `example` directory contains example tasks that are used to make sure templates can always be simulated. When adding a new template, you should add a new task in this directory and make sure that it is simulated via the `template_regression_tests` job in the `circleci/config.yml` file.
```
mds1, line 1: > make sure that it is simulated via the `template_regression_tests` job in the `circleci/config.yml` file.

How do you do this? It looks like that job calls `just simulate-all-templates`, so does this actually just happen automatically?

src/improvements/tasks/MultisigTask.sol
blmalone: It's not used anywhere else for now so I'll leave it as is.

src/improvements/tasks/example/README.md
blmalone: Yes, it'll happen automatically once you add a new task into this directory.

mds1: got it, so my nit is let's clarify that in the readme, e.g. this:
```suggestion
This `example` directory contains example tasks that are used to make sure templates can always be simulated. When adding a new template, you MUST add a new task in this directory. This will make sure that it is simulated via the `template_regression_tests` job in the `circleci/config.yml` file.
```

src/improvements/script/simulate-task.sh
```diff
@@ -0,0 +1,28 @@
...
+    single_just_file="${root_dir}/src/improvements/single.just"
+
+    rpcUrl=$("$root_dir"/src/improvements/script/get-rpc-url.sh "$task")
+    echo "Task: $task"
+    is_nested=$(forge script "$root_dir"/src/improvements/tasks/TaskRunner.sol --sig "isNestedTask(string)" "$task/config.toml" --rpc-url "$rpcUrl" --json | jq -r '.returns["0"].value')
+    echo "Is nested: $is_nested"
+    pushd "$task" > /dev/null
+    if [ "$is_nested" = "true" ]; then
+        echo "Simulating nested task: $task"
+        SIMULATE_WITHOUT_LEDGER=1 just --dotenv-path "$(pwd)"/.env --justfile "$nested_just_file" simulate foundation
```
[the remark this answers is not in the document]
blmalone: I didn't think we needed to simulate as chain gov and council also. I wasn't sure if it was giving up much more assurances in our tests. 

However, I think I should optionally pass the safe name into this script and let the caller decide what role should be used for a nested simulation. 

I'll add this as a fix.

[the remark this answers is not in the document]
blmalone: I'm still only simulating as foundation but I've made it more explicit.
https://github.com/ethereum-optimism/superchain-ops/pull/681/commits/e6639a250171dd2d5c547f3ed8e433151ffba57e

mds1 (approved):

Issue #646, opened 2025-02-21
[Tracking: Sim-Seq] Kill anvil after simulation


Ethnical:
An issue, if `ps aux` doesn't work in the CI. so when using the `ps` to find which pid we neeed to kill doesn't work and make the CI failling.
In the future, maybe we should get the PID in the env var and kill it directly at the end.  
```
#TODO: In the future, we should kill the anvil fork here?
```

https://github.com/ethereum-optimism/superchain-ops/pull/611/files

blmalone:
I think this can be closed as we do not use sim-seq anymore. Please let me know if I'm missing something.

PR #683, opened 2025-02-27
[SIMULATION] Sequence Superchain Version Sync Tasks

Ethnical:
**Description**
To sum up this short PR adds: 
- Added the support for Sepolia Sim Sequence.
- Added checks that `.env` file of the tasks simulated is correct to enforce more and avoid nonce issue that we had in the past with `SAFE_NONCE` instead of `SAFE_NONCE_XXXX`
- Fix the issue that Anvil was staying in background for ever (#646)
- Added the simulation of the task 30 for now. Since this is the **only** one merged of the sequence of the  _Superchain Version Sync_ 
PS: Once done, we should also add this to each tasks into each new PRs created like (031, 032, ..etc). 

**Tests**
I did bunch of tests locally and this is producing the expect  output with sep, and eth tasks.
599dce96a486d096fbab1efecc1622fefc68b4a45c2c9b535a69bbd4544d6ab5

**Additional context**

**Metadata**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -478,9 +478,10 @@ workflows:
       - just_simulate_ink_sep_holocene_system_config_upgrade:
           context:
             - circleci-repo-readonly-authenticated-github-token
+      # Simulation sequence for at the 27th of February 2025. the next tasks would be also 31 but we need to wait until this is merge and updated here.
+      # tasks: "030 031"
 
       - simulate_sequence:
           network: "eth"
-          tasks: "021 022 base-003 ink-001"
-          block_number: "21573136" # If not specified, the latest block number is used.
-
+          tasks: "030"
+          block_number: "21936400" # If not specified, the latest block number is used.
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -483,4 +483,4 @@ workflows:
       - simulate_sequence:
           network: "eth"
           tasks: "030"
-          block_number: "21936400" # If not specified, the latest block number is used.
+          # block_number: "21936400" # If not specified, the latest block number is used.
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -291,8 +291,9 @@ jobs:
         type: string
       tasks:
         type: string
-      block_number: # If not specified, the latest block number is used.
+      block_number:
         type: string
+        default: null # If not specified, the latest block number is used.
     docker:
       - image: <<pipeline.parameters.default_docker_image>>
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -478,7 +478,7 @@ workflows:
       - just_simulate_ink_sep_holocene_system_config_upgrade:
           context:
             - circleci-repo-readonly-authenticated-github-token
-      # Simulation sequence for at the 27th of February 2025. the next tasks would be also 31 but we need to wait until this is merge and updated here.
+      # Simulation sequence for Superchain Version Sync (at the 27th of February 2025). The next task would be 31 but we need to wait until the task is 31 is merged.
       # tasks: "030 031"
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -478,9 +478,9 @@ workflows:
...
-
+      # Simulation sequence for Superchain Version Sync (at the 27th of February 2025). The next task would be 31 but we need to wait until the task is 31 is merged.
+      # tasks: "030 031"
       - simulate_sequence:
           network: "eth"
-          tasks: "021 022 base-003 ink-001"
-          block_number: "21573136" # If not specified, the latest block number is used.
-
+          tasks: "030"
+          block_number: "21936400" # If not specified, the latest block number is used.
```
mds1, line 486: we should definitely use `latest` here since this is for a production task

.gitignore
```diff
@@ -25,3 +25,4 @@ cache/
 tmp-*.json
 draft*signer*.json
 ready-*
+broadcast/NestedSignFromJson.s.sol/1
```
mds1, line 28: just curious, why is this needed?

Ethnical: Just to avoid people to push all the broadcast tx that are created and will be accidentally pushed.

.circleci/config.yml
```diff
@@ -478,9 +478,9 @@ workflows:
...
             - circleci-repo-readonly-authenticated-github-token
-
+      # Simulation sequence for Superchain Version Sync (at the 27th of February 2025). The next task would be 31 but we need to wait until the task is 31 is merged.
+      # tasks: "030 031"
       - simulate_sequence:
           network: "eth"
-          tasks: "021 022 base-003 ink-001"
-          block_number: "21573136" # If not specified, the latest block number is used.
-
+          tasks: "030"
```
sbvegan, line 485: how does this tasks naming convention work? does it just pull the left side of the task directory's string?

sbvegan: Okay never mind, I found the script function that pattern matches the directory name:

```
# Find unique task folder(s) for a given task ID
find_task_folder() {
  local task_id="$1"
  matching_folders=$(find "$task_base_dir" -name "${task_id}*" -type d)

  if [[ -z "$matching_folders" ]]; then
    error_exit "No folder found matching task ID '$task_id'."
  fi

  matching_count=$(echo "$matching_folders" | wc -l)
  if [[ "$matching_count" -gt 1 ]]; then
    echo "Error: Multiple folders found matching task ID '$task_id':" >&2
    echo "$matching_folders" >&2
    exit 1
  fi

  echo "$matching_folders"
}
```

.circleci/config.yml
mds1: Yes, for this batch of tasks we should definitely have such a process in place! After tasks are executed and we update their status, we'd include this change in the PR

script/utils/sim-sequence.sh
```diff
@@ -86,9 +95,8 @@ cleanup() {
 
   # Kill the anvil fork at the end if it was started by this script
   if $DESTROY_ANVIL_AFTER_EXECUTION; then
-    echo "Anvil is still open. To kill it, please use the command below:"
-    echo "ps aux | grep anvil | grep -v grep | awk '{print \$3}' | xargs kill"
-    #TODO: In the future, we should kill the anvil fork here?
+    echo "Kill anvil with the PID: \"$ANVIL_PID\" since the flag DESTROY_ANVIL_AFTER_EXECUTION is set to \"TRUE\"."
```
mds1, line 98:
```suggestion
    echo "Killing anvil with the PID: \"$ANVIL_PID\" since the flag DESTROY_ANVIL_AFTER_EXECUTION is set to \"TRUE\"."
```

mds1 (approved):

PR #688, opened 2025-02-28
refactor: start decoupling MultisigTask and AddressRegistry

mds1:
Begins decoupling MultisigTask and AddressRegistry. Remaining step is moving `chains` specific things out of `_taskSetup` and into `L2TaskBase`, which will allow us to delete a lot of the duplicate `_deployAddressRegistry` code

cc @prat-gpt @ElliotFriedman

### DIFF
--- a/src/improvements/tasks/MultisigTask.sol
+++ b/src/improvements/tasks/MultisigTask.sol
@@ -521,21 +521,17 @@ abstract contract MultisigTask is Test, Script {
         _buildStarted = 1;
 
         _startBuild();
-
-        _buildSingle();
-
-        AddressRegistry.ChainInfo[] memory chains = addrRegistry.getChains();
-
-        for (uint256 i = 0; i < chains.length; i++) {
-            _buildPerChain(chains[i].chainId);
-        }
-
+        _build();
         actions = _endBuild();
 
         _buildStarted = 0;
 
         return actions;
     }
+
+    /// @notice This is essentially a solidity script of the calls you want to make, and its
+    /// contents are extracted into calldata for the task.
+    function _build() internal virtual;
 
     /// @notice print task description, actions, transfers, state changes and EOAs datas to sign
     function print(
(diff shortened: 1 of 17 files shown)


src/improvements/tasks/MultisigTask.sol
```diff
@@ -1168,3 +1156,9 @@ abstract contract MultisigTask is Test, Script {
         return true;
     }
 }
+
+abstract contract L2TaskBase is MultisigTask {
```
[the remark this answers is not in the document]
mds1: Will do that at the end once refactor is done to minimize diff

src/improvements/tasks/MultisigTask.sol
```diff
@@ -225,72 +261,23 @@ abstract contract MultisigTask is Test, Script {
...
-        );
+        require(bytes(config.safeAddressString).length == 0, "MultisigTask: already initialized");
 
         config.safeAddressString = safeAddressString();
         config.allowedStorageWriteAccesses = _taskStorageWrites();
         config.allowedStorageWriteAccesses.push(safeAddressString());
 
-        // set the AddressRegistry
-        addrRegistry = _addrRegistry;
+        IGnosisSafe _parentMultisig; // TODO parentMultisig should be of type IGnosisSafe
```
blmalone, line 270: Is this a todo to change the type of the state variable `parentMultisig` to have IGnosisSafe type?

blmalone (approved):
Looks great! Do you want to break L2TaskBase into its own file? Let me know and I can reapprove.

src/improvements/tasks/MultisigTask.sol
mds1: I'm going to move that into a separate file at the end, just to make diffs easier to read in the interim (https://github.com/ethereum-optimism/superchain-ops/pull/688#discussion_r1975709845)

And yea I do want to change that type, but it has a cascading effect through this file giving a large diff, so it's not urgent and I'll come back to that

Issue #203, opened 2024-05-21
Enforce folder numbering


maurelian:
We should add a check in CI to ensure that within a task folder, the numbering of tasks is increasing and does not overlap.

ie. #200 currently introduces a new dir `tasks/sep/009-fpac-upgrade` which is incorrect since there is already a `tasks/sep/009-presigned-pause`. 

This happens easily due to race conditions between new tasks, so we should automatically check for it.

ShubhSensei:
May i work on this issue?

mds1:
Hey @ShubhSensei, are you still interested in working on this? If so, here is my thinking for what we should do:
- Add a new `script/utils/check-task-numbering.sh`
- This is a shell script that loops through every `tasks/{network}` folder. These folders contain a bunch of tasks of the form `{number}-{optionalOtherNumber-}{taskName}`
- We call `{number}-{optionalOtherNumber-}` the task prefix
- For each folder, ensure that every task prefix is unique
- Run this new shell script in CI

I think we should only check for uniqueness, and not for any type of ordering or incrementing pattern. This is because:
1. Uniqueness is simpler
2. We do not always add tasks in order, i.e. sometimes for a bundle of upcoming tasks we'll add task 018 before task 017 is ready, but we don't want to block task 018 from being merged

@maurelian what do you think of this spec?

ShubhSensei:
Yes @mds1 of course... I'm on it 🫡 
BTW, is their any other medium where I can solve my queries otherwise it'll take forever like this.

mds1:
Thanks, assigned! And apologies for the slow response here, I had meant to come back to this after #152 was addressed but never did 😅

maurelian:
The spec LGTM yes.

ShubhSensei:
How do I run this script in CI? @mds1

mds1:
> How do I run this script in CI?

I left a comment on the PR with this info

alcueca:
@ShubhSensei, thanks for your initial work on this, I gave it another go.

blmalone:
This issue has been implemented and can now be closed.

PR #689, opened 2025-02-28
Enforce task numbering in CI

alcueca:
**Description**

A CI check to ensure tasks are appropriately numbered, defined as `nnn-task`.

**Additional context**

From issue #203, taken over where #340 went stale.


alcueca:
This is working, but there are a number of tasks that fail the numbering scheme. I suggest just renaming them from nnn-foo or nnn-n-foo to nnn.n-foo, which would do the trick.

mds1:
@blmalone IIRC you were working on something similar to this one? Assigning to you as a result :)

script/utils/check-task-numbering.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
```
blmalone, line 1: Can we put this in `src/improvements/script`? I don't think we should be applying this to old tasks.

blmalone: Also this PR is behind I think and needs updated.

script/utils/check-task-numbering.sh
```diff
@@ -0,0 +1,40 @@
...
+set -euo pipefail
+
+# This script checks that all task directories have a unique prefix, defined as a number
+# followed by a hyphen and optionally some text denoting the network before the number.
```
blmalone, line 5: Can you give an example, and what the upper and lower bounds of the file names could be?

script/utils/check-task-numbering.sh
alcueca: Done

script/utils/check-task-numbering.sh
alcueca: Outdated, but now that we are in `src-improvements` I've refactored so that we demand that all tasks are "nnn-<string>", unlike some old tasks that are "nnn.nnn-<string>" or "<network>-nnn-<string>".

If we run out of numbers or we really want to group tasks we can change it, but for now I would try this simpler approach.

alcueca: As for the file names, as long as you do "nnn-*", go wild.

blmalone:
@alcueca I'm going to close this PR, we have ordered numbering of tasks in the new superchain-ops system. This is a CI job and we get it for free with stacked simulations.

PR #690, opened 2025-02-28
Update 030-soneium to executed

maurelian:


.circleci/config.yml
```diff
@@ -468,9 +465,10 @@ workflows:
       - just_simulate_ink_sep_holocene_system_config_upgrade:
           context:
             - circleci-repo-readonly-authenticated-github-token
+
       # Simulation sequence for Superchain Version Sync (at the 27th of February 2025). The next task would be 31 but we need to wait until the task is 31 is merged.
       # tasks: "030 031"
```
maurelian, line 470:
```suggestion
      # tasks: "031"
```

Ethnical (dismissed):

.circleci/config.yml
maurelian, line 470:
```suggestion
      # Simulation sequence for Superchain Version Sync. The next task would be 31 but we need to wait until task 31 is merged.
      # tasks: "030 031"
```

.circleci/config.yml
```diff
@@ -468,9 +465,10 @@ workflows:
       - just_simulate_ink_sep_holocene_system_config_upgrade:
           context:
             - circleci-repo-readonly-authenticated-github-token
-      # Simulation sequence for Superchain Version Sync (at the 27th of February 2025). The next task would be 31 but we need to wait until the task is 31 is merged.
-      # tasks: "030 031"
+
+      # Simulation sequence for Superchain Version Sync. The next task would be 31 but we need to wait until task 31 is merged.
+      # tasks: "031"
       - simulate_sequence:
```
Ethnical, line 471: Here, should be until I allow to simulate the sequence empty. 

```
   # Simulation sequence for Superchain Version Sync. The next task would be 31 but we need to wait until task 31 is merged.
      # For now we are simulating with task 030 at a previous block. 
      # tasks: "031"
      - simulate_sequence:
          network: "eth"
          tasks: "030"
          block_number: "21938584" # If not specified, the latest block number is used.
```

Ethnical (approved):
Thanks!!

PR #693, opened 2025-02-28
refactor simulate-nested command into a just-simulate-nested job

maurelian:
I think that the way we setup task specific jobs is unnecessary, and we can instead just convert simulate-nested into a reusable job.

This should make task PRs a bit less noisy and reduce churn in this repo.


Ethnical (dismissed):
Looking good to me just need to rebase the recent fix. 
Thanks for the improvements :)

blmalone (approved):

PR #701, opened 2025-03-01
fix: _isValidAction bug - only record top level call access kind

blmalone:
I encountered this issue when I was trying to make a delegatecall from within a `_build()` function in a template. If the delegatecall contains _calls_ inside the top level delegate call, then they incorrectly get added as a valid action. I do not believe this is intended behavior. 

Would be good to get @prat-gpt to review as I know you wrote this logic originally. 

I was wondering "why don't non-top level calls get included in valid actions right now?". I believe the reason they don't is because of the line `bool accessorIsParent = (access.accessor == parentMultisig);` in `_isValidAction`.


blmalone:
Closing in favor of: https://github.com/ethereum-optimism/superchain-ops/pull/877

PR #703, opened 2025-03-02
refactor: superchain address registry

mds1:
Now, the only thing `MultisigTask` needs to know about an address registry is its address, so it can be filtered out in the `_isValidAction` method

- Renames AddressRegistry to SuperchainAddressRegistry
- Deduplicates the two corresponding test files into a single test file
- Cleaner `addresses.toml` format, and key names now match the network short names
- Removes unused functions from SuperchainAddressRegistry
- Simplifies code and data structures in SuperchainAddressRegistry
- Organizes functions in SuperchainAddressRegistry

cc @blmalone @AmadiMichael @ElliotFriedman @prat-gpt


test/registry/SuperchainAddressRegistry.t.sol
```diff
@@ -0,0 +1,237 @@
...
+        addrRegistry.getAddressInfo(address(0x1234567890123456789012345678901234567890));
+    }
+
+    function testInvalidChainIdInSuperchainsFailsInConstruction() public {
+        string memory networkConfigFilePath = "test/registry/mock/invalidChainIdNetworkConfig.toml";
+
+        vm.expectRevert("SuperchainAddressRegistry: Invalid chain ID in config");
+        new SuperchainAddressRegistry(networkConfigFilePath);
+    }
+
```
blmalone, line 213: One extra test we could add is: 

```solidity
    function testMissingConfigTomlFileFails() public {
        string memory missingConfigPath = "test/registry/mock/nonexistentConfig.toml";

        vm.expectRevert("SuperchainAddressRegistry: Config file not found test/registry/mock/nonexistentConfig.toml");
        new SuperchainAddressRegistry(missingConfigPath);
    }
```

Then inside the constructor of `SuperchainAddressRegistry.sol` we could change the read file logic to be: 
```solidity
        require(gasleft() > 500_000, "insufficient gas for readFile() call"); // EIP-150 safe.
        try vm.readFile(configPath) returns (string memory fileContent) {
            try vm.parseToml(fileContent, ".l2chains") returns (bytes memory parsedChainListContent) {
                chainListContent = parsedChainListContent;
            } catch {
                revert(
                    string.concat("SuperchainAddressRegistry: Failed to parse network config file path ", configPath)
                );
            }
        } catch {
            revert(string.concat("SuperchainAddressRegistry: Config file not found ", configPath));
        }
```

blmalone (dismissed):
Looks good to me.

test/registry/SuperchainAddressRegistry.t.sol
mds1: I don't think we need this since `vm.readFile` will fail if the file is not found. I just tested it and got

> [FAIL: vm.readFile: failed to read from "/Users/mds/Documents/projects/superchain-ops/test/registry/mock/nonexistentConfig.toml": No such file or directory (os error 2)] testMissingConfigTomlFileReverts() (gas: 6008105)

blmalone (approved):

Issue #679, opened 2025-02-26
Check domain hash from simulation matches VALIDATIONS.md hash entry.


blmalone:
In CI we should sim tasks and check that the domain and message hash matches what's in the validations file.

This assumes that we have already added hashes to the VALIDATIONS file. This should be a separate task. Perhaps the decodAndPrint function on the AccountAccessParser lib can help with this.

alcueca:
Working on retrieving the domain and message hashes from a simulation in #704.

blmalone:
This code below (subject to change) goes into the circle ci config:
```
  # Simulate any non-terminal tasks that will eventually be executed.
  simulate_non_teminal_tasks:
    circleci_ip_ranges: true
    docker:
      - image: <<pipeline.parameters.default_docker_image>>
    environment:
      FOUNDRY_PROFILE: ci
    steps:
      - utils/checkout-with-mise
      - run:
          name: simulate non terminal tasks
          command: |
            (cd src/improvements && just simulate-non-terminal-tasks)
      - notify-failures-on-develop:
			mentions: "@security-team”
```


This code below (subject to improvements and change) would go into `src/improvements/justfile`. 
```
simulate-non-terminal-tasks:
    #!/usr/bin/env bash
    set -euo pipefail
    root_dir=$(git rev-parse --show-toplevel)
    forge build

    # Later this networks list should be dynamically generated from the src/improvements/tasks directory.
    networks=("eth" "sep")
    for network in ${networks[@]}; do
        if [ "$network" != "src/improvements/tasks/example" ]; then # skip example tasks
            for task in ${root_dir}/src/improvements/tasks/${network}/*; do
                ${root_dir}/src/improvements/script/simulate-task.sh $task
            done
        fi
        echo "Done simulating non-terminal tasks for network: $network"
    done
```

alcueca:
Did part of the work on #736, still pending:

- Decide on a format for including hashes in VALIDATIONS.md
- Extend scripts to all safes, not just foundation
- Exclude examples directory
- Add just simulate-non-terminal-tasks to CI with tenderly context (for the Tenderly Access Token)

alcueca:
Closed #736, but now #783 has all the scripts necessary. A different PR will get them on CI.

Issue #195, opened 2024-05-10
Automatically check the domain and message hashes sent to Ledger matches the ones in the simulation


zchn:
Recently, due to a bug fixed in https://github.com/base-org/contract-deployments/pull/166/files, we were getting different hashes sent to the ledger devices compared to the ones in the Tenderly simulation. It would be great if we can detect this by automatically checking the domain and message hashes sent to Ledger matches the ones in the simulation. This may require us to migrate away from Tenderly to a more programmatically parsable one.

See https://www.notion.so/oplabs/PM-48-Wrong-message-hash-sent-to-Signers-Ledger-during-an-L2-contract-upgrade-c893131d9daa440bad6fe09c848db54d for the original postmortem

mds1:
We can use the Tenderly simulation API for this. This may require users providing their own tenderly data (I believe `TENDERLY_ACCESS_TOKEN`, `TENDERLY_USER`, and `TENDERLY_PROJECT_SLUG` are the three components needed). This API will return JSON that contains everything shown in the UI. We can parse this to extract the domain and message hashes and ensure it matches what forge provides

In the `_postCheck` hooks (one in [`SignFromJson`](https://github.com/ethereum-optimism/superchain-ops/blob/808004cd933e0bfa36e6d9124701e296639a6d41/script/SignFromJson.s.sol#L31-L39) and one in [`NestedSignFromJson`](https://github.com/ethereum-optimism/superchain-ops/blob/808004cd933e0bfa36e6d9124701e296639a6d41/script/NestedSignFromJson.s.sol#L40-L50), we have access to the [`SimulationPayload`](https://github.com/base-org/contracts/blob/6df77b49e16c2da29d1530b624eb6df4ff00cb2e/script/universal/Simulator.sol#L19-L24) which should be all the info needed for the simulation (aside from Tenderly API auth credentials)

To see a sample Tenderly response you can make a request to their API for an existing transaction hash: `api.tenderly.co/api/v1/public-contract/{chainId}/trace/{txHash}`. No API key is needed for existing transactions, and the data returned should be a nearly identical format to the simulation API

alcueca:
Working on retrieving the domain and message hashes from a simulation in 704

alcueca:
#783 does this, a different PR will get it on CI

PR #704, opened 2025-03-02
Retrieve domain and message hashes from tenderly

alcueca:
**Description**

#679 and #195 both depend on being able to retrieve the domain and message hash from a task execution simulated on tenderly.

This PR makes MultisigTask.sol output to console the json payload necessary to create a simulation on tenderly. We write the console output to a temporary file where it can be picked by a script that extracts the payload, creates the simulation, and extracts the domain and message hash.

### DIFF
--- a/src/improvements/tasks/MultisigTask.sol
+++ b/src/improvements/tasks/MultisigTask.sol
@@ -685,7 +685,7 @@ abstract contract MultisigTask is Test, Script {
         // Add each storage override
         for (uint j = 0; j < stateOverrides[0].overrides.length; j++) {
             string memory comma = j < stateOverrides[0].overrides.length - 1 ? "," : "";
-            console.log("\"0x%s\":\"0x%s\"%s",
+            console.log("\"%s\":\"%s\"%s",
                 vm.toString(bytes32(stateOverrides[0].overrides[j].key)),
                 vm.toString(stateOverrides[0].overrides[j].value),
                 comma
--- a/src/improvements/tasks/MultisigTask.sol
+++ b/src/improvements/tasks/MultisigTask.sol
@@ -658,13 +658,43 @@ abstract contract MultisigTask is Test, Script {
 
     /// @notice print the tenderly simulation link with the state overrides
     function printTenderlySimulationLink(Action[] memory actions) internal view {
-        Simulation.StateOverride[] memory overrides = new Simulation.StateOverride[](1);
-        overrides[0] =
+        Simulation.StateOverride[] memory stateOverrides = new Simulation.StateOverride[](1);
+        stateOverrides[0] =
             Simulation.overrideSafeThresholdOwnerAndNonce(parentMultisig, msg.sender, _getNonce(parentMultisig));
         bytes memory txData = _execTransationCalldata(
             parentMultisig, getMulticall3Calldata(actions), Signatures.genPrevalidatedSignature(msg.sender)
         );
-        Simulation.logSimulationLink({_to: parentMultisig, _data: txData, _from: msg.sender, _overrides: overrides});
+        Simulation.logSimulationLink({_to: parentMultisig, _data: txData, _from: msg.sender, _overrides: stateOverrides});
+
+        // Log the simulation payload
+        console.log("\n\n------------------ Tenderly Simulation Payload ------------------");
+        
+        // Log the Tenderly JSON payload
+        string memory part1 = string.concat(
+            "{\"network_id\":\"", vm.toString(block.chainid),
+            "\",\"from\":\"", vm.toString(msg.sender),
+            "\",\"to\":\"", vm.toString(parentMultisig), "\""
+        );
+        string memory part2 = string.concat(
+            ",\"input\":\"", vm.toString(txData),
+            "\",\"value\":\"0x0\",\"state_objects\":{\"",
+            vm.toString(parentMultisig), "\":{\"storage\":{"
+        );
+        console.log("%s%s", part1, part2);
+        
+        // Add each storage override
+        for (uint j = 0; j < stateOverrides[0].overrides.length; j++) {
+            string memory comma = j < stateOverrides[0].overrides.length - 1 ? "," : "";
+            console.log("\"0x%s\":\"0x%s\"%s",
+                vm.toString(bytes32(stateOverrides[0].overrides[j].key)),
+                vm.toString(stateOverrides[0].overrides[j].value),
+                comma
+            );
+        }
+        
+        // Close the JSON structure
+        console.log("}}}}");
+        console.log("------------------ End of Payload ------------------");
     }
 
     /// @notice get the hash for this safe transaction
--- a/src/improvements/script/get-tenderly-trace.sh
+++ b/src/improvements/script/get-tenderly-trace.sh
@@ -1,0 +1,79 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Simple script to get a Tenderly trace from a simulation payload
+# Usage: ./get-tenderly-trace.sh payload.json
+
+# Check if a payload file was provided
+if [ $# -lt 1 ]; then
+  echo "Error: Payload file is required"
+  echo "Usage: $0 <payload_file>"
+  exit 1
+fi
+
+PAYLOAD_FILE="$1"
+
+# Check if the payload file exists
+if [ ! -f "$PAYLOAD_FILE" ]; then
+  echo "Error: Payload file not found: $PAYLOAD_FILE"
+  exit 1
+fi
+
+# Check for required environment variables
+if [ -z "${TENDERLY_ACCESS_TOKEN:-}" ]; then
+  echo "Error: TENDERLY_ACCESS_TOKEN environment variable is required"
+  exit 1
+fi
+
+if [ -z "${TENDERLY_USER:-}" ]; then
+  echo "Error: TENDERLY_USER environment variable is required"
+  exit 1
+fi
+
+if [ -z "${TENDERLY_PROJECT_SLUG:-}" ]; then
+  echo "Error: TENDERLY_PROJECT_SLUG environment variable is required"
+  exit 1
+fi
+
+# Read the payload from file
+PAYLOAD=$(cat "$PAYLOAD_FILE")
+
+# Call Tenderly simulation API
+RESPONSE=$(curl -s -X POST \
+  -H "Content-Type: application/json" \
+  -H "X-Access-Key: $TENDERLY_ACCESS_TOKEN" \
+  -d "$PAYLOAD" \
+  "https://api.tenderly.co/api/v1/account/$TENDERLY_USER/project/$TENDERLY_PROJECT_SLUG/simulate")
+
+# Check if the response contains an error
+if echo "$RESPONSE" | jq -e '.error' >/dev/null 2>&1; then
+  echo "Error from Tenderly API: $(echo "$RESPONSE" | jq -r '.error')" >&2
+  exit 1
+fi
+
+# Extract and output the response
+echo "$RESPONSE" | jq -r '.simulation'
+
+# Extract the simulation ID from the response
+SIMULATION_ID=$(echo "$RESPONSE" | jq -r '.simulation.id')
+
+# Output the Tenderly dashboard URL
+echo "View the simulation in Tenderly dashboard:"
+echo "https://dashboard.tenderly.co/$TENDERLY_USER/$TENDERLY_PROJECT_SLUG/simulator/$SIMULATION_ID"
+echo ""
+
+# Retrieve the full simulation details
+echo "Retrieving full simulation details..."
+FULL_SIMULATION=$(curl -s -H "X-Access-Key: $TENDERLY_ACCESS_TOKEN" \
+  "https://api.tenderly.co/api/v1/account/$TENDERLY_USER/project/$TENDERLY_PROJECT_SLUG/simulations/$SIMULATION_ID")
+
+# Check if there was an error retrieving the simulation
+if echo "$FULL_SIMULATION" | jq -e '.error' >/dev/null 2>&1; then
+  echo "Error retrieving simulation details: $(echo "$FULL_SIMULATION" | jq -r '.error')" >&2
+  exit 1
+fi
+
+# Output the full simulation details
+echo "Full simulation details:"
+echo "$FULL_SIMULATION" | jq '.'
+echo ""
--- a/src/improvements/script/simulate-task.sh
+++ b/src/improvements/script/simulate-task.sh
@@ -28,11 +28,23 @@ simulate_task() {
             echo "Error: this task requires a nested safe name e.g. foundation, council, chain-governor."
             exit 1
         fi
-        SIMULATE_WITHOUT_LEDGER=1 just --dotenv-path "$(pwd)"/.env --justfile "$nested_just_file" simulate "$nested_safe_name"
+        SIMULATE_WITHOUT_LEDGER=1 just --dotenv-path "$(pwd)"/.env --justfile "$nested_just_file" simulate "$nested_safe_name" | tee ./simulation_output.txt
     else
         echo "Simulating single task: $task"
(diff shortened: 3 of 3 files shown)


alcueca:
Right now I'm blocked because my simulations seem to be created, but can't be retrieved. Checking with Tenderly support on that.

I'm also not sure we can get the full trace of a simulation, it doesn't seem to be documented. Asking support about that as well.

alcueca:
`MultisigTask.sol` now logs out a json payload to simulate the transaction on tenderly, instead of a tenderly link.
`simulate-task.sh` parses the payload from the forge script output, and passes it to `get-tenderly-hashes`.
`get-tenderly-hashes` simulates the task on tenderly and retrieves the domain and message hashes from the trace, showing them on the console.
For this to run, the `TENDERLY_ACCESS_TOKEN`, `TENDERLY_USER` and `TENDERLY_PROJECT_SLUG` environment variables need to be set up. The latter two could be hardcoded, but the first one needs to go into the CI secrets.

Full output of a task simulation below.

```
Task: /Users/alcueca/Code/OpLabs/superchain-ops/src/improvements/tasks/example/eth/003-transfer-owner-template
Is nested: true
Simulating nested task: /Users/alcueca/Code/OpLabs/superchain-ops/src/improvements/tasks/example/eth/003-transfer-owner-template
RPC URL: mainnet
Using script /Users/alcueca/Code/OpLabs/superchain-ops/src/improvements/template/TransferOwnerTemplate.sol
getting signer address for foundation...
safe: 0x847B5c174615B1B7fDF770882256e2D3E95b9D92
Simulating call to foundation at 0x847B5c174615B1B7fDF770882256e2D3E95b9D92
Simulating without ledger using the first owner account: 0x42d27eEA1AD6e22Af6284F609847CB3Cd56B9c64

No files changed, compilation skipped
No files changed, compilation skipped
Script ran successfully.

== Logs ==
  Running assertions on the state diff
  
------------------ Task Actions ------------------
  1). Call OP_MAINNET_ProxyAdmin @0x543bA4AADBAb8f9025686Bd03993043599c6fB04 with 0 eth and 0xf2fde38b0000000000000000000000005a0aae59d09fccbddb6c6cceb07b7279367c3d2a data.
  target: OP_MAINNET_ProxyAdmin @0x543bA4AADBAb8f9025686Bd03993043599c6fB04
payload
  0xf2fde38b0000000000000000000000005a0aae59d09fccbddb6c6cceb07b7279367c3d2a
  

  
----------------- Task Transfers -------------------
  No ETH or ERC20 transfers.
  
----------------- Task State Changes -------------------
  
----- DecodedStateDiff[0] -----
  Who:               0x5a0Aae59D09fccBdDb6C6CcEB07B7279367C3d2A
  Contract:          ProxyAdminOwner (GnosisSafe)
  Chain ID:          10
  Raw Slot:          0x0000000000000000000000000000000000000000000000000000000000000005
  Raw Old Value:     0x000000000000000000000000000000000000000000000000000000000000000a
  Raw New Value:     0x000000000000000000000000000000000000000000000000000000000000000b
  Decoded Kind:      uint256
  Decoded Old Value: 10
  Decoded New Value: 11
  Summary:           nonce
  Detail:            
  

------------------ Task Calldata ------------------
  0x174dea71000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000020000000000000000000000000543ba4aadbab8f9025686bd03993043599c6fb040000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000800000000000000000000000000000000000000000000000000000000000000024f2fde38b0000000000000000000000005a0aae59d09fccbddb6c6cceb07b7279367c3d2a00000000000000000000000000000000000000000000000000000000
  

------------------ Nested Multisig EOAs Data to Sign ------------------
  Child multisig: OP_MAINNET_FoundationUpgradeSafe @0x847B5c174615B1B7fDF770882256e2D3E95b9D92
  vvvvvvvv
  0x1901a4a9c312badf3fcaa05eafe5dc9bee8bd9316c78ee8b0bebe3115bb21b7326720cf4726b9f9362539e93299497f0265649e3abd184da2d5bf6a7bf6ee9a73513
  ^^^^^^^^

  

------------------ Nested Multisig EOAs Hash to Approve ------------------
  Nested multisig: OP_MAINNET_FoundationUpgradeSafe @0x847B5c174615B1B7fDF770882256e2D3E95b9D92
  0xdaa7b261fa888e59717d131cb373ffcbfb3e0a87cd6336c56f2c8003d65208eb
  

------------------ Tenderly Simulation Payload ------------------
  {"network_id":"1","from":"0x42d27eEA1AD6e22Af6284F609847CB3Cd56B9c64","to":"0x5a0Aae59D09fccBdDb6C6CcEB07B7279367C3d2A","save":true,"input":"0x6a761202000000000000000000000000ca11bde05977b3631167028862be2a173976ca110000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000014000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002c00000000000000000000000000000000000000000000000000000000000000144174dea71000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000020000000000000000000000000543ba4aadbab8f9025686bd03993043599c6fb040000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000800000000000000000000000000000000000000000000000000000000000000024f2fde38b0000000000000000000000005a0aae59d09fccbddb6c6cceb07b7279367c3d2a0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000004100000000000000000000000042d27eea1ad6e22af6284f609847cb3cd56b9c6400000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000","value":"0x0","state_objects":{"0x5a0Aae59D09fccBdDb6C6CcEB07B7279367C3d2A":{"storage":{
  "0x0000000000000000000000000000000000000000000000000000000000000004":"0x0000000000000000000000000000000000000000000000000000000000000001",
  "0x0000000000000000000000000000000000000000000000000000000000000003":"0x0000000000000000000000000000000000000000000000000000000000000001",
  "0xe90b7bceb6e7df5418fb78d8ee546e97c83a08bbccc01a0644d599ccd2a7c2e0":"0x00000000000000000000000042d27eea1ad6e22af6284f609847cb3cd56b9c64",
  "0xf910e665d88aa8ac9ceb7eb15ee995b4540a315cb9e04ad419161ca8aa36d15d":"0x0000000000000000000000000000000000000000000000000000000000000001",
  "0x0000000000000000000000000000000000000000000000000000000000000005":"0x000000000000000000000000000000000000000000000000000000000000000a"
  }}}}
  ------------------ End of Payload ------------------


Tenderly simulation payload saved to: ./tenderly_payload.json
View the simulation in Tenderly dashboard:
https://dashboard.tenderly.co/alberto-oplabs/project/simulator/a4ef411a-122c-4106-960b-658c375092ce


------------------ Domain and Message Hashes ------------------
0x1901daf670b31fdf41fdaae2643ed0ebe709283539c0e61540c160b5a6403d79073f02c8403dd8a8fc9971ffcc9a913dc6270fcf10d78371438240e3be60009be4d8


Done simulating task: /Users/alcueca/Code/OpLabs/superchain-ops/src/improvements/tasks/example/eth/003-transfer-owner-template
```

alcueca:
Also, I would like an exception on `forge fmt` to maintain the json building code lines in MultisigTask.sol sane.

src/improvements/tasks/MultisigTask.sol
```diff
@@ -656,15 +656,41 @@ abstract contract MultisigTask is Test, Script {
...
+        Simulation.StateOverride[] memory stateOverrides = new Simulation.StateOverride[](1);
+        stateOverrides[0] =
             Simulation.overrideSafeThresholdOwnerAndNonce(parentMultisig, msg.sender, _getNonce(parentMultisig));
         bytes memory txData = _execTransationCalldata(
             parentMultisig, getMulticall3Calldata(actions), Signatures.genPrevalidatedSignature(msg.sender)
         );
-        Simulation.logSimulationLink({_to: parentMultisig, _data: txData, _from: msg.sender, _overrides: overrides});
+        
+        // Log the Tenderly JSON payload
+        string memory payload = string.concat(
```
mds1, line 669: > Also, I would like an exception on `forge fmt` to maintain the json building code lines in MultisigTask.sol sane.

You can use `// forgefmt: disable-start` and `// forgefmt: disable-end` as needed

smartcontracts:
Are we also computing these hashes ourselves? Having a cross-check against Tenderly seems like a good idea but definitely should continue computing it ourselves too

alcueca:
I don't think we were calculating the hashes ourselves, so I did a forge script to do that and run it just before the tenderly simulation so that the outputs can be compared.

```
...
------------------ Tenderly Simulation Payload ------------------
  ...
  ------------------ End of Payload ------------------


Tenderly simulation payload saved to: ./tenderly_payload.json
[⠊] Compiling...
No files changed, compilation skipped
Script ran successfully.
Gas used: 333709

== Logs ==
  

-------- Domain Separator and Message Hashes from Payload --------
  Domain Separator:  0xa4a9c312badf3fcaa05eafe5dc9bee8bd9316c78ee8b0bebe3115bb21b732672
  Message Hash:  0xeaa4f6145c92c0cf0973f3fd404cc219a8ff40c5d4f875a748ffa0f45959cc56


-------- Domain Separator and Message Hashes from Tenderly --------
  Domain Separator: 0xa4a9c312badf3fcaa05eafe5dc9bee8bd9316c78ee8b0bebe3115bb21b732672
  Message Hash: 0xeaa4f6145c92c0cf0973f3fd404cc219a8ff40c5d4f875a748ffa0f45959cc56

View the simulation in Tenderly dashboard:
https://dashboard.tenderly.co/alberto-oplabs/project/simulator/940c2ee8-5a23-4d81-934b-e46cbaa70cc0


Done simulating task: /Users/alcueca/Code/OpLabs/superchain-ops/src/improvements/tasks/example/eth/000-gas-config-template
```

alcueca:
Example of hash mismatch:

image

alcueca:
Strangely, now that I'm getting CI to submit simulations to tenderly, the computed domain name is different:

image

alcueca:
It fails on `src/improvements/tasks/example/eth/002-set-game-type-template`. The previous tasks are one single and one nested so that's not the issue.

alcueca:
I think that this PR is generally fine in terms of UX, but I wouldn't mind some feedback.

My concern is that I'm making `just simulate-all-templates` heavier with a dependency on tenderly. That's probably fine for CI, but now task developers should simulate individual tasks instead if they wish to avoid the tenderly integration. They probably simulate individual tasks anyway, so I guess that's fine. They will automatically obtain the hashes and the simulation, so their job of verifying them and adding them to VALIDATIONS.md is easier.

We probably need a just command for signers to simulate a single task with tenderly verification. Signers will now need to obtain an access token from tenderly, and set the `TENDERLY_USER`, `TENDERLY_PROJECT` and `TENDERLY_ACCESS_TOKEN`. I guess that's a fair trade-off.

Having these hash comparisons in CI is probably somewhat useful, although it will be more useful in CI to have them checked against VALIDATIONS.md

Ethnical:
>We probably need a just command for signers to simulate a single task with tenderly verification. Signers will now need to obtain an access token from tenderly, and set the TENDERLY_USER, TENDERLY_PROJECT and TENDERLY_ACCESS_TOKEN. I guess that's a fair trade-off.

Seems to be a leap here, why we don't want to use the tenderly link simulation here like in the past? @alcueca ?
I am not a fan to add for all signers a new secrets or any env var to be honest.

alcueca:
From a chat with @Ethnical:

I'm going to implement this feature to be used by signers only, using a shared access token which will be kept in a vault.

Then #679 will reuse the same code to check the hashes in VALIDATIONS.md against tenderly.

alcueca:
The only thing I'm leaving as a TODO is to add to AddressRegistry a mapping for contracts with non-standard domain separators, and checking against it before calling `GnosisSafeHashes.domainSeparator`

src/libraries/GnosisSafeHashes.sol
```diff
@@ -0,0 +1,140 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.15;
+
+import "forge-std/console.sol";
+import {GnosisSafe} from "lib/safe-contracts/contracts/GnosisSafe.sol";
```
smartcontracts, line 5: I think this dependency is unused

src/libraries/GnosisSafeHashes.sol
```diff
@@ -0,0 +1,140 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.15;
+
+import "forge-std/console.sol";
```
smartcontracts, line 4: Appears to be unused

src/libraries/GnosisSafeHashes.sol
```diff
@@ -0,0 +1,140 @@
+// SPDX-License-Identifier: UNLICENSED
```
smartcontracts, line 1:
```suggestion
// SPDX-License-Identifier: MIT
```

src/improvements/tasks/MultisigTask.sol
```diff
@@ -656,15 +657,69 @@ abstract contract MultisigTask is Test, Script {
...
+        // Log the Tenderly JSON payload
+        console.log("\nSimulation payload:");
+        logTenderlySimulationPayload(txData, stateOverrides[0].overrides);
+
+        // Log the simulation link
+        console.log("\nSimulation link:");
+        Simulation.logSimulationLink({_to: parentMultisig, _data: txData, _from: msg.sender, _overrides: stateOverrides});
+
+        // Calculate domain separator
+        // TODO: Read from AddressRegistry for some addresses that have non compliant domain separators, such as the mainnet Foundation Operations Safe
```
smartcontracts, line 678: Alternative option here is just to pass in the Safe version. Versions < 1.3.0 have a different domain separator, everything else uses the newer separator that includes chain ID.

src/improvements/justfile
```diff
@@ -130,3 +130,17 @@ simulate-all-templates:
         echo "Not all templates files have a corresponding task. Please add a task for a template in ${root_dir}/src/improvements/tasks/example" >&2
         exit 1
     fi
+
+simulate-verify-task TASK="":
```
smartcontracts, line 134: What's the expected usage here? Is this something we want all signers to execute?

smartcontracts (commented):
Generally looking good, left a few comments!

src/improvements/tasks/MultisigTask.sol
alcueca: That was a bit of a side quest, but I came up with a minimal implementation to check whether the Safe version is below 1.3.0

src/improvements/justfile
alcueca: Yes, that is the expected use case, should I go into modifying `single.just` and `nested.just` along with the documentation?

Or should we merge this as internal use in `src/improvements` and work on the process in a later PR?

alcueca: I'll leave it as is now, and we refactor the signing flow in a separate PR

Ethnical:
Looks cool one thing, I think we should create a README here such as the same for simulate_seq: 
https://github.com/ethereum-optimism/superchain-ops/blob/main/SIMULATE_SEQ.md

Ethnical (approved):
Re-Approve since the merge dismissed the previous one.

PR #708, opened 2025-03-03
feat: SimpleAddressRegistry

mds1:
- Adds `SimpleAddressRegistry` and tests. See the fixtures for how to use this, basically you just list key-value pairs under an `[addresses]` section in the config file
- Adds the same functionality to `SuperchainAddressRegistry`
- We currently have mocks that are used for tests in our the `test/mock/configs` dir. These seem more like fixtures to me since they are for test setup and not always realistic, so I've added fixtures for these tests in `tests/fixtures` and in the future we can move the other ones there as well


src/improvements/SuperchainAddressRegistry.sol
```diff
@@ -78,13 +83,8 @@ contract SuperchainAddressRegistry is StdChains {
             string.concat("SuperchainAddressRegistry: Unsupported task chain ID ", vm.toString(block.chainid))
         );
 
-        bytes memory chainListContent;
-        require(gasleft() > 500_000, "insufficient gas for readFile() call"); // EIP-150 safe.
-        try vm.parseToml(vm.readFile(configPath), ".l2chains") returns (bytes memory parsedChainListContent) {
-            chainListContent = parsedChainListContent;
-        } catch {
-            revert(string.concat("SuperchainAddressRegistry: Failed to parse network config file path ", configPath));
-        }
```
mds1, line 87: This try/catch was unnecessary, since forge reverts if there is a problem, so I removed it to simplify things

src/improvements/SuperchainAddressRegistry.sol
```diff
@@ -137,8 +147,19 @@ contract SuperchainAddressRegistry is StdChains {
...
+            registry[identifier][chain.chainId] == address(0),
+            string.concat(
+                "SuperchainAddressRegistry: duplicate key ", identifier, "for chain ", vm.toString(chain.chainId)
+            )
+        );
+        // TODO We should be able to have this check, but currently tests revert if uncommented.
+        // require(
+        //     bytes(addressInfo[addr].identifier).length == 0,
+        //     string.concat("SuperchainAddressRegistry: address already registered for ", vm.toString(addr))
+        // );
```
mds1, line 162: It seems the existing addresses like `FoundationUpgrades` safe get added for each chain, but instead they should just be added once under the sentinel key. I think the root `addresses.toml` should get read in during construction of both address registries and saved off. For this contract we'd save it with the sentinel key, and need to make sure to save off the right section based off `block.chainid`. But this can come in a future PR

src/improvements/SuperchainAddressRegistry.sol
```diff
@@ -151,25 +172,38 @@ contract SuperchainAddressRegistry is StdChains {
     }
 
     /// @notice Retrieves an address by its identifier for a specified L2 chain
+    /// This is deprecated in favor of the `get` function.
```
mds1, line 175: We know it returns an address from the return type, and since we'll use `registry.get(...)` a lot in tasks I figured we should just shorten the interface here. Similar to how we only have `get` in the simple address registry also

blmalone (approved):

PR #716, opened 2025-03-05
Load hardcoded addresses simple address registry

prat-gpt:
**Description**
- Add warning in `getAddressInfo` method that `there might be multiple infos for the same address, this function
    will return the last info that was saved for the _who address`. [Ref](https://discord.com/channels/1244729134312198194/1345058837748777080/1346484963410710579).
- load hardcoded addresses once in `SuperchainAddressRegistry` corresponding to the sentinel chain.
- load hardcoded addresses in `SimpleAddressRegistry`
- Added a hack in `L2TaskBase` to support `EnableDeputyPauseModuleTemplate`. Also added todos to remove this hack once we have `SimpleTaskBase` contract and `EnableDeputyPauseModuleTemplate` is derived from `SimpleTaskBase` rather than `L2TaskBase`.

**Metadata**
This PR is based on this comment https://github.com/ethereum-optimism/superchain-ops/pull/708#discussion_r1977903596


src/improvements/SimpleAddressRegistry.sol
```diff
@@ -29,15 +30,16 @@ contract SimpleAddressRegistry {
...
-            keys[who] = key;
-            vm.label(who, key);
+            _registerAddress(key, who);
         }
+
+        string memory chainKey;
+        if (block.chainid == getChain("mainnet").chainId) chainKey = ".eth";
+        else if (block.chainid == getChain("sepolia").chainId) chainKey = ".sep";
+        else if (block.chainid == getChain("optimism").chainId) chainKey = ".oeth";
+        else revert(string.concat("SimpleAddressRegistry: Unknown task chain ID ", vm.toString(block.chainid)));
```
mds1, line 40: Why do we have to limit the chains this supports? We should be able to delete this and allow this to be used on any chain

src/improvements/tasks/MultisigTask.sol
```diff
@@ -1065,32 +1065,50 @@ abstract contract L2TaskBase is MultisigTask {
...
         // or figure out why the storage slots are being changed when they should not be.
         for (uint256 i = 0; i < config.allowedStorageWriteAccesses.length; i++) {
-            for (uint256 j = 0; j < chains.length; j++) {
-                _allowedStorageAccesses.add(
-                    superchainAddrRegistry.getAddress(config.allowedStorageWriteAccesses[i], chains[j].chainId)
-                );
+            // TODO: This is a hack to support the EnableDeputyPauseModuleTemplate.
+            // remove this once we have SimpleTaskBase contract as then this template will derive from SimpleTaskBase
+            // instead of L2TaskBase.
+            if (keccak256(bytes(config.allowedStorageWriteAccesses[i])) == keccak256(bytes("FoundationOperationSafe")))
```
mds1, line 1103: Whenever we check string equality we should just use solady’s LibString and do `someString.eq(otherString)`

mds1: But since these checks are a temporary hack and will be deleted it’s ok to leave them

test/registry/SimpleAddressRegistry.t.sol
```diff
@@ -5,32 +5,35 @@ import {Test} from "forge-std/Test.sol";
...
     address constant charlie = address(3);
 
-    // Path to test fixture files
-    string constant FIXTURES_PATH = "test/fixtures/SimpleAddressRegistry/";
-
     string registryName; // Contract name being tested.
     string idReturnKind; // "identifier" or "AddressInfo"
 
     function setUp() public virtual {
+        vm.createSelectFork("mainnet");
```
mds1, line 32: Why does this need to become a fork test?

test/registry/SimpleAddressRegistry.t.sol
```diff
@@ -91,3 +94,37 @@ contract SimpleAddressRegistryTest is Test {
         registry.get(address(5));
     }
 }
+
+contract SimpleAddressRegistryTest_HardcodedAddresses is Helper, Test {
```
mds1, line 98: Why do we need this as a separate contract that inherits from Helper? Can these tests just be in the original, single test file?

src/improvements/SimpleAddressRegistry.sol
prat-gpt: I remembered from our discussion in the meeting that it will be either Mainnet or OP mainnet. So I restricted it. I thought we should allow only the chains where we know tasks will be executed.

src/improvements/SimpleAddressRegistry.sol
```diff
@@ -54,4 +56,30 @@ contract SimpleAddressRegistry {
             string.concat("SimpleAddressRegistry: identifier not found for ", vm.toString(_who))
         );
     }
+
+    /// @notice Reads in hardcoded addresses from the addresses.toml file and registers them.
+    function _loadHardcodedAddresses(string memory chainKey) internal {
+        string memory toml = vm.readFile("./src/improvements/addresses.toml");
+        string[] memory keyStrings = vm.parseTomlKeys(toml, chainKey);
+        require(keyStrings.length > 0, string.concat("SimpleAddressRegistry: no keys found for ", chainKey));
```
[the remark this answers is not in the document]
prat-gpt: agree, will remove.

src/improvements/SimpleAddressRegistry.sol
mds1: Yea in practice that will likely be the set of chains, but if this check is just an artificial restriction we should remove it since it doesn’t add value. It also lets us write faster tests since we don’t need to fork

test/registry/SimpleAddressRegistry.t.sol
prat-gpt: Which hardcoded addresses to load are dependent on chain id and there is a revert when chain id is anything other than mainnet, sepolia or op mainnet. If that check is not needed we can remove this as well.

mds1: You can also use cheat codes to directly change the chain ID without forking

test/registry/SimpleAddressRegistry.t.sol
prat-gpt: I created a separate contract to avoid switching forks. But with this [check](https://github.com/ethereum-optimism/superchain-ops/pull/716#discussion_r1981389575) removed there is only one test where this switch will occur. Updated the tests to be in the original single test file.

test/registry/SuperchainAddressRegistry.t.sol
```diff
@@ -158,29 +158,6 @@ abstract contract SuperchainAddressRegistryTest_Base is Test {
             assertNotEq(addrRegistry.getAddress("ProxyAdminOwner", chainId), address(0), "320");
             assertNotEq(addrRegistry.getAddress("SystemConfigOwner", chainId), address(0), "330");
             assertNotEq(addrRegistry.getAddress("UnsafeBlockSigner", chainId), address(0), "340");
-
-            // Define expected superchain auth addresses for mainnet and testnet.
-            address fus = block.chainid == 1
```
blmalone, line 163: Should we have a test case in `SuperchainAddressRegistry.t.sol` that checks `addresses.toml` file addresses are added correctly?

test/registry/SimpleAddressRegistry.t.sol
```diff
@@ -4,31 +4,37 @@ pragma solidity ^0.8.0;
...
+    function _deployRegistry(string memory configFile) internal virtual returns (address) {
+        return address(new SimpleAddressRegistry(_getPath(configFile)));
+    }
+}
+
+contract SimpleAddressRegistryTest is Helper, Test {
     // Test addresses
     address constant alice = address(1);
     address constant bob = address(2);
     address constant charlie = address(3);
```
blmalone, line 27: This variable isn't used.

test/registry/SimpleAddressRegistry.t.sol
```diff
@@ -4,31 +4,37 @@ pragma solidity ^0.8.0;
...
-    // Path to test fixture files
-    string constant FIXTURES_PATH = "test/fixtures/SimpleAddressRegistry/";
-
     string registryName; // Contract name being tested.
     string idReturnKind; // "identifier" or "AddressInfo"
 
+    bool isSimpleAddressRegistry;
+
     function setUp() public virtual {
         registryName = "SimpleAddressRegistry";
```
blmalone, line 35: Doesn't this file only test `SimpleAddressRegistry`?

test/registry/SimpleAddressRegistry.t.sol
```diff
@@ -90,4 +96,38 @@ contract SimpleAddressRegistryTest is Test {
...
+    function test_hardcodedAddresses_sepolia() public {
+        vm.createSelectFork("sepolia");
+        SimpleAddressRegistry registry = SimpleAddressRegistry(_deployRegistry("valid_addresses_sepolia.toml"));
+        assertEq(registry.get("FoundationOperationSafe"), 0x837DE453AD5F21E89771e3c06239d8236c0EFd5E, "10");
+        assertEq(registry.get("FoundationUpgradeSafe"), 0xDEe57160aAfCF04c34C887B5962D0a69676d3C8B, "20");
+        assertEq(registry.get("SecurityCouncil"), 0xf64bc17485f0B4Ea5F06A96514182FC4cB561977, "30");
+    }
+
+    function test_hardcodedAddresses_opMainnet() public {
+        if (!isSimpleAddressRegistry) {
```
blmalone, line 118: This never gets skipped right?

src/improvements/addresses.toml
```diff
@@ -8,3 +8,5 @@ SecurityCouncil = "0xc2819DC788505Aac350142A7A707BF9D03E3Bd03"
 FoundationOperationSafe = "0x837DE453AD5F21E89771e3c06239d8236c0EFd5E"
 FoundationUpgradeSafe = "0xDEe57160aAfCF04c34C887B5962D0a69676d3C8B"
 SecurityCouncil = "0xf64bc17485f0B4Ea5F06A96514182FC4cB561977"
+
+[oeth]
```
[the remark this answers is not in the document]
prat-gpt: yes. as we don't know what addresses to hardcode for op mainnet.

src/improvements/SuperchainAddressRegistry.sol
```diff
@@ -155,11 +155,6 @@ contract SuperchainAddressRegistry is StdChains {
                 "SuperchainAddressRegistry: duplicate key ", identifier, "for chain ", vm.toString(chain.chainId)
             )
         );
-        // TODO We should be able to have this check, but currently tests revert if uncommented.
```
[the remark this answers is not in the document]
prat-gpt: we cannot have this check. please refer this discussion [here](https://discord.com/channels/1244729134312198194/1345058837748777080/1346479359342547016).

src/improvements/SuperchainAddressRegistry.sol
```diff
@@ -200,6 +195,8 @@ contract SuperchainAddressRegistry is StdChains {
     }
 
     /// @notice Retrieves the identifier and chain info for a given address.
+    /// There might be multiple infos for the same address, this function
```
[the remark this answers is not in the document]
prat-gpt: yes. please refer [here](https://discord.com/channels/1244729134312198194/1345058837748777080/1346479359342547016).

test/registry/SuperchainAddressRegistry.t.sol
prat-gpt: the test is already added in `test_hardcodedAddresses_mainnet`, `test_hardcodedAddresses_sepolia` and `test_hardcodedAddresses_opMainnet`. There is a contract `SuperchainAddressRegistryTest_Addresses` that inherits `SimpleAddressRegistryTest` and that's how same tests run for both simple and superchain registry.

test/registry/SimpleAddressRegistry.t.sol
prat-gpt: It also test `SuperchainAddressRegistry` using `SuperchainAddressRegistryTest_Addresses` contract.

test/registry/SimpleAddressRegistry.t.sol
prat-gpt: skipped when run from `SuperchainAddressRegistryTest_Addresses` as there is no addresses discovert possible for opmainnet.

test/registry/SimpleAddressRegistry.t.sol
prat-gpt: updated.

foundry.toml
```diff
@@ -33,3 +33,4 @@ sepolia = "https://ci-sepolia-l1-archive.optimism.io" # Must have 'circleci_ip_r
 localhost = "http://127.0.0.1:8545"
 mainnet = "https://ethereum.publicnode.com"
 sepolia = "https://gateway.tenderly.co/public/sepolia"
+opMainnet = "https://mainnet.optimism.io"
```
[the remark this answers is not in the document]
prat-gpt: updated.

src/improvements/SimpleAddressRegistry.sol
blmalone: I'm curious if the same applies for SuperchainAddressRegistry? Conceptually having only sepolia and mainnet makes sense though.

blmalone (approved):

test/registry/SimpleAddressRegistry.t.sol
prat-gpt: i am using `vm.chainId` now.

Issue #726, opened 2025-03-05
Add state override functionality to superchain-ops 2.0


blmalone:
For feature parity with the old superchain-ops system, tasks need a way to be able to override state. This is commonly used by task developers to override the nonce of the parent multisig (see [here](https://github.com/base/contracts/blob/bab87b7927c5c77ba65e83cdc36610afa1afee95/script/universal/MultisigBase.sol#L38-L59)). 

However, task developers should be able to define _any_ state overrides, not limited to nonces. The base contracts [Simulation.sol](https://github.com/base/contracts/blob/bab87b7927c5c77ba65e83cdc36610afa1afee95/script/universal/Simulation.sol) library has helper functions to assist with this.

### Requirements for OPCM
- Override nonces
- Override thresholds
- Override implementation addresses on superchain level proxies e.g. this [line](https://github.com/ethereum-optimism/optimism/blob/op-contracts/v2.0.0-rc.1/packages/contracts-bedrock/src/L1/OPContractsManager.sol#L478) will revert if we're doing a simulation where we are upgrading Base _before_ OP has updated (Base doesn't use upgradeController address). OPCM enforces ordering in that OP mainnet/sepolia has to execute upgrade first before any other chain.

blmalone:
This issue was completed via: https://github.com/ethereum-optimism/superchain-ops/pull/727

PR #727, opened 2025-03-05
feat: Generic State Overrides

blmalone:
As per this [issue](https://github.com/ethereum-optimism/superchain-ops/issues/726), this PR implements state override functionality for MultisigTask. I created a `StateOverrideManager` (inspired by the [Manager pattern common in GnosisSafe](https://github.com/safe-global/safe-smart-account/blob/main/contracts/Safe.sol#L39-L46)). MultisigTask inherits this contract. Any changes to the `_stateOverrides` state variable is contained within this file. All task types (`L2TaskBase`/`StandaloneTaskBase`) will have access to the functionality contained within the `StateOverrideManager`.

Below is an example of what state overrides for a single contract looks like in `config.toml`.
```
[stateOverrides]
0x5a0Aae59D09fccBdDb6C6CcEB07B7279367C3d2A = [{slot = "0x0000000000000000000000000000000000000000000000000000000000000005", value = "0x0000000000000000000000000000000000000000000000000000000000000FFF"}]
```


src/improvements/tasks/MultisigTask.sol
```diff
@@ -1050,9 +1050,64 @@ abstract contract MultisigTask is Test, Script {
...
+
+    struct StateOverride {
+        address target;
+        KeyValueOverride keyValueOverride;
+    }
+
+    struct KeyValueOverride {
+        bytes32 slot;
+        bytes32 value;
+    }
```
maurelian, line 1063: nit, but it might make sense to just call this `StorageOverride` like in [base-contracts](https://github.com/ethereum-optimism/base-org-contracts/blob/7cca28be0f94134f9de1c6f409779b57d5f269b7/script/universal/Simulation.sol#L19-L22), to minimize differences.

src/improvements/tasks/MultisigTask.sol
```diff
@@ -1050,9 +1050,64 @@ abstract contract MultisigTask is Test, Script {
 
 abstract contract L2TaskBase is MultisigTask {
     using EnumerableSet for EnumerableSet.AddressSet;
+    using stdToml for string;
+
+    struct StateOverride {
+        address target;
+        KeyValueOverride keyValueOverride;
```
maurelian, line 1057: Would it be difficult to make this an array, similar to how it's handled in [base-contracts](https://github.com/ethereum-optimism/base-org-contracts/blob/7cca28be0f94134f9de1c6f409779b57d5f269b7/script/universal/Simulation.sol#L16)? 

```suggestion
        KeyValueOverride[] keyValueOverrides;
```

If it's not easy, then I think we can skip it for upgrade 13, but we'll very likely need it eventually.

maurelian: Actually, I just noticed that we're still using base's `Simulation.sol` in this file, and there's even a [reference](https://github.com/ethereum-optimism/superchain-ops/blob/646559e5c260a03bc22fd5d034c17d6cf142d1ae/src/improvements/tasks/MultisigTask.sol#L663C59-L663C83) to the `Simulation.StateOverride` type in `printTenderlySimulationLink()`. 

Can we: 
- just use those types
- parse the toml into those types
- modify `printTenderlySimulationLink()` to accept the overrides as an argument or read them from a state variable where they get pushed to?

test/tasks/MultisigTask.t.sol
```diff
@@ -243,14 +243,84 @@ contract MultisigTaskUnitTest is Test {
...
     }
 
-    function testNonceStateOverrideApplied() public {
-        string memory taskConfigFilePath = "test/tasks/mock/configs/NonceStateOverrideConfig.toml";
-        runTestSimulation(taskConfigFilePath);
+    string constant commonToml = "l2chains = [{name = \"OP Mainnet\", chainId = 10}]\n" "\n"
+        "templateName = \"DisputeGameUpgradeTemplate\"\n" "\n"
+        "implementations = [{gameType = 0, implementation = \"0xf691F8A6d908B58C534B624cF16495b491E633BA\", l2ChainId = 10}]\n";
+
+    function testNonceAndThresholdStateOverrideApplied() public {
```
blmalone, line 250: Using inline testing to reduce bloat by creating more tests and possibility of interference by other tests using the same config.toml file.

src/improvements/tasks/MultisigTask.sol
blmalone: good call: https://github.com/ethereum-optimism/superchain-ops/pull/727/commits/007e2702f8b187925c3b54aed091c391f1c57985

src/improvements/tasks/MultisigTask.sol
blmalone: I made it an array: https://github.com/ethereum-optimism/superchain-ops/pull/727/commits/007e2702f8b187925c3b54aed091c391f1c57985 

Right now, in the `config.toml` file the UX is a little janky because users have to supply well-formed (padded) bytes32 hex strings. When users supply them correctly it works great. When they don't they're not parsed correctly and it leads to errors.

blmalone: I've added this functionality to the `printTenderlySimulationLink` function. I noticed that we had some default state overrides being applied (threshold, nonce, owner mapping) already. I had to make sure these still get applied but if the user added the same slots to their config, the users config would take preference. I then combin the arrays. Let me know what you think.

maurelian: > I noticed that we had some default state overrides being applied (threshold, nonce, owner mapping) already. 

Is that something we can easily add a test for?

blmalone: I've added a bunch of tests for this. The default state overrides where being created and then inserted into the tenderly URL, they aren't being applied to the simulation (unless the user specifies them).

blmalone: I've refactored this PR quite a bit since this discussion.

src/improvements/tasks/StateOverrideManager.sol
```diff
@@ -0,0 +1,179 @@
...
+                parentMultisigIndex = i;
+                break;
+            }
+        }
+        return parentMultisigIndex;
+    }
+
+    /// @notice Combines the default Tenderly overrides with the existing parent multisig overrides.
+    /// The parent multisig overrides will take precedence over the default Tenderly overrides.
+    function _createCombinedOverrides(uint256 parentMultisigIndex, address parentMultisig, uint256 parentMultisigNonce)
```
blmalone, line 122: Open to better ways to combine the the two overrides arrays for the parent multisig.

prat-gpt: I think we don't need to create the combined override. we can simply add default tenderly state override at index 0 and rest of the overrides after index 0. This way when all the state overrides are applied one by one, automatically the parent multisig overrides will take precedence over the default Tenderly override. `base-contracts` lib does something similar [here](https://github.com/base/contracts/blob/7cca28be0f94134f9de1c6f409779b57d5f269b7/script/universal/MultisigBuilder.sol#L157-L158).

src/improvements/tasks/MultisigTask.sol
```diff
@@ -1067,6 +1076,9 @@ abstract contract L2TaskBase is MultisigTask {
         SuperchainAddressRegistry.ChainInfo[] memory chains = superchainAddrRegistry.getChains();
         parentMultisig_ = IGnosisSafe(superchainAddrRegistry.getAddress(config.safeAddressString, chains[0].chainId));
 
+        _applyStateOverrides(taskConfigFilePath);
```
prat-gpt, line 1079: `_configureTask()` is called in `_taskSetup()` which is called in `simulateRun()` as well as `executeRun()`. I think there should be no state overrides for `executeRun()`. So we should create a new function `mock()` which is added in `simulateRun()` after `_taskSetup()` and before `build()`. This function can then be overridden to apply state overrides. This way state overrides will be applied only when simulating a task. @blmalone

src/improvements/tasks/StateOverrideManager.sol
blmalone: Nice, I've added this feature in this commit: https://github.com/ethereum-optimism/superchain-ops/pull/727/commits/bcd23d136df2fb66337bb52e77d3b3f17790d673

src/improvements/tasks/MultisigTask.sol
blmalone: https://github.com/ethereum-optimism/superchain-ops/pull/727/commits/82ed821f90806d49002bb135889f043526a65dd0

maurelian (approved):

PR #736, opened 2025-03-11
Compare hashes obtained from local scripts and tenderly simulations against the ones in VALIDATIONS.md

### LINKED ISSUE #679: Check domain hash from simulation matches VALIDATIONS.md hash entry.

alcueca:
**Description**

Fixes #679, more specifically:
1. Signer flows now accept an extra optional parameter to verify message hash and domain separator against a toml block in VALIDATIONS.md. Options:
   - `verify-local` uses a local forge script
   - `verify-tenderly` uses tenderly, but requires an access token
   - `verify-all` uses both methods in sequence
   - `skip-verify` does no verification
 
The default in both single and nested flows is to verify locally so that we don't break CI and flows everywhere with tenderly access token requirements, which can be changed later.

2. There is now a `just simulate-non-terminal-tasks` command which does `verify-all` on all tasks inside `test/tasks/example`. This will be changed to exclude `example` and target real tasks once there are some of those to test CI with.

**Test**
A tenderly access token for `task-simulation` is available in the [1password vault](https://oplabs.1password.com/app#/m3eh5cvkwq4hlzcsmcdfqogdme/AllItems).

```
cd src/improvements.
just simulate-non-terminal-tasks     <-- Needs access token, local and tenderly verification
just simulate-all-templates.              <-- You'll see verification being skipped
```

```
cd test/tasks/example/eth/003-transfer-owner-template
SIMULATE_WITHOUT_LEDGER=1 just --justfile ../../../../../src/improvements/nested.just simulate foundation.   <-- No access token needed, local verification
```

```
SIMULATE_WITHOUT_LEDGER=1 just --justfile ../../../../../src/improvements/nested.just simulate foundation 0 verify-all    <-- Needs access token, local and tenderly verification
```


blmalone:
Can you provide an example command that I can run this to test?

src/improvements/justfile
```diff
@@ -143,4 +144,35 @@ simulate-verify-task TASK="":
     # For testing purposes, we do not gain anything by simulating as other nested safes.
     nested_safe_name="foundation"
 
-    ${root_dir}/src/improvements/script/simulate-verify-task.sh {{TASK}} $nested_safe_name
+    # If TASK doesn't start with / make it absolute using $PWD
+    if [[ "{{TASK}}" != /* ]]; then
```
blmalone, line 148: So is the user expected to pass through an absolute path here?

blmalone: Also, where is the `simulate-verify-task TASK="":` just command expected to be executed from?

src/improvements/justfile
```diff
@@ -143,4 +144,35 @@ simulate-verify-task TASK="":
...
+    root_dir=$(git rev-parse --show-toplevel)
+    forge build
+
+    # If the task is nested then we only simulate as the foundation.
+    # In the future we could simulate as other nested safes. 
+    # For testing purposes, we do not gain anything by simulating as other nested safes.
+    nested_safe_name="foundation"
+
+    # Later this networks list should be dynamically generated from the src/improvements/tasks directory.
+    for task in ${root_dir}/src/improvements/tasks/example/*/*; do
```
blmalone, line 168: We want to be simulating non-terminal tasks from the networks directories and not the example directory. 

```
simulate-non-terminal-tasks:
    #!/usr/bin/env bash
    set -euo pipefail
    root_dir=$(git rev-parse --show-toplevel)
    forge build

    # Later this networks list should be dynamically generated from the src/improvements/tasks directory.
    networks=("eth" "sep")
    for network in ${networks[@]}; do
        if [ "$network" != "src/improvements/tasks/example" ]; then # skip example tasks
            for task in ${root_dir}/src/improvements/tasks/${network}/*; do
               ${root_dir}/src/improvements/script/simulate-verify-task.sh $task $nested_safe_name
            done
        fi
        echo "Done simulating non-terminal tasks for network: $network"
    done
    ```

Comment (blmalone) [reply src/improvements/justfile:168]: I think this should be put this in a separate PR. The feature being added here is simulating non-terminal tasks in CI for the new superchain-ops repo.

File: script/CalculateSafeHashes.s.sol [hunk 3]
```diff
@@ -0,0 +1,39 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.15;
+
+import "forge-std/Script.sol";
+import {GnosisSafeHashes} from "src/libraries/GnosisSafeHashes.sol";
+
+contract CalculateSafeHashes is Script {
```
blmalone, line 7: Who calls this? What's the `TENDERLY_PAYLOAD`? I know its mentioned in the `simulate-verify-task.sh` file below but more information in this file as to what it's purpose is would be helpful.

blmalone (commented):
first pass

src/improvements/script/simulate-verify-task.sh
```diff
@@ -8,46 +8,73 @@ simulate_verify_task() {
     root_dir=$(git rev-parse --show-toplevel)
     local_output_file="./simulation_output.txt"
     remote_output_file="./remote_output.txt"
+    forge_output_file="./forge_output.txt"
```
blmalone, line 11: Do we need all the file i/o here with local/remote/forge_output_file? When it fails for whatever reason before cleanup we leave these files lying around for devs to delete manually.

src/improvements/justfile
alcueca: I added comments about when the command is supposed to be used, and what the parameter is

src/improvements/justfile
alcueca: The thing is that I don't have any tasks outside `examples` to test the functionality with, so if I exclude it now it does nothing.

I suggest that we don't put this into CI, and leave it as a developer tool, until we have one or more tasks outside `examples`, then we can exclude it from simulate-non-terminal-tasks and put it on CI.

script/CalculateSafeHashes.s.sol
alcueca: Good point, I added a detailed comments about that.

src/improvements/script/simulate-verify-task.sh
alcueca: We are scraping the output of other scripts to capture the payload and the hashes.

I tried to capture the output in a variable and scrape that, but we also want the output on console for debugging. It then becomes really hard to capture the output from both stdout, stderr in the variable and always show it regardless of the underlying script exit code.

The current system is more robust, as you always get all the output on the console for debugging. However, I've move the commands to remove files so that they are removed as soon as possible. If a script fails, only one file would remain, which also would have the relevant output if needed. The file would be removed in the next successful execution.

I hope that's a reasonable compromise.

alcueca:
@blmalone, thanks a lot for your comments, and apologies for not putting enough documentation upfront to avoid you getting confused.

To try the functionality, you will need API access to Tenderly. You can either:
Get an access token for `oplabs/task-simulation` [here](https://dashboard.tenderly.co/oplabs/task-simulation/settings) and store it in a `TENDERLY_ACCESS_TOKEN` environment variable
OR
Get an access token for your own personal account, store it in `TENDERLY_ACCESS_TOKEN`, and also store your tenderly user name and project in `TENDERLY_USER` and `TENDERLY_PROJECT_SLUG`, respectively. These are used in `get-tenderly-hashes.sh`

Once you have that sorted out, you can go to `src/improvements` and execute `just simulate-non-terminal-tasks`, it will skip the first task because it is `EXECUTED`, simulate and verify the hashes of the second task, and then exit with an error on the third due to a lack of a VALIDATIONS.md file. That shows the functionality.

From `src/improvements` you can also simulate and verify any individual task using `just simulate-verify-task relative/path/to/task/directory`

As a question, where would be the best place for me to put this usage info?

alcueca:
The CI will break because the default now is to verify all tasks, and the step to test all templates doesn't have the tenderly credentials needed to do so.

I'll add a parameter so that the verification can be skipped, while leaving the simplest cli command for the signer flow

lib/superchain-registry
smartcontracts, line 1: Looks like this is updated, should double check if that's intended.

src/improvements/nested.just
```diff
@@ -1,7 +1,7 @@
 set dotenv-load
 
-# Common variables used across multiple recipes
-# Export variables in uppercase for shell script usage
+# # Common variables used across multiple recipes
```
smartcontracts, line 3: Nit: looks like this was double-commented

src/improvements/nested.just
```diff
@@ -15,37 +15,61 @@ export SCRIPT_NAME := shell(cmd)
 export signatures := env_var_or_default('SIGNATURES', '')
 export randomPersonEoa := "0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045"
 
-simulate whichSafe hdPath='0':
+# Simulate a task
+# whichSafe is the safe name to simulate
+# hdPath is the derivation path for the safe
+# verify is optional and defaults to "verify-local". Other values are "verify-all", "verify-tenderly" and "skip-verify".
```
smartcontracts, line 21: Any reason not to make `verify-all` the default?

smartcontracts (approved):
Generally looks good. I vote we ship it.

alcueca:
Testing now on `$root_dir/test/tasks/example`. For correctness, the VALIDATIONS.md files should refer to the right safe that we are simulating against, and the `simulate-non-terminal-tasks` should pick all the safe name blocks in VALIDATIONS.md and simulate against all of them.

Maybe the format needs to be:
```
[validation_hashes]
  [<safe_name_1>]
    domain_separator = <0x...>
    message_hash = <0x...>
  ...
  [<safe_name_N>]
    domain_separator = <0x...>
    message_hash = <0x...>
```

alcueca:
Multiple signers can now be entered in VALIDATIONS.md

However, it is not clear to me how the domain separator and message hash are always the same for a task, whether we simulate it as the council or the foundation.

alcueca:
The broken CI check is broken on `main`.

alcueca:
Mmm, I don't think this gets the domain separator and message hash for the child multisigs, only for the parent.

alcueca:
Alright, so the tenderly simulation is a transaction directed to the parent multisig. For tenderly validation to work we need to extract the domain separator and message hash for the **child multisig** from within. I will log out the trace of a simulation and see if a string search through it reveals it.

If we have to do complex logic with custom scripts to derive the hashes from tenderly that will weaken using tenderly for verification as a security measure, because it would be difficult for others to replicate.

MultisigTask already generates the right hashes locally and is executed by `nested.just` in `src/improvements`. The call to `verify-hashes-local.sh` in `nested.just simulate` does this a second time, and wrong. That needs to be fixed.

I am still not sure how the safe addresses are provided to `src/improvements/nested.just`, it seems that it should be from an addresses.toml file that is not always present in the examples. Also, isn't that what SuperchainRegistry.sol is for?

alcueca:
Rewrote from scratch in #783

PR #739, opened 2025-03-11
Finance template

prat-gpt:
**Description**
- Add `FinanceTemplate` to execute finance related operations that include approve, transfer, increaseAllowance, decreaseAllowance.

**Tests**
- Add a test finance template to example tasks so that it is checked in `template_regression_tests` CI job.
- Add a regression test to match expected call data and data to sign

**Metadata**
- This PR is a part of generic refactor task.


src/improvements/tasks/example/sep/005-test-finance-operation/config.toml
mds1, line 1: I know we talked about this, but here is the modified config file we discussed. (I'm not sure if the `ETH = "0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE"` key/value is actually needed, if not it's ok to remove that)

```toml
templateName = "FinanceTemplate"

# Operation must be one of:
#   - approve
#   - increaseAllowance
#   - decreaseAllowance
#   - transfer
operationType = "approve"

# The operations to perform.
#   - The token strings are defined in the `[addresses]` section below.
#   - The target is also an address, mapped to a string in the `[addresses]` section.
#   - The amount is a human readable amount, e.g. 1 sends 1 ETH or 1e18 wei.
operations = [
  {token = "OP", amount = 20, target = "SecurityCouncil"},
  {token = "ETH", amount = 100, target = "MyRecipient"}
]

[addresses]
# Define the safe that should execute the transaction. This must always be
# called "SafeToSendFrom".
SafeToSendFrom = "0x0xBB4daAC11B4446EE1c6146De1e26ecf1Ab8b3EB6"

# Define any named target addresses.
MyRecipient = "0x1234567890123456789012345678901234567890"

# Tokens
ETH = "0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE"
OP = "0x4200000000000000000000000000000000000042"
```

src/improvements/template/FinanceTemplate.sol
```diff
@@ -0,0 +1,216 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import {VmSafe} from "forge-std/Vm.sol";
+
+import "forge-std/Test.sol";
+
+import {SimpleBase} from "src/improvements/tasks/MultisigTask.sol";
+import {SimpleAddressRegistry} from "src/improvements/SimpleAddressRegistry.sol";
```
mds1, line 9: unused import
```suggestion
```

src/improvements/template/FinanceTemplate.sol
```diff
@@ -0,0 +1,216 @@
...
+    /// @notice Initial allowances, before the operations are executed
+    mapping(address => mapping(address => uint256)) public initialAllowances;
+    /// @notice Initial balances, before the operations are executed,
+    mapping(address => mapping(address => uint256)) public initialBalances;
+
+    /// @notice Returns the safe address string identifier
+    /// @return The string "FoundationOperationSafe"
+    function safeAddressString() public pure override returns (string memory) {
+        return "FoundationOperationSafe";
+    }
```
mds1, line 56: per our modified toml file this becomes
```suggestion
    /// @notice The safe to send this transaction from must be specified in the config file.
    function safeAddressString() public pure override returns (string memory) {
        return "SafeToSendFrom";
    }
```

src/improvements/template/FinanceTemplate.sol
```diff
@@ -0,0 +1,216 @@
...
+    /// @notice List of operations to be executed
+    Operation[] public operations;
+    /// @notice Operation name
+    string public operationType;
+    /// @notice Operation type enum
+    OperationType public operationTypeEnum;
+    /// @notice Initial allowances, before the operations are executed
+    mapping(address => mapping(address => uint256)) public initialAllowances;
+    /// @notice Initial balances, before the operations are executed,
+    mapping(address => mapping(address => uint256)) public initialBalances;
```
mds1, line 50: nit on spacing
```suggestion
    /// @notice List of operations to be executed
    Operation[] public operations;

    /// @notice Operation name
    string public operationType;

    /// @notice Operation type enum
    OperationType public operationTypeEnum;

    /// @notice Initial allowances, before the operations are executed
    mapping(address => mapping(address => uint256)) public initialAllowances;

    /// @notice Initial balances, before the operations are executed,
    mapping(address => mapping(address => uint256)) public initialBalances;
```

src/improvements/template/FinanceTemplate.sol
```diff
@@ -0,0 +1,216 @@
...
+        storageWrites = new string[](1);
+        storageWrites[0] = "TEST";
+        return storageWrites;
+    }
+
+    /// @notice Sets up the template with module configuration from a TOML file
+    /// @param taskConfigFilePath Path to the TOML configuration file
+    function _templateSetup(string memory taskConfigFilePath) internal override {
+        string memory file = vm.readFile(taskConfigFilePath);
+        operationType = vm.parseTomlString(file, ".operationType");
```
mds1, line 72: nit, I have a small preference for `using StdToml for string`, then
```suggestion
        string memory toml = vm.readFile(taskConfigFilePath);
        operationType = toml.readString(".operationType");
```

src/improvements/template/FinanceTemplate.sol
```diff
@@ -0,0 +1,216 @@
...
+        operationTypeEnum = _getOperationType();
+
+        // Cannot decode directly to storage array, decode to memory first
+        // and then push to storage array
+        Operation[] memory operationsMemory = abi.decode(vm.parseToml(file, ".operations"), (Operation[]));
```
mds1, line 77: similar nit
```suggestion
        Operation[] memory operationsMemory = abi.decode(toml.parseRaw(".operations"), (Operation[]));
```

src/improvements/template/FinanceTemplate.sol
```diff
@@ -0,0 +1,216 @@
...
+                (address token, address target) = _getTokenAndTarget(operation.token, operation.target);
+                IERC20(token).safeDecreaseAllowance(target, operation.amount);
+            }
+        } else if (operationTypeEnum == OperationType.Transfer) {
+            for (uint256 i = 0; i < operations.length; i++) {
+                Operation memory operation = operations[i];
+                (address token, address target) = _getTokenAndTarget(operation.token, operation.target);
+                IERC20(token).safeTransfer(target, operation.amount);
+            }
+        }
```
mds1, line 119: we can dedupe some code by moving the for loop to the outside
```solidity
for (uint256 i = 0; i < operations.length; i++) {
    Operation memory operation = operations[i];
    (address token, address target) = _getTokenAndTarget(operation.token, operation.target);

    if (operationTypeEnum == OperationType.Approve) {
        IERC20(token).safeApprove(target, operation.amount);
    } else if (operationTypeEnum == OperationType.IncreaseAllowance) {
        IERC20(token).safeIncreaseAllowance(target, operation.amount);
    } else if (operationTypeEnum == OperationType.DecreaseAllowance) {
        IERC20(token).safeDecreaseAllowance(target, operation.amount);
    } else if (operationTypeEnum == OperationType.Transfer) {
        IERC20(token).safeTransfer(target, operation.amount);
    }
}
```

src/improvements/template/FinanceTemplate.sol
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    /// @notice Validates that the module was enabled correctly.
+    function _validate(VmSafe.AccountAccess[] memory, Action[] memory) internal view override {
+        if (operationTypeEnum == OperationType.Approve) {
```
mds1, line 124: Same comment on reducing code by putting the for loop on the outside

src/improvements/template/FinanceTemplate.sol
```diff
@@ -0,0 +1,216 @@
...
+        view
+    {
+        (address token, address target) = _getTokenAndTarget(tokenIdentifier, targetIdentifier);
+
+        assertEq(IERC20(token).allowance(address(parentMultisig), target), initialAllowances[token][target] - amount);
+        assertEq(IERC20(token).balanceOf(target), initialBalances[token][target]);
+    }
+
+    /// @notice Validates transfer operations
+    function _validateTransfer(string memory tokenIdentifier, string memory targetIdentifier, uint256 amount)
```
mds1, line 196: For the transfer case, it would be nice to also assert that balances for all tokens in the `SafeToSendFrom` account decreased by the sum of all operations

src/improvements/tasks/example/sep/005-test-finance-operation/config.toml
```diff
@@ -0,0 +1,15 @@
+# this is the file used to determine the network configuration
+templateName = "FinanceTemplate"
+
+# Operation is one of:
+# approve, increaseAllowance, decreaseAllowance, transfer
+operationType = "approve"
+
+operations = [
+  {token = "TEST", amount = 100, target = "SecurityCouncil"},
+  {token = "TEST", amount = 100, target = "USER1"}
```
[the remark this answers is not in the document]
mds1: I would prefer to keep target, it's a bit more intentionally vague since there is a range of operations and "destination" for an allowance change sounds a little odd

src/improvements/tasks/example/sep/005-test-finance-operation/config.toml
```diff
@@ -0,0 +1,15 @@
+# this is the file used to determine the network configuration
+templateName = "FinanceTemplate"
+
+# Operation is one of:
+# approve, increaseAllowance, decreaseAllowance, transfer
+operationType = "approve"
```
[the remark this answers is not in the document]
mds1: We don't currently need the ability to mix and match. Over time we may want to expand functionality but the initial version can be simple

src/improvements/template/FinanceTemplate.sol
prat-gpt: This logic dedupes code but it will be more computations heavy when compared with current logic that's why I didn't implement it this way. I think this logic will be more feasible when each operation can be of different type, which is not the case in our config as there can only be a single type.

src/improvements/tasks/example/sep/005-test-finance-operation/config.toml
prat-gpt: `SafeToSendFrom` cannot be an addresse in `addresses.toml` as the same address cannot be added with two different names in simple address registry. Would that be ok?

mds1: yep that's ok! I don't think we'll ever need a name of `SafeToSendFrom` in the global `addresses.toml`

src/improvements/template/FinanceTemplate.sol
prat-gpt: done here [94014fd](https://github.com/ethereum-optimism/superchain-ops/pull/739/commits/94014fd50b98a05818998ddddd151b36c8c9edbf)

src/improvements/template/FinanceTemplate.sol
prat-gpt: done here [94014fd](https://github.com/ethereum-optimism/superchain-ops/pull/739/commits/94014fd50b98a05818998ddddd151b36c8c9edbf)

src/improvements/template/FinanceTemplate.sol
prat-gpt: done here [94014fd](https://github.com/ethereum-optimism/superchain-ops/pull/739/commits/94014fd50b98a05818998ddddd151b36c8c9edbf)

src/improvements/template/FinanceTemplate.sol
prat-gpt: done here [94014fd](https://github.com/ethereum-optimism/superchain-ops/pull/739/commits/94014fd50b98a05818998ddddd151b36c8c9edbf)

src/improvements/template/FinanceTemplate.sol
prat-gpt: done here [94014fd](https://github.com/ethereum-optimism/superchain-ops/pull/739/commits/94014fd50b98a05818998ddddd151b36c8c9edbf)

src/improvements/template/FinanceTemplate.sol
```diff
@@ -40,111 +43,132 @@ contract FinanceTemplate is SimpleBase {
...
+        return new string[](0);
     }
 
     /// @notice Sets up the template with module configuration from a TOML file
     /// @param taskConfigFilePath Path to the TOML configuration file
     function _templateSetup(string memory taskConfigFilePath) internal override {
-        string memory file = vm.readFile(taskConfigFilePath);
-        operationType = vm.parseTomlString(file, ".operationType");
+        string memory toml = vm.readFile(taskConfigFilePath);
+        operationType = toml.readString(".operationType");
```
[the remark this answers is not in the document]
mds1: If they don't have bytecode, the `allowance` and `balanceOf` calls in `_templateSetup` will revert

src/improvements/template/FinanceTemplate.sol
```diff
@@ -40,111 +43,132 @@ contract FinanceTemplate is SimpleBase {
...
+        // also add the token to the set of tokens
         for (uint256 i = 0; i < operations.length; i++) {
             Operation memory operation = operations[i];
             (address token, address target) = _getTokenAndTarget(operation.token, operation.target);
+            tokens.add(token);
             initialAllowances[token][target] = IERC20(token).allowance(address(parentMultisig), target);
             initialBalances[token][target] = IERC20(token).balanceOf(target);
+            if (operationTypeEnum == OperationType.Transfer) {
+                tokensTransferred[token] += operations[i].amount;
+            }
```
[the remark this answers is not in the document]
mds1: What would the equivalent check look like for allowance? For reference, this is the comment that inspired this check: https://github.com/ethereum-optimism/superchain-ops/pull/739#discussion_r1989918070

prat-gpt:
> What would the equivalent check look like for allowance? For reference, this is the comment that inspired this check: https://github.com/ethereum-optimism/superchain-ops/pull/739#discussion_r1989918070

While transferring target's balance increases and parent multisig's balance decreases. To check the total amount of tokens that were sent by parent multisig for each token this storage is required. This is not the case with allowance, only target's allowance is updated nothing happens to the parent multisig allowance, that's why I don't think there is an equivalent check for allowance. @mds1

mds1 (approved):

mds1 (approved):

PR #743, opened 2025-03-11
AccountAccessParser: sorting functionality for state diffs

blmalone:
Ensuring that sorting happens before state diffs are processed so that decodeAndPrint will print state diffs in address order. Added a test to show this working.

### DIFF
--- a/src/libraries/AccountAccessParser.sol
+++ b/src/libraries/AccountAccessParser.sol
@@ -31,6 +31,9 @@ import {LibSort} from "@solady/utils/LibSort.sol";
 ///
 ///         // Get an array of all unique accounts that had state changes.
 ///         address[] memory accountsWithStateChanges = accountAccesses.getUniqueWrites();
+///
+///         // Get all new contracts created.
+///         address[] memory newContracts = accountAccesses.getNewContracts();
 ///     }
 /// }
 /// ```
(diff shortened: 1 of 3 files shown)


src/libraries/AccountAccessParser.sol
```diff
@@ -133,17 +136,22 @@ library AccountAccessParser {
...
     /// @notice Convenience function that wraps decode and print together.
     function decodeAndPrint(VmSafe.AccountAccess[] memory _accesses) internal view {
-        (DecodedTransfer[] memory transfers, DecodedStateDiff[] memory stateDiffs) = decode(_accesses);
+        (DecodedTransfer[] memory transfers, DecodedStateDiff[] memory stateDiffs) = decode(_accesses, true);
         print(transfers, stateDiffs);
     }
 
     /// @notice Decodes the provided AccountAccess array into decoded transfers and state diffs.
-    function decode(VmSafe.AccountAccess[] memory _accountAccesses)
+    function decode(VmSafe.AccountAccess[] memory _accountAccesses, bool _sort)
```
maurelian, line 144: Eventually I think we'll want an enum with different sort type options, but no need for now.

maurelian (approved):
This is well tested. LGTM TY

maurelian (approved):

test/libraries/AccountAccessParser.t.sol
```diff
@@ -54,27 +62,37 @@ contract AccountAccessParser_decodeAndPrint_Test is Test {
     address constant addr9 = address(9);
     address constant addr10 = address(10);
 
-    // Proxy1 (1 Storage Write) -> Impl1 -> Proxy2 (1 Storage Write) -> Impl2
+    // Proxy1 (1 Storage Write) -> Impl1 -> ProxyA (1 Storage Write) -> Impl2
```
blmalone, line 65: I made this test more complex as I was checking an assumption was correct.

PR #756, opened 2025-03-14
fix: Monorepo int tests failing because of TaskRunner.

blmalone:
Integration tests are failing because now we have to pass through the child multisig when simulating a nested task.

### DIFF
--- a/src/improvements/tasks/TaskRunner.sol
+++ b/src/improvements/tasks/TaskRunner.sol
@@ -56,7 +56,13 @@ contract TaskRunner is Script {
                 string.concat("out/", config.templateName, ".sol/", config.templateName, ".json");
 
             MultisigTask task = MultisigTask(deployCode(templatePath));
-            task.simulateRun(config.path);
+            
+            if (isNestedTask(config.path)) {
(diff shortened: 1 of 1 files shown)


src/improvements/tasks/TaskRunner.sol
```diff
@@ -56,7 +56,13 @@ contract TaskRunner is Script {
                 string.concat("out/", config.templateName, ".sol/", config.templateName, ".json");
 
             MultisigTask task = MultisigTask(deployCode(templatePath));
-            task.simulateRun(config.path);
+            
+            if (isNestedTask(config.path)) {
+                // TODO: Replace with the actual child multisig address - for now we're using the mainnet Security Council.
+                task.signFromChildMultisig(config.path, 0xc2819DC788505Aac350142A7A707BF9D03E3Bd03);
```
mds1, line 62: When should we do this TODO, i.e. any reason not to do it in this PR?

blmalone: Added.

mds1 (approved):

PR #764, opened 2025-03-17
Finance Template Decimal Scaling

ElliotFriedman:
Scale token amounts in finance template based on the underlying token decimals.

This was written as a library so the same code for parsing token amounts can be re-used for the Eth transfer template once that PR lands.


src/improvements/template/FinanceTemplate.sol
```diff
@@ -74,6 +88,18 @@ contract FinanceTemplate is SimpleBase {
         return new string[](0);
     }
 
+    /// @notice converts string to a scaled up token amount in decimal form
+    /// @param amount string representation of the amount
+    /// @param token address of the token to send, used discovering the decimals
```
blmalone, line 93: typo

src/libraries/DecimalNormalization.sol
```diff
@@ -0,0 +1,30 @@
...
+    /// @notice converts string to a scaled up token amount in decimal form
+    /// @param amount string representation of the amount
+    /// @param tokenDecimals number of decimals for the token
+    /// returns the scaled up token amount
+    function normalizeTokenAmount(string memory amount, uint8 tokenDecimals) internal pure returns (uint256) {
+        // Get the scaled amount and decimals using parseDecimals
+        (uint256 scaledAmount, uint8 parsedDecimals) = StringParser.parseDecimals(amount);
+
+        // Ensure amount decimals don't exceed token decimals
+        require(parsedDecimals <= tokenDecimals, "amount decimals must be less than or equal to token decimals");
```
blmalone, line 25: For error messages, can you include the file name like we've done in the other files? e.g. https://github.com/ethereum-optimism/superchain-ops/blob/main/src/improvements/SuperchainAddressRegistry.sol#L150-L151

```suggestion
        require(parsedDecimals <= tokenDecimals, "DecimalNormalization: Amount decimals must be less than or equal to token decimals");
```

src/libraries/StringParser.sol
```diff
@@ -0,0 +1,79 @@
...
+        uint8 decimalPlaces;
+        bool hasDecimals;
+    }
+
+    /// @notice Validates that the amount string has a valid format
+    /// @param components The split components of the amount string
+    /// reverts if components are invalid
+    function validateAmountFormat(string[] memory components) internal pure {
+        // Check for invalid format (more than one decimal point)
+        require(components.length <= 2, "invalid amount");
```
blmalone, line 25: Can you do the same with all the require functions in this file (comment above).

src/libraries/StringParser.sol
```diff
@@ -0,0 +1,79 @@
...
+    /// @return True if either part results in a non-zero value
+    function isNonZeroAmount(uint256 wholePart, uint256 decimalPart) internal pure returns (bool) {
+        return wholePart > 0 || decimalPart > 0;
+    }
+
+    /// @notice Parses a string amount into its components
+    /// @param amount The amount string to parse
+    /// @return components The parsed amount components
+    function parseAmountComponents(string memory amount) internal pure returns (AmountComponents memory components) {
+        VmSafe vm = VmSafe(address(uint160(uint256(keccak256("hevm cheat code")))));
```
blmalone, line 45: Can you move this declaration to the top level of this file? Like this: https://github.com/ethereum-optimism/superchain-ops/blob/main/src/libraries/AccountAccessParser.sol#L46-L47 

It's what we're doing in other files and it's easy for me to know that this file requires vm access.

src/libraries/StringParser.sol
```diff
@@ -0,0 +1,79 @@
...
+import {LibString} from "@solady/utils/LibString.sol";
+
+/// @title StringParser
+/// @notice Library for parsing string representations of decimal numbers
+library StringParser {
+    using LibString for string;
+
+    /// @notice Struct to hold parsed amount components
+    struct AmountComponents {
+        uint256 wholePart;
```
blmalone, line 14: I personally prefer the following: 
- Integer part (instead of wholePart)
- Fractional part (instead of decimalPart)

I think it's more common to head the above than what is in this PR. But I could be wrong. I had to look up `wholePart` meaning.

src/libraries/DecimalNormalization.sol
```diff
@@ -0,0 +1,30 @@
...
+
+        // Scale the amount to match token decimals
+        return scaleDecimals(scaledAmount, tokenDecimals - parsedDecimals);
```
blmalone, line 28: What would `tokenDecimals - parsedDecimals` as a variable be called? The arg to `function scaleDecimals(uint256 amount, uint8 decimals)` is called `decimals` but there seems to be something more happening here. Some normalization has already occurred at this point (in the string parser).

blmalone (commented):
Looks good. I found it hard to follow the two libraries. It seems like StringParser does some normalization (`parseDecimals`) then DecimalNormalization finishes it. I think it would be clearer if DecimalNormalization did it all.

src/libraries/StringParser.sol
```diff
@@ -0,0 +1,79 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import {VmSafe} from "forge-std/Vm.sol";
+import {LibString} from "@solady/utils/LibString.sol";
+
+/// @title StringParser
+/// @notice Library for parsing string representations of decimal numbers
+library StringParser {
```
blmalone, line 9: I'm not sure about this library name, it's v generic when this logic is pretty specific to parsing decimal amounts to then later be normalized.

test/libraries/DecimalNormalization.t.sol
```diff
@@ -0,0 +1,147 @@
...
+pragma solidity 0.8.15;
+
+import {MockERC20} from "lib/solady/test/utils/mocks/MockERC20.sol";
+
+import {Test} from "forge-std/Test.sol";
+
+import {StringParser} from "src/libraries/StringParser.sol";
+import {DecimalNormalization} from "src/libraries/DecimalNormalization.sol";
+
+library MockDecimalNormalization {
```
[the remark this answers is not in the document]
mds1: Agreed with a comment, but I would recommend making this a contract called `DecimalNormalizationHarness`, instead of a library, to avoid library linking which sometimes can cause issues

src/libraries/StringParser.sol
mds1: Agreed with your above comment that we should just merge the two into one library

blmalone (approved):
lgtm

test/libraries/DecimalNormalization.t.sol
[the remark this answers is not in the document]
blmalone: Interesting is this a new feature.

src/libraries/DecimalNormalization.sol
blmalone: Will close this now that it's been moved all into the same file.

src/libraries/DecimalNormalization.sol
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    /// @notice returns the amount with the length of the decimals parsed
+    /// @param amount The amount to parse
+    /// @return The amount and the number of decimals
+    function parseDecimals(string memory amount) internal pure returns (uint256, uint8) {
+        AmountComponents memory components = parseAmountComponents(amount);
+
+        if (!components.hasDecimals || components.decimal == 0) {
+            return (components.integer, 0);
```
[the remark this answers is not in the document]
alcueca: I would probably revert if there is no decimal part, to force users to write amounts like 1.00 ETH.

src/libraries/DecimalNormalization.sol
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        uint256 finalAmount = components.integer * (10 ** uint256(components.decimalPlaces)) + components.decimal;
```
alcueca, line 105: You shouldn't be scaling amounts in a `parseDecimals` function, that's confusing.

alcueca: I would probably have two functions:
- `parseAmountComponents`: Gets a string amount and a decimal count and returns an AmountComponents struct.
- `normalizeTokenAmount`: Gets a string amount and a token, does some verification, obtains the decimals, calls `parseAmountComponents`, scales up the integer part, and returns the scaled up integer part plus the decimal part.

alcueca (changes requested):

src/libraries/DecimalNormalization.sol
```diff
@@ -0,0 +1,109 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import {Vm, VmSafe} from "forge-std/Vm.sol";
```
blmalone, line 4: VmSafe is no longer used.

src/libraries/DecimalNormalization.sol
```diff
@@ -0,0 +1,109 @@
...
+        if (components.length == 2) {
+            require(bytes(components[1]).length != 0, "DecimalNormalization: decimals cannot be 0");
+        }
+    }
+
+    /// @notice Checks if the amount represents a non-zero value
+    /// @param wholePart The whole number part of the amount
+    /// @param decimalPart The decimal part of the amount
+    /// @return True if either part results in a non-zero value
+    function isNonZeroAmount(uint256 wholePart, uint256 decimalPart) internal pure returns (bool) {
```
blmalone, line 67: Can you move away from the wholePart terms in this function too (related to an earlier discussion we had). The natspec will also need updated.

src/libraries/DecimalNormalization.sol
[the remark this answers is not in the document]
blmalone: Can we also make some reasonable assumptions here about the users behavior. If they're deviating from that behavior then error where appropriate. e.g. if they're sending an amount with decimals they should use decimal points.

src/improvements/template/FinanceTemplate.sol
```diff
@@ -76,6 +87,18 @@ contract FinanceTemplate is SimpleBase {
         return new string[](0);
     }
 
+    /// @notice converts string to a scaled up token amount in decimal form
+    /// @param amount string representation of the amount
+    /// @param token address of the token to send, used for discovering the amount of decimals
+    /// returns the scaled up token amount
+    function getTokenAmount(string memory amount, address token) public view returns (uint256) {
+        // Get token decimals
+        uint8 tokenDecimals = ERC20(token).decimals();
```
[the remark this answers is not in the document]
alcueca: OpenZeppelin [does assume](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/d9933585b6c134589d6819909a932de4c1a0db7f/contracts/token/ERC20/extensions/ERC20Wrapper.sol#L40) that ERC20s could have no `decimals` function, because it is not well defined in the standard.

However, we don't need to be as careful as OZ, and you are right that there are no good examples of ERC20s without a `decimals` function.

src/libraries/DecimalNormalization.sol
alcueca: Wrote a PR instead of going back and forth:
https://github.com/solidity-labs-io/superchain-ops/pull/36

alcueca:
I have no problem approving this once CI passes

blmalone (approved):

alcueca (approved):

PR #766, opened 2025-03-18
Restore the original U12 runbook

maurelian:
This runbook was modified to simplify the creation and review of ugrade 12a. This PR restores it for posterity. It is already executed and does not require careful review.


ajsutton (approved):

ajsutton:
For reference https://github.com/ethereum-optimism/superchain-ops/commit/773c02249537847e0bce8cc045f24ff1aa86da2c said it copied these files but actually wound up moving them.

PR #773, opened 2025-03-19
Fix links

Hopium21:
The primary goal of these changes is to update links to the correct releases.json


blmalone:
Thank you, but we don't accept typo PR updates in this repo.

PR #775, opened 2025-03-19
Choose task type in just new recipe

prat-gpt:
**Description**
- Pass an extra argument to just new recipe to specify the task type
- Create template that inherits from task type contracts, `L2TaskBase` or `SimpleBase`
- Create individual files for `L2TaskBase` and `SimpleBase`
- Remove `_configureTask` and `taskType` functions as they are already implemented in the task type contracts.

**Tests**
- Add tests for just new recipe
- Includes a test that makes sure that if a new task type is added, the test will fail if just new recipe has not been updated.
- Add just new tests to ci

**Metadata**
This PR addresses the discussion [here](https://discord.com/channels/1244729134312198194/1342572064175030293/1351378726645927988)


src/improvements/justfile
```diff
@@ -16,20 +16,51 @@ new COMMAND="":
         echo "  • task      - Create a new task"
         echo
     }
+
+    show_task_types() {
+        echo
+        echo "Available task types:"
+        echo "  • l2taskbase - L2Base task type, these tasks require a superchain address registry and can iterate over multiple l2chains"
```
blmalone, line 23:
```suggestion
        echo "  • l2taskbase - 'L2TaskBase' task type, these tasks require a superchain address registry and can iterate over multiple L2 chains"
```

src/improvements/justfile
```diff
@@ -16,20 +16,51 @@ new COMMAND="":
...
+        echo "    eg: https://github.com/ethereum-optimism/superchain-ops/blob/main/test/tasks/mock/template/GasConfigTemplate.sol"
+        echo "  • simplebase - SimpleBase task type, these tasks require simple key value registry and are independent of the l2chains"
```
blmalone, line 25:
```suggestion
        echo "  • simplebase - 'SimpleBase' task type, these tasks require simple key value registry and are independent of the L2 chains"
```

src/improvements/justfile
```diff
@@ -16,20 +16,51 @@ new COMMAND="":
...
+        echo "    eg: https://github.com/ethereum-optimism/superchain-ops/blob/main/src/improvements/template/FinanceTemplate.sol"
+        echo
```
blmalone, line 27: Can you also add in OPCMBaseTask? Which inherits from type L2BaseTask. Most of our upgrades going forward will be through OPCM so this would be very helpful to have available.

src/improvements/justfile
```diff
@@ -16,20 +16,51 @@ new COMMAND="":
...
+                show_task_types
+                exit 1
+            fi
+
+            case "{{TASK_TYPE}}" in
+                l2taskbase)
+                    TASK_TYPE="L2TaskBase"
+                    ;;
+                simplebase)
+                    TASK_TYPE="SimpleBase"
```
blmalone, line 49: As above, OPCMBaseTask would be good here.

src/improvements/script/create-template.sh
```diff
@@ -23,8 +27,20 @@ create_template() {
...
             mkdir -p "$(dirname "$template_path")"
             # Replace all occurrences of "EmptyTemplate" in the template file with the chosen contract name.
-            sed "s/EmptyTemplate/${contract_name}/g" "$template_source" > "$template_path"
+            # Replace MultisigTask with the chosen task type.
+            # Replace TASk_TYPE.sol with MultisigTask.sol
+            # Remove the _configureTask and taskType functions. As these are
+            # already implemented in the TASK_TYPE contract.
+            # TODO: Import the TASK_TYPE contract from TASK_TYPE.sol once
+            # individual task type files are created.
+            sed -e "s/EmptyTemplate/${contract_name}/g" \
```
blmalone, line 36: When I run `just new template l2taskbase` and then look at the file that gets created, I have 2 unused import and one incorrect import. 

See here:
CleanShot 2025-03-21 at 20 32 35@2x

It should be: 
CleanShot 2025-03-21 at 20 33 16@2x

blmalone: The same goes for SimpleBase templates: 

CleanShot 2025-03-21 at 20 34 48@2x

src/improvements/script/test-just-new.sh
```diff
@@ -0,0 +1,45 @@
...
+
+# Test cases
+root_dir=$(git rev-parse --show-toplevel)
+test_command "just new" "Error: No command specified"
+test_command "just new template" "Error: No task type specified"
+test_command "just new template invalid" "Error: Invalid task type 'invalid'"
+test_command "just new task l2" "Error: Task type should not be specified for 'task' command"
+# Test that all task types are supported by just new template
+# Iterate over all Solidity files in the types folder
+for file in $root_dir/src/improvements/tasks/types/*.sol; do
```
blmalone, line 33: Again can the test be updated to support OPCMBaseTask too?

blmalone (commented):
It's great to get tests in for this. Just some comments after going through the usage.

src/improvements/justfile
prat-gpt: Done.

src/improvements/script/create-template.sh
prat-gpt: Done.

src/improvements/script/test-just-new.sh
prat-gpt: I have moved `OPCMBaseTask` to types directory and this for loop makes sure that `OPCMBaseTask` is supported.

blmalone (approved):
Great!

blmalone (dismissed):

PR #783, opened 2025-03-22
Integrate with Tenderly and perform hash matching

alcueca:
**Description**

_Please export the tenderly access token in the 1password EVM Safety vault as TENDERLY_ACCESS_TOKEN for the below to work._

- get-tenderly-hashes.sh now takes a simulation link as input
- MultisigTask.sol no longer generates a tenderly json payload
- `just simulate-tenderly <path-to-task> <optional-safe-name>` in `src/improvements` simulates a task both locally and on tenderly, both single and nested. Returns the usual output, the tenderly simulation link (already simulated, to save some time) and the hashes.
- `just verify-non-terminal-tasks` simulates locally and on tenderly all tasks that are not SIGNED or EXECUTED, and compares the domain and message hashes in VALIDATIONS.md against the ones in the simulations.


alcueca:
Note that `just verify-non-terminal-tasks` fails with `sep/001-opcm-upgrade-v200` because MultisigTask doesn't print out its individual domain and message hash. I think that's a bug in MultisigTask.

blmalone (commented):
Left a comment on the PR. I ran into some errors.

src/improvements/justfile
```diff
@@ -171,16 +171,97 @@ simulate-all-templates:
...
+# cd src/improvements
+# just simulate-tenderly tasks/sep/000-opcm-upgrade-v200 council
+# just simulate-tenderly tasks/sep/003-unichain-superchain-config-fix
+simulate-tenderly TASK="" SAFE_NAME="":
+    #!/usr/bin/env bash
+    set -euo pipefail
+    root_dir=$(git rev-parse --show-toplevel)
+    forge build
+
+    # If the task path doesn't start with a slash, append the current directory to the task path
```
blmalone, line 188: This line makes my editor appear misconfigured. Removing the apostrophe helps. 

CleanShot 2025-03-27 at 16 33 34@2x
CleanShot 2025-03-27 at 16 33 44@2x

src/improvements/justfile
```diff
@@ -171,16 +171,97 @@ simulate-all-templates:
...
 
-    # If the task is nested then we only simulate as the foundation.
-    # In the future we could simulate as other nested safes. 
-    # For testing purposes, we do not gain anything by simulating as other nested safes.
-    nested_safe_name="foundation"
+    for task in ${root_dir}/src/improvements/tasks/*/*; do
+        if [ -d "$task" ]; then
 
-    ${root_dir}/src/improvements/script/simulate-verify-task.sh {{TASK}} $nested_safe_name
+            # Read the README.md file and skip if it contains "Status: [SIGNED] or "Status: [EXECUTED]"
```
blmalone, line 212: CleanShot 2025-03-27 at 16 35 17@2x
Same thing with this comment.

src/improvements/justfile
```diff
@@ -171,16 +171,97 @@ simulate-all-templates:
         exit 1
     fi
 
-simulate-verify-task TASK="":
+# Simulate a task locally and against Tenderly.
```
blmalone, line 174: Why do we need a new base contracts version?

src/improvements/justfile
```diff
@@ -171,16 +171,97 @@ simulate-all-templates:
...
+                if grep -q "Status: \[SIGNED\]" "$readme_file" || grep -q "Status: \[EXECUTED\]" "$readme_file"; then
+                    echo "Skipping signed/executed task: $task"
+                    continue
+                fi
+            fi
+
+            # Read the validations file and extract the nested safe names
+            validation_file="${task}/VALIDATION.md"
+            
+            # The safe names and how they appear in validations are:
```
blmalone, line 224: Can we not get these programmatically some how? This approach seems precarious. The config.toml should have these addresses.

blmalone (commented):
I've left some comments on this. For some reason simulations aren't working for our U13 tasks anymore with these changes. I think it's because of the base contracts submodule change. Why is this included? It's not adding the `TENDERLY_GAS` env var to the Tenderly simulation link anymore. 

I also ran `just simulate-tenderly tasks/sep/001-opcm-upgrade-v200 child-safe-1` and I get the following error: 
```
View the simulation in Tenderly dashboard:
https://dashboard.tenderly.co/oplabs/task-simulation/simulator/52633456-f754-41d8-8a16-f992f3c278a2
Error: Transaction failed in Tenderly
error: Recipe `simulate-tenderly` failed with exit code 1
```

When is this meant to get executed? I was expecting to see some CI job that runs this on every PR (or something like that - could be every merge to main).

Also, given how much churn has been included in this PR, can you add some tests? This could be a CI job that runs these checks.

src/improvements/justfile
alcueca: Interesting, I use the standard "Shell Script" highlighting from VSCode for justfiles and it works fine. Fixed anyway.

src/improvements/justfile
alcueca: We don't, really. Reverted the update.

src/improvements/justfile
alcueca: Added the missing quotes, but I think that you are just using the wrong highlighting.

src/improvements/justfile
alcueca: Added a new section in `config.toml` with the equivalences between the short names used in `just simulate` and the long names in `VALIDATIONS.md`. The result is much more robust.

alcueca:
I left some comments on slack, but the short of it is that the tasks in `main` are all broken.

The commands in this PR fail because that is what they need to do when they find broken tasks.

I don't think that the broken tasks should be fixed in this PR, but on `main`. You could use the scripts in this PR to do so.

alcueca:
I'll try again to get this merged when I'm back from leave. The conflicts shouldn't build up much, just on `config.toml` files that can be easily fixed.

PR #791, opened 2025-03-25
refactor: addresses.toml no longer holding safe addresses.

blmalone:
When trying to simulate for base, we run into an issue where it tries to use the nested safe addresses for our op-mainnet 2-of-2. This is because `addresses.toml` file only contained the 2-of-2 configuration. I've moved that configuration out of the `addresses.toml` file and opted to add those nested safe addresses to the respective `config.toml` file.


test/registry/SimpleAddressRegistry.t.sol
```diff
@@ -109,21 +109,4 @@ contract SimpleAddressRegistryTest is Test {
         assertEq(registry.get("FoundationUpgradeSafe"), 0xDEe57160aAfCF04c34C887B5962D0a69676d3C8B, "20");
         assertEq(registry.get("SecurityCouncil"), 0xf64bc17485f0B4Ea5F06A96514182FC4cB561977, "30");
     }
-
-    function test_hardcodedAddresses_opMainnet() public {
```
blmalone, line 113: Removing this test as I don't really see how it makes sense. I think addresses.toml shouldn't contain chain specific addresses e.g. nested safes. This PR opts to add nested addresses in the config.toml file of each test.

JosepBove: Makes sense

JosepBove (approved):

JosepBove (approved):

PR #797, opened 2025-03-26
refactor: State Override Parity

blmalone:
Adding some documentation to explain the state overrides that you'll see in Tenderly for the new superchain-ops flow.

When creating this documentation I noticed a bug where I was using the nuclear option for single simulation state overrides. 
I was overriding the owner mapping if it was a nested simulation when I didn't need to do this. I only need to override the parent multisig owner mapping when it is a single safe simulation because I need to make sure we can call execute on the parent multisig from the EOA initiating the transaction.

I've confirm tenderly links work for both single and tested tasks with these new changes.


src/improvements/tasks/StateOverrideManager.sol
```diff
@@ -97,6 +97,16 @@ abstract contract StateOverrideManager is CommonBase {
         defaultOverride = Simulation.addOwnerOverride(parentMultisig, defaultOverride, owner);
     }
 
+    /// @notice Parent multisig override for nested execution.
```
blmalone, line 100: For nested simulations, we should not perform the owner override.

blmalone: For single simulations, we do perform the owner override on the parent multisig so we can execute the transaction from the sender.

test/tasks/StateOverrideManager.t.sol
```diff
@@ -291,14 +291,30 @@ contract StateOverrideManagerUnitTest is Test {
...
         );
-        // 5 overrides: threshold, nonce, owner count, owner mapping, owner mapping 2
-        assertTrue(
-            parentDefaultOverride.overrides.length == 5,
-            string.concat(
-                "Default override must have 5 overrides, found: ",
-                LibString.toString(parentDefaultOverride.overrides.length)
-            )
-        );
+        // 5 possible overrides: <threshold>, <nonce>, [owner count], [owner mapping], [owner mapping 2]
```
blmalone, line 294: Updating the tests with the new assumptions.

NESTED-VALIDATION.md
```diff
@@ -114,3 +114,75 @@ The only other state changes related to the nested execution are _three_ nonce i
 `0x0000000000000000000000000000000000000000000000000000000000000005` on a
 `GnosisSafeProxy`.
 - One increment of the nonce of the EOA that is the first entry in the owner set of the Safe Signer.
+
+## Visualisation of State Overrides
```
maurelian, line 118:
```suggestion
## Visualisation of State Overrides for Nested Execution
```

NESTED-VALIDATION.md
```diff
@@ -114,3 +114,75 @@ The only other state changes related to the nested execution are _three_ nonce i
...
+
+The following diagram shows the state overrides for a nested execution. Where appropriate, some state overrides are marked as optional to indicate that they are not present for a single execution.
+
+```mermaid
+graph LR
```
maurelian, line 123: I think that it'd be [more readable](https://mermaid.live/edit#pako:eNqVVV1P2zAU_SuWeeClVLHjJDSTJtGm0l4Yk2B7WIsm4zhpRGpHdjJglP--6yRACCuMl7b3-pxzP-3eY6FTiWOcG15t0FohZJurzjhtyroQvCx950bohKx4nhuZ81r6P3jZSOTdkoilkkfkssP8g4wW8Gm741aGrpwHkRjJWykuDFeWi7rQ6nIA8jsQ3QeSKnU_-q-nsItNUaZdcFvkHXROViMNyDvkUUioR8d5L07P0XkNFaKz39KYIpWD1Od0dbEx0m40BCkssrJGtUZkkPjcX33VSkh3LIt6Iw0SjTFS1UgbpHvNIYGtzm4UwBa6AdAe1aAHnfKqKlSOCDoathhYSt4g7TBDWjii0be796z4i3bAxbszH2t8422xL0eQgExVGSj_C7cbkOAhZxlhos8noR8Y0be3RpT4744oYR8cUQLdr1xSvIzRnmHBPM5UeYcqiO2UMpBSWh0paWuZIltsm5I7DTsdKoevlIcT_l_FngrpHG5tPgV4Ks3hi0DR3kD046m_uz6JLNuVcXe4oyzJao2fVwnGS7n0QxGINd7_dgx1uncEPda0pKvvFeBTiWpur5GtpCiyQqCsUfsfihNo69FneINagzpjR3bwRDh73h7u6A7W3tmLzvZ3sL8t3m9tBnbLTzp-sIPqnL3sxJftoSi5tYnM0JW7NFlRlvGBEDLIsomtjb6W8YHnRVyIidClNs7y_TD89IIL7ZKqJ6dMpil_ItNjHrHgkUyCIKLsmewKdYEHDtqpDTz-2DMfkxZjR_LKQUeO5Rix7APjCd5Ks-VFCn839-58jeHybeUax_Az5eZ6jdfqAXC8qfX5nRI4rk0jJ9joJt_gOOOlBaupUliJpOAw_e2Tt-Lqp9bbR0puXJye3t6H9sriGHrmsDi-x7c4psfRNCDEI37gBZTRcILvcDybTX3PD-mM-SxilIUPE_ynFSdTL2QUun0czWZeRGbs4S8s4k5d) when oriented top to bottom 

```suggestion
graph TD
```

src/improvements/tasks/StateOverrideManager.sol
maurelian: nit: it seems like `_ownerMultisig` might be slightly more accurate, as it's not really a parent in the single case, but I think it's probably not worth changing right now.

maurelian (commented):
Left some suggestions.

src/improvements/tasks/StateOverrideManager.sol
blmalone: Yea I think we started using parent and child in the new scops and now I keep using it.

maurelian (approved):

PR #799, opened 2025-03-26
Add correct safe lookup for simulation to the new single.just file

trianglesphere:
**Description**

This does a safe address lookup (must be specified in the config.toml) and then fetches a signer address if the SIMULATE_WITHOUT_LEDGER is set.

PR #800, opened 2025-03-26
refactor: single simulation should give the option to change the from…

blmalone:
Alternative to this [PR](https://github.com/ethereum-optimism/superchain-ops/pull/799/files). We should allow users the ability to change the from address in the Tenderly simulation link.

### DIFF
--- a/src/improvements/single.just
+++ b/src/improvements/single.just
@@ -15,16 +15,33 @@ export SCRIPT_NAME := shell(cmd)
 export signatures := env_var_or_default('SIGNATURES', '')
 export forkBlockNumber := env_var_or_default('FORK_BLOCK_NUMBER', '-1')
 
-simulate hdPath='0':
+# 'senderAddress' is the address you want to run this simulation as.
+# The 'senderAddress' will be the 'from' field in the Tenderly simulation. 
+simulate senderAddress="" hdPath='0':
   #!/usr/bin/env bash
 
   config=${TASK_PATH}/config.toml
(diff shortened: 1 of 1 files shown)


src/improvements/single.just
```diff
@@ -15,16 +15,33 @@ export SCRIPT_NAME := shell(cmd)
...
   script=${SCRIPT_PATH}/template/${SCRIPT_NAME}.sol
 
   echo "RPC URL: ${rpcUrl}"
-
   echo "Using script ${script}"
-  echo ""
+
+  echo "Setting sender address..."
+  # Default fallback sender: address(uint160(uint256(keccak256("foundry default caller"))))
+  default_sender=0x1804c8AB1F12E6bbf3894d4083f33e07309d1f38
```
maurelian, line 31: shouldn't the default_sender get an owner [from the safe owners like here](https://github.com/ethereum-optimism/superchain-ops/blob/15397a1c4034e008e5fc89c9e5136bcffc250944/src/improvements/nested.just#L31)?

blmalone: If we back up for a second, what is the reason we want to see the 'from' address in Tenderly be an owner of a safe or a ledger? 
I agree, it looks nice but I don't think it's adding anything? Ultimately the state overrides mean the sender can be from the default foundry account. 

If I add the line in you're talking about we're effectively removing a feature that we wanted in the new superchain-ops i.e. no having to enter the safe addresses in .env or config files.

PR #803, opened 2025-03-26
isFeaturedEnabled helper

AmadiMichael:
create utils library and isFeatureEnabled helper function, add test for it

### DIFF
--- a/src/libraries/Utils.sol
+++ b/src/libraries/Utils.sol
@@ -1,0 +1,12 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import {VmSafe} from "lib/forge-std/src/Vm.sol";
+
+library Utils {
+    VmSafe private constant vm = VmSafe(address(uint160(uint256(keccak256("hevm cheat code")))));
+
+    function isFeatureEnabled(string memory _feature) internal view returns (bool) {
+        return vm.envOr(_feature, false) || vm.envOr(_feature, uint256(0)) == 1;
+    }
+}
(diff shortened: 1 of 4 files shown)


src/libraries/Utils.sol
```diff
@@ -0,0 +1,12 @@
...
+pragma solidity 0.8.15;
+
+import {VmSafe} from "lib/forge-std/src/Vm.sol";
+
+library Utils {
+    VmSafe private constant vm = VmSafe(address(uint160(uint256(keccak256("hevm cheat code")))));
+
+    function isFeatureEnabled(string memory _feature) internal view returns (bool) {
+        return vm.envOr(_feature, false) || vm.envOr(_feature, uint256(0)) == 1;
+    }
```
mds1, line 11: Let's add a natspec comment explaining that this method is to help simplify feature-flagging by reading an env var and defaulting to false

test/libraries/Utils.t.sol
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.15;
+
+import {Test} from "forge-std/Test.sol";
+import {Utils} from "src/libraries/Utils.sol";
+
+contract UtilsTest is Test {
+    function test_isFeatureEnabled() public {
```
mds1, line 8: A few other test cases to add:
- empty string (false)
- the true/false strings with whitespace around them (false)
- other casings (true/True/TRUE should all be true and false/False/FALSE should all be false)
- hex input, e.g. 0x1 and 0x0
- a number formatted as 01

AmadiMichael: fixed

src/libraries/Utils.sol
AmadiMichael: done

mds1 (approved):

mds1:
nice tests 👌

PR #805, opened 2025-03-27
Fix Sepolia 001-opcm-upgrade-v200

xenoliss:
**Description**

This PR fixes 2 things related to the Base Sepolia 001-opcm-upgrade-v200 task:
1. The `nested.just` file was calling the `get-safe.sh` bash script with the incorrect number of parameters.
2. The expected message hash in the `VALIDATION.md` was not correct and has been replaced with `0xbdd593a729dadfa12c86ebac23914145d1ae0418ef1909ffecf78882d6720212`


blmalone (approved):
Good catch on nested.just file! Our CI should've caught that so I'll have to look into it. The message hashes match what I see. LGTM

PR #818, opened 2025-03-31
feat: template for setRespectedGameType

smartcontracts:
Creates a new template for running setRespectedGameType on one or more chains at the same time.


src/improvements/template/EnableDeputyPauseModuleTemplate.sol
```diff
@@ -20,7 +20,7 @@ contract EnableDeputyPauseModuleTemplate is SimpleBase {
     address public newModule;
 
     /// @notice Constant safe address string identifier
-    string constant _SAFE_ADDRESS = "FoundationOperationSafe";
+    string constant _SAFE_ADDRESS = "FoundationOperationsSafe";
```
Ethnical, line 23: Since we have only FoS I am wondering yet why Operation**s**. 
This is my first review I will delete this comment If I found why.

src/improvements/SimpleAddressRegistry.sol
```diff
@@ -74,7 +74,7 @@ contract SimpleAddressRegistry is StdChains {
         require(bytes(key).length > 0, "SimpleAddressRegistry: empty key");
         require(who != address(0), string.concat("SimpleAddressRegistry: zero address for ", key));
         require(addresses[key] == address(0), string.concat("SimpleAddressRegistry: duplicate key ", key));
-        require(bytes(keys[who]).length == 0, string.concat("SimpleAddressRegistry: address already registered"));
+        require(bytes(keys[who]).length == 0, string.concat("SimpleAddressRegistry: address already registered: ", key));
```
Ethnical, line 77: I like it but if we want to be consistent with just above we should remove the second `:` or add them everywhere.
```suggestion
        require(bytes(keys[who]).length == 0, string.concat("SimpleAddressRegistry: address already registered ", key));
```

src/improvements/template/EnableDeputyPauseModuleTemplate.sol
smartcontracts: In the specs we call it "Foundation Operations Safe" - unifying the naming here to account for that.

test/tasks/Regression.t.sol
```diff
@@ -103,6 +104,29 @@ contract RegressionTest is Test {
         _assertDataToSignSingleMultisig(multisigTask, actions, expectedDataToSign);
     }
 
+    /// @notice expected call data and data to sign generated by manually running the EnableDeputyPauseModuleTemplate at block 7745524 on sepolia
```
blmalone, line 107: Small nit on this comment, needs to be for SetRespectedGameType with the correct block on mainnet.

blmalone (approved):

PR #819, opened 2025-03-31
feat: Stacked Simulator

blmalone:
### ** This PR is being kept in draft state until the old superchain-ops system is deprecated **

This PR introduces stacked simulations for the new superchain-ops improvements. The TLDR of stacked simulations is that we need them to allow for task developers to work in parallel. Some future tasks depends on the resultant state of a prior task. Running the simulation of your task as a stack will give your task access to the latest state changes.

We previously had a similar feature implemented [here](https://github.com/ethereum-optimism/superchain-ops/pull/611). Instead of reusing that logic I've reimplemented a native version in solidity that makes use of preexisting code with tests. 

**Features:**
- enforced naming conventions to tasks
- enforced file layout for tasks
- natively integrated with new superchain-ops
- ability to simulate all non terminal tasks for a network 
- ability to simulate subset of non terminal tasks for a network
- ci job that runs all non-terminal tasks and doesn't need to be updated when new tasks are added

**Usage:**
```
# just simulate-stack <network> [task-name] [owner-address]
just simulate-stack eth
just simulate-stack eth 002-opcm-upgrade-v200
just simulate-stack eth 002-opcm-upgrade-v200 0xabcdef...
```

**Tips for reviewing**
- while this PR is large, most of the changes are additional tests
- try a stacked simulation for yourself using the command above

PR #822, opened 2025-04-01
feat: StateOverrideManager appending overrides.

blmalone:
### Problem

Originally we followed the approach of applying user overrides the same way base-contracts did [here](https://github.com/base/contracts/blob/6ca464d118d2d33331ea44893f0930f753663f60/script/universal/NestedMultisigBase.sol#L124-L133):

```
    function _overrides(address _signerSafe, address _safe) internal view returns (Simulation.StateOverride[] memory) {
        Simulation.StateOverride[] memory simOverrides = _simulationOverrides();
        Simulation.StateOverride[] memory overrides = new Simulation.StateOverride[](2 + simOverrides.length);
        overrides[0] = _safeOverrides(_signerSafe, MULTICALL3_ADDRESS);
        overrides[1] = _safeOverrides(_safe, address(0));
        for (uint256 i = 0; i < simOverrides.length; i++) {
            overrides[i + 2] = simOverrides[i];
        }
        return overrides;
    }
```

Notice how the `simOverrides` are appended to the end of the `overrides` array. This leads to undesirable behavior in Tenderly. I personally observed how ALL existing state overrides for a contract were completely overridden when a user specified even just a single override for the same contract. This behavior was confirmed by Tenderly: 

> Based on this, keys should always be unique and if two identical are provided only one will be taken into account, can not guarantee that it will always be the last one.

The reason base doesn’t hit this tenderly bug is because the use an environment variable “[SAFE_NONCE](https://github.com/base/contracts/blob/6ca464d118d2d33331ea44893f0930f753663f60/script/universal/MultisigBase.sol#L38)” to get the nonce for the simulation. It **ISN’T** added the to [_simulationOverrides](https://github.com/base/contracts/blob/6ca464d118d2d33331ea44893f0930f753663f60/script/universal/MultisigBase.sol#L212C14-L212C34) function. The “SAFE_NONCE” is added to the default overrides. They would hit this issue if a user tried to pass a state override for a parent or child save inside the `_simulationOverrides` function. 

We've been able to 'get away' with this for now because users aren't specifying complex state overrides. However, on the odd occasions we might've seen the Tenderly link break. 

### Solution
Get all state overrides for simulation. Combines default Tenderly overrides with user-defined overrides. User defined overrides either replace or append to the default overrides.

The image below is what it used to look like:
CleanShot 2025-04-02 at 11 00 08@2x

**EDIT:**
I've since edited this functionality and confirmed that it works with the pending unichain task: `src/improvements/tasks/eth/002-opcm-upgrade-v200`:

Without `[stateOverrides]` in the `config.toml`:
https://dashboard.tenderly.co/oplabs/eth-mainnet/simulator/14bbc382-1066-48d8-8a4c-f3b73803374f

Without `[stateOverrides]` in the `config.toml`:
https://dashboard.tenderly.co/oplabs/eth-mainnet/simulator/16ec8a64-ed51-4312-b6ca-f176cf4ffa9e

The state overrides were as follows: 
```
[stateOverrides]
0x6d5B183F538ABB8572F5cD17109c617b994D5833 = [
     {key = "0x0000000000000000000000000000000000000000000000000000000000000005", value = "0x000000000000000000000000000000000000000000000000000000000000001a"}
]
0xb0c4C487C5cf6d67807Bc2008c66fa7e2cE744EC = [
     {key = "0x0000000000000000000000000000000000000000000000000000000000000005", value = "0x0000000000000000000000000000000000000000000000000000000000000023"}
]
```


blmalone:
* **#819** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/superchain-ops/819?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#822** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/superchain-ops/822?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/ethereum-optimism/superchain-ops/822?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

blmalone:
Closing this PR as we aren't actually going to require having tenderly links in stacked simulations.

test/tasks/StateOverrideManager.t.sol
```diff
@@ -283,58 +283,7 @@ contract StateOverrideManagerUnitTest is Test {
         removeFile(fileName);
     }
 
-    function testSanitizeOverrides_NoOverrides() public {
```
blmalone: don't need this anymore.

maurelian (dismissed):

maurelian (approved):

mds1 (approved):

PR #826, opened 2025-04-01
fix: superchain-registry dep update

blmalone:
Our build is failing so we're going to update the superchain-registry sub module to fix it. 

It's safe to update the registry because the change in the addresses.json file is expected. Specifically it is the superchainConfig address for unichain sepolia getting updated after a recent superchain-ops upgrade.


blmalone:
Closing in favor of https://github.com/ethereum-optimism/superchain-ops/pull/823 - for this pr to be merged it needs the fix from 823 and for 823 to be merged it needs this fix.

PR #840, opened 2025-04-04
refactor: cleaner terminal output for signers

blmalone:
When a signer calls simulate, there is far to much log output to the terminal. They should be looking at only the information that they need to know about. 

This PR performs some trivial refactors and fixes natspec to be in accordance with Foundrys best practices https://book.getfoundry.sh/guides/best-practices#comments


src/improvements/tasks/MultisigTask.sol
```diff
@@ -207,11 +207,6 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager {
     }
 
     /// @notice Runs the task with the given configuration file path.
-    /// Sets the address registry, initializes and simulates the single multisig
```
blmalone, line 210: Lots of these comments are out dated and not using foundrys best practices for natspec. https://book.getfoundry.sh/guides/best-practices#comments

smartcontracts (approved):

PR #841, opened 2025-04-04
refactor: task developer mode

blmalone:
Introduce a non breaking change to the simulate command on single and nested justfiles. 

Task developers not have to manually pass though the verbose flag to be true in order to get more output in the terminal. 

e.g.
for task devs:
```
SIMULATE_WITHOUT_LEDGER=1 just --dotenv-path $(pwd)/.env --justfile ../../../nested.just simulate child-safe-1 0 true
```

vs 

for signers:
```
SIMULATE_WITHOUT_LEDGER=1 just --dotenv-path $(pwd)/.env --justfile ../../../nested.just simulate child-safe-1
```


nested.just
```diff
@@ -284,20 +284,3 @@ execute hdPath='0':
     --sender ${sender} \
     --sig "runJson(string)" \
     ${bundlePath}
-
-simulated-run hdPath='0':
```
blmalone, line 288: We don't use this and I don't see when we would use it.

src/improvements/nested.just
```diff
@@ -140,18 +148,3 @@ execute hdPath='0':
     --sig "executeRun(string,bytes)" \
     ${config} \
     "0x"
-
-simulated-run hdPath='0':
```
blmalone, line 144: We don't use this and I don't see when we would use it.

blmalone:
Closing in favor of: https://github.com/ethereum-optimism/superchain-ops/pull/912

PR #869, opened 2025-04-11
feat: Added in link validation

ControlCplusControlV:
**Description**

Added in a go tool which looks up a given address and tells you where it is from and what it actually is, right now it uses the superchain-registry currently installed in `lib` as a submodule. It also has the ability to link out to github where that address comes from, with line numbers.

I have then added it to our scripts so that when we generate a state diff all addresses are now properly linked.

**Tests**

Now after running 
```
006-opcm-upgrade-v300

SIMULATE_WITHOUT_LEDGER=1 just --dotenv-path $(pwd)/.env --justfile ../../../nested.just simulate child-safe-2
```

The state diff produced is properly linked 

```
### [`0xf272670eb55e895584501d564AfEB048bEd26194`](https://github.com/ethereum-optimism/superchain-registry/blob/4791054602a56a72705ac7ecd601433c9c955791/superchain/configs/sepolia/base.toml#L60)  (SystemConfig) - Chain ID: 84532

- **Key:**          `0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc`
  - **Decoded Kind:**      `address`
  - **Before:** `0x760C48C62A85045A6B69f07F4a9f22868659CbCc`
  - **After:** [`0x340f923e5c7cbb2171146f64169ec9d5a9ffe647`](https://github.com/ethereum-optimism/superchain-registry/blob/4791054602a56a72705ac7ecd601433c9c955791/validation/standard/standard-versions-mainnet.toml#L9)
  - **Summary:**           ERC-1967 implementation slot
  - **Detail:**            Standard slot for storing the implementation address in a proxy contract that follows the ERC-1967 standard.

**<TODO: Insert links for this state change then remove this line.>**

  ---
  ```
  
  Please let me know if there is a better way to test this, I also know I will need to make changes to the justfile for the installation of it as a go tool / building the binary.

**Additional context**

Linking was taking very long when making upgrades and so I cut it down in size.

**Metadata**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -316,7 +316,16 @@ jobs:
       - utils/checkout-with-mise
       - run:
           name: "Run simulation sequence for network: <<parameters.network>> | tasks: <<parameters.tasks>>"
-          command: git submodule update; just clean; just install && echo "yes" | bash ./script/utils/sim-sequence.sh <<parameters.network>> "<<parameters.tasks>>" "<<parameters.block_number>>"
+       
+  whois_build:
+    docker:
+      - image: <<pipeline.parameters.default_docker_image>>
+    steps:
+      - utils/checkout-with-mise
+      - run:
+          name: op-whois build
+          command: |
+            just go-install
 
   forge_build:
     docker:
--- a/justfile
+++ b/justfile
@@ -17,6 +17,13 @@ install-eip712sign:
   mkdir -p bin || true
   GOBIN="$REPO_ROOT/bin" go install github.com/base/eip712sign@v0.0.11
 
+install-whois:
+    #!/usr/bin/env bash
+    REPO_ROOT=`git rev-parse --show-toplevel`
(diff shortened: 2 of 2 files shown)


maurelian:
@JosepBove IIRC you wrote something similar, but maybe it's not running in CI at the moment?

blmalone:
I think the general approach should be:
1. Raise an issue with a clear problem statement and proposed spec
2. Get buy-in
3. Then raise a PR

(That said, if the problem is well understood, going straight to a PR is fine.)

A few quick thoughts on this PR specifically:
1. Automatically linking to addresses is a nice touch.
2. I'd prefer not to add the Go toolchain to the new superchain-ops repo.
3. For Foundry FFI, we've been using bash scripts consistently.
4. This should include unit tests.
5. I'm a bit concerned about maintainability. This is tightly coupled to the scr, which will likely mean having to do a bunch of work to keep it functional.

PR #874, opened 2025-04-11
refactor: renaming sepolia tasks with correct task numbers.

blmalone:
We will encounter issues down the road if we have duplicate task names. This PR updates the task names so they are incrementing by one.


maurelian (dismissed):

Ethnical (approved):
LGTM! Thanks for the rename sorry for the collision!

maurelian (approved):

PR #877, opened 2025-04-11
fix: Update Foundry dependency to include fix from Foundry PR #10069

blmalone:
1. Updated the foundry dependency to: [7a7ad4ea7282c30477d7fc0f39bf631acde7e7bd](https://github.com/foundry-rs/foundry/releases/tag/nightly-7a7ad4ea7282c30477d7fc0f39bf631acde7e7bd). This is to include a fix we've been waiting for to make writing templates work with delegate calls. I opened [this issue](https://github.com/foundry-rs/foundry/issues/9990) on Foundry a number of weeks ago.
2. Included a related fix to the _isValidAction function: https://github.com/ethereum-optimism/superchain-ops/pull/701
3. Added tests to the `_isValidAction` function as they didn't exist before.

### DIFF
--- a/src/improvements/template/OPCMUpdatePrestateV300.sol
+++ b/src/improvements/template/OPCMUpdatePrestateV300.sol
@@ -84,14 +84,10 @@ contract OPCMUpdatePrestateV300 is OPCMTaskBase {
             });
         }
 
-        // We expect this call to revert when invoked from the _build() function.
-        // The _build function's job is to record the high-level calls that will later be executed by the L1 Proxy Admin Owner safe.
-        // It is not responsible for ensuring correct execution — that responsibility lies with the 'execTransaction' function in MultisigTask.sol.
-        (bool success,) = OPCM.call(abi.encodeWithSelector(IOPCMPrestateUpdate.updatePrestate.selector, opChainConfigs));
-        require(
-            !success,
-            "OPCMUpdatePrestateV300: Call unexpectedly succeeded; expected revert due to non-delegatecall in _build."
-        );
+        vm.startPrank(parentMultisig, true);
+        (bool success,) =
+            OPCM.delegatecall(abi.encodeWithSelector(IOPCMPrestateUpdate.updatePrestate.selector, opChainConfigs));
+        require(success, "OPCMUpdatePrestateV300: Delegatecall failed.");
     }
 
     /// @notice This method performs all validations and assertions that verify the calls executed as expected.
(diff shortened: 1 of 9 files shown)


src/improvements/template/OPCMUpdatePrestateV300.sol
```diff
@@ -84,14 +84,9 @@ contract OPCMUpdatePrestateV300 is OPCMTaskBase {
...
-        // The _build function's job is to record the high-level calls that will later be executed by the L1 Proxy Admin Owner safe.
-        // It is not responsible for ensuring correct execution — that responsibility lies with the 'execTransaction' function in MultisigTask.sol.
-        (bool success,) = OPCM.call(abi.encodeWithSelector(IOPCMPrestateUpdate.updatePrestate.selector, opChainConfigs));
-        require(
-            !success,
-            "OPCMUpdatePrestateV300: Call unexpectedly succeeded; expected revert due to non-delegatecall in _build."
-        );
+        (bool success,) =
+            OPCM.delegatecall(abi.encodeWithSelector(IOPCMPrestateUpdate.updatePrestate.selector, opChainConfigs));
+        require(success, "OPCMUpdatePrestateV300: Delegatecall failed.");
```
sebastianst, line 89: but now you could change this back to a regular call, right?

blmalone: No, we don't want this to be `call`. We want this to be `delegatecall`. It's subtle, but this is because for upgrades that use OPCM, we use [Multicall3Delegatecall](https://github.com/ethereum-optimism/superchain-ops/blob/main/src/Multicall3Delegatecall.sol). This means, the transaction that we send from the safe, instead of going through the regular Multicall3 contract (which performs `call` ops), goes through Multicall3Delegatecall (which performs `delegatecall` ops). 

No, we don't want this to be a `call`, it should be a `delegatecall`. For upgrades that use OPCM, we rely on [Multicall3Delegatecall](https://github.com/ethereum-optimism/superchain-ops/blob/main/src/Multicall3Delegatecall.sol). This means the transaction sent from the Safe doesn't go through the standard Multicall3 contract (which uses `call`), but instead through Multicall3Delegatecall (which uses `delegatecall`). 
Before writing code inside the `_build` function, keep the following in mind:
1. Which Multicall contract does this upgrade use?
2. Does it use `call` or `delegatecall`?
3. Make all function calls using the method from step 2.

See [this diagram](https://github.com/ethereum-optimism/superchain-ops/blob/main/NESTED-VALIDATION.md#visualisation-of-state-overrides-for-nested-execution) for a visual on the full flow of OPCM transactions.

AmadiMichael: is this documented in the code, maybe above the fn? i think this would be helpful info for someone new to the codebase

blmalone: Good call, added natspec to this function and added to the boilerplate templates so all new templates will contain this natspec. https://github.com/ethereum-optimism/superchain-ops/pull/877/commits/954c1a923ac4bf7058a5995e862ad45f160a4ea3

AmadiMichael (approved):

PR #891, opened 2025-04-15
feat: opcm v300 add errors as config

blmalone:
Like OPCMUpdatePrestateV300 template, we're requiring that tasks pass their error strings through their config files.


src/improvements/nested.just
```diff
@@ -140,18 +143,3 @@ execute hdPath='0':
     --sig "executeRun(string,bytes)" \
     ${config} \
     "0x"
-
-simulated-run hdPath='0':
```
blmalone, line 144: We don't need this and it's not used so I'm removing it as part of this PR.

PR #919, opened 2025-04-21
feat: mvp of normalized state diff

mds1:
Adds a `normalizedStateDiffHash` method to the `AccountAccessParser` library return a normalized state diff hash for a recorded state diff. See the comments above that method for the spec.

There is some currently unimplemented functionality that can be done in a follow up PR, and this puts the scaffolding in place to align on how we normalize state diffs

### DIFF
--- a/src/libraries/AccountAccessParser.sol
+++ b/src/libraries/AccountAccessParser.sol
@@ -78,6 +78,7 @@ library AccountAccessParser {
     }
 
     // Temporary struct used during deduplication.
+    // TODO Rename this, as it's also now used in `normalizedStateDiffHash`.
     struct TempStateChange {
         address who;
         bytes32 slot;
@@ -342,6 +343,10 @@ library AccountAccessParser {
         // which is a characteristic of an approve hash operation. We assume that if such a storage
         // change is seen in a very large slot, it is an approve hash operation. Setting modules
         // and guards in storage would not result in the slot having a value of 1.
+        // TODO Consider making this more robust to compute slots, if it adds value. In the safe
+        // storage layout, the only other mapping that sets slots to 1 is the `signedMessages`
+        // mapping but we don't use that. (The owners linked list can also have a value of 1 for the
+        // sentinel owner but that is only written when the safe is originally setup)
         return _diff.oldValue == bytes32(0) && _diff.newValue == bytes32(uint256(1))
             && _diff.slot > bytes32(uint256(0x1000000000));
     }
--- a/src/libraries/AccountAccessParser.sol
+++ b/src/libraries/AccountAccessParser.sol
@@ -119,6 +119,8 @@ library AccountAccessParser {
     bytes32 internal constant GAS_PAYING_TOKEN_SLOT = bytes32(uint256(keccak256("opstack.gaspayingtoken")) - 1);
     bytes32 internal constant GAS_PAYING_TOKEN_NAME_SLOT = bytes32(uint256(keccak256("opstack.gaspayingtokenname")) - 1);
     bytes32 internal constant GAS_PAYING_TOKEN_SYMBOL_SLOT = bytes32(uint256(keccak256("opstack.gaspayingtokensymbol")) - 1);
+
+    bytes32 internal constant GNOSIS_SAFE_NONCE_SLOT = bytes32(uint256(5));
     // forgefmt: disable-end
 
     modifier noGasMetering() {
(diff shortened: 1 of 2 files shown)


src/libraries/AccountAccessParser.sol
```diff
@@ -239,6 +241,127 @@ library AccountAccessParser {
...
+        // change is seen in a very large slot, it is an approve hash operation. Setting modules
+        // and guards in storage would not result in the slot having a value of 1.
+        return _diff.oldValue == bytes32(0) && _diff.newValue == bytes32(uint256(1))
+            && _diff.slot > bytes32(uint256(0x1000000000));
+    }
+
+    /// @notice Checks if the storage slot contains a timestamp that should be normalized
+    function slotContainsTimestamp(address _account, StateDiff memory _diff) internal pure returns (bool) {
+        // TODO Check out the storage layout snapshots and hardcode the slots that contain timestamps.
+        // We will have to call getter methods on _account to infer what protocol contract it is.
```
mds1, line 352: This is the core missing piece to be done in a follow up PR

src/libraries/AccountAccessParser.sol
```diff
@@ -239,6 +241,127 @@ library AccountAccessParser {
...
+    /// @notice Checks if the state diff represents an EOA nonce increment
+    function isEOANonceIncrement(address _account, StateDiff memory _diff) internal view returns (bool) {
+        uint256 codeSize = _account.code.length;
+        return codeSize == 0 && _diff.slot == bytes32(0) && uint256(_diff.newValue) == uint256(_diff.oldValue) + 1;
+    }
+
+    /// @notice Checks if the state diff represents a Gnosis Safe nonce increment
+    function isGnosisSafeNonceIncrement(StateDiff memory _diff) internal pure returns (bool) {
+        // In Gnosis Safe, the nonce is stored at slot 5. See `GnosisSafeStorage.sol` to verify.
+        return _diff.slot == GNOSIS_SAFE_NONCE_SLOT && uint256(_diff.newValue) == uint256(_diff.oldValue) + 1;
```
blmalone, line 334: I think this is a stronger check right? Someone might have a state override applied that's > 1 ahead of the current nonce (for whatever reason). 
```suggestion
        return _diff.slot == GNOSIS_SAFE_NONCE_SLOT && uint256(_diff.newValue) >= uint256(_diff.oldValue) + 1;
```

src/libraries/AccountAccessParser.sol
```diff
@@ -239,6 +241,127 @@ library AccountAccessParser {
...
+
+    /// @notice Checks if the state diff represents setting an approve hash in a Gnosis Safe
+    function isGnosisSafeApproveHash(StateDiff memory _diff) internal pure returns (bool) {
+        // In Gnosis Safe, approvedHashes is a mapping at slot 8
+        // The slot for a specific hash approval is calculated with the keccak256 hash of the address
+        // and slot. We can't know the specific slot value here, but we can detect a change from 0 to 1
+        // which is a characteristic of an approve hash operation. We assume that if such a storage
+        // change is seen in a very large slot, it is an approve hash operation. Setting modules
+        // and guards in storage would not result in the slot having a value of 1.
+        return _diff.oldValue == bytes32(0) && _diff.newValue == bytes32(uint256(1))
```
blmalone, line 345: I'm thinking maybe you can compute the slot here by requiring the parent multisig address as an argument to the ``normalizedStateDiffHash` function. You can see I've implemented similar logic already in `printApproveHashInfo`.

src/libraries/AccountAccessParser.sol
mds1: I believe the state override would get applied at the start, before the increment, so we'd still see the nonce only increment by 1?

src/libraries/AccountAccessParser.sol
```diff
@@ -239,6 +241,127 @@ library AccountAccessParser {
...
+        // Create a temporary array to store normalized state changes.
+        // We set the size to 1000 because we will likely never have more than 1000 state changes.
+        uint256 maxStateChanges = 1000;
+        TempStateChange[] memory normalizedChanges = new TempStateChange[](maxStateChanges);
+        uint256 normalizedCount = 0;
+
+        // Process each account with storage writes.
+        for (uint256 i = 0; i < uniqueAddresses.length; i++) {
+            address account = uniqueAddresses[i];
+            StateDiff[] memory diffs = getStateDiffFor({accesses: _accountAccesses, who: account, _sort: false});
```
blmalone, line 274: I'm assuming the hash is still deterministic even when we don't sort at various places in this funciton.

src/libraries/AccountAccessParser.sol
mds1: Can we leave this for a future PR or do you think this adds value? Looking at the safe storage layout, the only other mapping that sets slots to 1 is the `signedMessages` mapping but we don't use that. (The owners linked list can also have a value of 1 for the sentinel owner but that is only written when the safe is originally setup)

src/libraries/AccountAccessParser.sol
mds1: The order of the state diff returned by forge is the order the diff occurs so that should be deterministic—that feels more stable than our sort oder

src/libraries/AccountAccessParser.sol
blmalone: Makes sense. Happy with either. If it's not implemented as part of this PR I think it need a 'TODO' then.

mds1: Added in 51a674c0

blmalone (approved):

PR #921, opened 2025-04-21
fix: more specific approve hash and liveness guard lastLive detection 

blmalone:
Fixing one of the TODOs in the AccountAccessParser for the normalization of state diffs.

### DIFF
--- a/src/libraries/AccountAccessParser.sol
+++ b/src/libraries/AccountAccessParser.sol
@@ -122,6 +122,8 @@ library AccountAccessParser {
 
     bytes32 internal constant GNOSIS_SAFE_NONCE_SLOT = bytes32(uint256(5));
     bytes32 internal constant GNOSIS_SAFE_APPROVE_HASHES_SLOT = bytes32(uint256(8));
+
+    bytes32 internal constant LIVENESS_GUARD_LAST_LIVE_SLOT = bytes32(uint256(0));
     // forgefmt: disable-end
 
     modifier noGasMetering() {
@@ -140,14 +142,14 @@ library AccountAccessParser {
     // ==============================================================
 
     /// @notice Convenience function that wraps decode and print together.
-    function decodeAndPrint(VmSafe.AccountAccess[] memory _accesses, address _multisig, bytes32 _txHash)
+    function decodeAndPrint(VmSafe.AccountAccess[] memory _accesses, address _parentMultisig, bytes32 _txHash)
         internal
         view
     {
         // We always want to sort all state diffs before printing them.
         (DecodedTransfer[] memory transfers, DecodedStateDiff[] memory stateDiffs) = decode(_accesses, true);
         if (!Utils.isFeatureEnabled("SIGNING_MODE_IN_PROGRESS")) {
-            print(transfers, stateDiffs, _multisig, _txHash);
+            print(transfers, stateDiffs, _parentMultisig, _txHash);
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/libraries/AccountAccessParser.sol
```diff
@@ -258,7 +259,11 @@ library AccountAccessParser {
     ///              timestamp, since some timestamps are packed into slots with other data.
     ///   4. The hash to return is computed as `keccak256(abi.encode(normalizedArray))`.
     /// @return bytes32 hash of the normalized state diff
-    function normalizedStateDiffHash(VmSafe.AccountAccess[] memory _accountAccesses) internal view returns (bytes32) {
+    function normalizedStateDiffHash(VmSafe.AccountAccess[] memory _accountAccesses, address multisig, bytes32 txHash)
```
mds1, line 262: Is `multisig` supposed to be the parent multisig or child multisig? would be great to have a more clear variable name there (and also prefix with leading underscores per our regular convention)
```suggestion
    function normalizedStateDiffHash(VmSafe.AccountAccess[] memory _accountAccesses, address _multisig, bytes32 _txHash)
```

src/libraries/AccountAccessParser.sol
```diff
@@ -335,19 +340,22 @@ library AccountAccessParser {
...
-        // which is a characteristic of an approve hash operation. We assume that if such a storage
-        // change is seen in a very large slot, it is an approve hash operation. Setting modules
-        // and guards in storage would not result in the slot having a value of 1.
-        // TODO Consider making this more robust to compute slots, if it adds value. In the safe
-        // storage layout, the only other mapping that sets slots to 1 is the `signedMessages`
-        // mapping but we don't use that. (The owners linked list can also have a value of 1 for the
-        // sentinel owner but that is only written when the safe is originally setup)
-        return _diff.oldValue == bytes32(0) && _diff.newValue == bytes32(uint256(1))
-            && _diff.slot > bytes32(uint256(0x1000000000));
+    function isGnosisSafeApproveHash(StateDiff memory _diff, address multisig, bytes32 _txHash)
```
mds1, line 343: same var name / underscore comment

src/libraries/AccountAccessParser.sol
```diff
@@ -335,19 +340,22 @@ library AccountAccessParser {
...
+        internal
+        view
+        returns (bool)
+    {
+        bytes32[] memory hashSlots = calculateApproveHashSlots(IGnosisSafe(multisig).getOwners(), _txHash);
```
mds1, line 348: Do we also need to add Multicall3 as an "owner" since we use use state overrides to make it an owner sometimes?

blmalone: We don't add the state variables for the local simulation, we only add those for the tenderly links. So no.

src/libraries/AccountAccessParser.sol
blmalone: https://github.com/ethereum-optimism/superchain-ops/pull/921/commits/c21496c65edf407d563cdfa08a32dcf028f636c6

src/libraries/AccountAccessParser.sol
blmalone: https://github.com/ethereum-optimism/superchain-ops/pull/921/commits/c21496c65edf407d563cdfa08a32dcf028f636c6

mds1 (approved):

PR #922, opened 2025-04-22
feat: Enable audit hashing and txverify

blmalone:
Tried to add similar formatting/colors as op-txverify. This PR is safe to merge as it is only cosmetic and should be fine for existing in progress tasks in the signing ceremony if CI passes. We can let signers know to ignore the tx verify link along with the hash for now. 

### Screenshots
These screenshots show how the output in the terminal looks for signers. 
CleanShot 2025-04-22 at 14 13 26@2x
CleanShot 2025-04-22 at 14 13 17@2x

### DIFF
--- a/src/improvements/tasks/MultisigTask.sol
+++ b/src/improvements/tasks/MultisigTask.sol
@@ -323,17 +323,24 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager {
     /// @notice Print the data to sign.
     function printEncodedTransactionData(bytes memory dataToSign) public pure {
         // logs required for using eip712sign binary to sign the data to sign with Ledger
-        console.log("\nData to sign:");
+        printTitle("DATA TO SIGN");
         console.log("vvvvvvvv");
         console.logBytes(dataToSign);
         console.log("^^^^^^^^\n");
 
-        console.log("---------- ATTENTION SIGNERS ----------");
+        printTitle("ATTENTION SIGNERS");
         console.log("Please verify that the 'Data to sign' displayed above matches:");
         console.log("1. The data shown in the Tenderly simulation.");
         console.log("2. The data shown on your hardware wallet.");
         console.log("This is a critical step. Do not skip this verification.");
-        console.log("---------------------------------------");
+    }
+
+    function printTitle(string memory title) private pure {
+        // forgefmt: disable-start
+        console.log("");
+        console.log(string.concat("\x1b[1m\x1b[36m", vm.toUppercase(title), "\x1b[0m"));
+        console.log(string.concat("\x1b[36m", unicode"━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━", "\x1b[0m"));
+        // forgefmt: disable-end
     }
 
     /// @notice Print the hash to approve by EOA for parent/root multisig.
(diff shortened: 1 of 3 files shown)


src/improvements/tasks/MultisigTask.sol
```diff
@@ -323,17 +323,24 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager {
...
         console.log("1. The data shown in the Tenderly simulation.");
         console.log("2. The data shown on your hardware wallet.");
         console.log("This is a critical step. Do not skip this verification.");
-        console.log("---------------------------------------");
+    }
+
+    function printTitle(string memory title) private pure {
+        // forgefmt: disable-start
+        console.log("");
+        console.log(string.concat("\x1b[1m\x1b[36m", vm.toUppercase(title), "\x1b[0m"));
```
mds1, line 341: I'd suggest using forge-std's `StdStyle.sol` library for readability

src/improvements/tasks/MultisigTask.sol
```diff
@@ -409,8 +413,9 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager {
         );
 
         string memory base64Json = Base64.encode(bytes(json));
+        printTitle("OP-TXVERIFY LINK");
         console.log(
-            "\nTo verify this transaction, run `op-txverify qr` on your machine, then open the following link on your mobile device: https://op-txverify.optimism.io/?tx=%s",
+            "To verify this transaction, run `op-txverify qr` on your machine, then open the following link on your mobile device: https://op-txverify.optimism.io/?tx=%s",
```
mds1, line 418: How do signers install op-txverify, given that it's in a private repo?

src/improvements/tasks/MultisigTask.sol
blmalone: huh, didn't know about this. I'll add it.

src/improvements/tasks/MultisigTask.sol
```diff
@@ -667,25 +672,48 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager {
...
+
+    /// @notice Prints the audit report information.
+    function printAuditReportInfo(VmSafe.AccountAccess[] memory _accAccesses, address _parentMultisig, bytes32 _txHash)
+        private
+        view
+    {
+        console.log("");
+        printTitle("AUDIT REPORT INFORMATION");
+        // forgefmt: disable-start
+        console.log("The normalized state diff hash MUST match the hash created by the state changes attested to in the audit report.");
```
mds1, line 704:
```suggestion
        console.log("The normalized state diff hash MUST match the hash created by the state changes attested to in the state diff audit report.");
```

src/improvements/tasks/MultisigTask.sol
blmalone: Need to figure that out still. I think it's okay to turn this feature on even if the worst case scenario happened where we didn't have signers use it for U15.

src/libraries/AccountAccessParser.sol
```diff
@@ -332,6 +306,30 @@ library AccountAccessParser {
...
+            if (isGnosisSafeNonceIncrement(diff) || isGnosisSafeApproveHash(diff, _parentMultisig, _txHash)) {
+                return false;
+            }
+            // 3. If the slot contains a timestamp, normalize it to zeroes.
+        } else if (slotContainsTimestamp(account, diff)) {
+            diff = normalizeTimestamp(diff);
+            // 4. If the slot is on the LivenessGuard, don't include it.
+        } else if (isLivenessGuardTimestamp(account, diff, _parentMultisig)) {
+            return false;
+        }
```
mds1, line 329: nit, formatting of comments looks odd here, perhaps we should move them inside their `else if` blocks since putting the comment above makes it get indented

blmalone: https://github.com/ethereum-optimism/superchain-ops/pull/922/commits/7f0898394f0bea4d1f3979e97b1070577ac837d8

src/improvements/tasks/MultisigTask.sol
blmalone: https://github.com/ethereum-optimism/superchain-ops/pull/922/commits/7f0898394f0bea4d1f3979e97b1070577ac837d8

mds1 (approved):

PR #936, opened 2025-04-25
Replace deprecated releases.json with fetcher.go reference

PixelPil0t1:
Replaced broken links to releases.json with references to fetcher.go, which now handles standard prestates.

The old releases.json file no longer exists in the repository, and fetcher.go is the updated source of truth.

If you have a better suggestion for referencing standard prestates, feel free to propose it.


blmalone:
Hey, as per our contributing policy we don't accept trivial changes in this repo.

PR #938, opened 2025-04-25
fix: Bug in AccountAccessParser resolved

blmalone:
Fixed bug that was causing `[FAIL: panic: array out-of-bounds access (0x32)]` because an assumption was being made that there could never be more storage writes than accesses.


Ethnical (dismissed):

src/libraries/AccountAccessParser.sol
```diff
@@ -271,10 +280,7 @@ library AccountAccessParser {
         // Get all storage writes as a state diff.
         address[] memory uniqueAddresses = getUniqueWrites({accesses: _accountAccesses, _sort: false});
```
Ethnical, line 282: NIT: we maybe want to have comment on top here like previously to explain this array?

blmalone: yea will add in

Ethnical (approved):

PR #944, opened 2025-04-29
task: eth/008-gas-params-op

bitwiseguy:
# Overview
Replacement for task `eth/006-gas-params-op`. After viewing grafana metrics, the platforms team opted for a less risky approach of first increasing the gasTarget from 10Mgas/block --> 15Mgas/block before going to 20Mgas/block. If, after a week, the network looks healthy and metrics indicate we have plenty of headroom at 15Mgas/block, we will create a follow-up task to increase from 15Mgas/block --> 20Mgas/block.

This is a task to update the gas params for op-mainnet by sending a transaction from the `FoundationUpgradeSafe` (which is the `SystemConfigOwner`) to op-mainnet's `SystemConfigProxy` contract.

# Details
The following change will be made for op-mainnet:

* gas target increasing from 10Mgas/block --> 15Mgas/block


blmalone (approved):
lgtm

PR #952, opened 2025-05-05
refactor(part1): Consolidate simulate command.

blmalone:
The simulate command right now looks something like: 

Nested:
```
SIMULATE_WITHOUT_LEDGER=1 just --dotenv-path $(pwd)/.env --justfile ../../../nested.just simulate foundation
```
Single:
```
SIMULATE_FROM_LEDGER=1 just --dotenv-path $(pwd)/.env --justfile ../../../single.just simulate
```

this PR simplifies the command to be:
Nested:
```
SIMULATE_WITHOUT_LEDGER=1 just simulate foundation
```
Single:
```
SIMULATE_FROM_LEDGER=1 just simulate
```
note: you'll still need to pass through the .env file if needed. 

We can do this for other commands and eventually delete both single.just and nested.just.

### DIFF
--- a/src/improvements/justfile
+++ b/src/improvements/justfile
@@ -178,20 +178,7 @@ simulate-stack NETWORK="" TASK="" OWNER_ADDRESS="0x00000000000000000000000000000
     }
 
     export FOUNDRY_PROFILE="${FOUNDRY_PROFILE:-default}"
-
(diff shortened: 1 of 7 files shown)


blmalone:
This functionality has been added in a separate PR.

PR #959, opened 2025-05-07
fix: Update Execution statuses for tasks.

blmalone:
We've executed 4 tasks as part of the recent signing ceremony. We must update their execution statuses.


Ethnical (approved):

Ethnical:
we executed 5 tasks but there is 4 status changed is that intended?

PR #961, opened 2025-05-07
feat: ensure signer is owner on safe before signing

blmalone:
It is possible to accidentally sign for a safe with the wrong key. This functionality performs a check before signing to make sure the signer is using the correct signing key.


Ethnical:
+1 really nice thank you!

Ethnical (approved):
Really nice! A small upgrade could be also to manage for `simulate`.

PR #962, opened 2025-05-08
sep: Add convenient reset ASR task

Inphi:
This is a convenient task template to simplify the process of resetting the `AnchorStateRegistry` after a devnet is spun up.


mds1:
@Inphi I believe this was a temporary task that we needed for the first devnet attempt. Since it uses the legacy superchain-ops style I'm going to close this, and if we need, we can create a template for this in the new superchain-ops format

PR #963, opened 2025-05-08
docs: installation steps & general improvements

blmalone:
Signers are having trouble with `mise` installation. I've added concrete steps on how to install `mise` as well as locking `mise` down to accepted versions.

Copying the approach used to lock down `mise` from here: https://github.com/ethereum-optimism/circleci-utils/pull/13


src/improvements/script/install-mise.sh
```diff
@@ -0,0 +1,295 @@
+#!/bin/sh
+set -eu
+
+# Sourced from https://mise.run/.
+# A local, static copy is maintained to reduce the risk of introducing a vulnerable `mise` version.
```
mds1, line 5: Can you add more comments here about this, e.g. if we want to update mise to the latest, does the code at https://mise.run/ change and we just copy that in here?

blmalone: Yes, exactly. I'll add this in.

CONTRIBUTING.md
```diff
@@ -28,10 +28,54 @@ working inside of the `optimism` directory.
...
+> ⚠️ **IMPORTANT**: **Do not** update `mise` to a newer version unless you're told to do so by the maintainers of this repository. We pin to specific allowed versions of `mise` to reduce the likelihood of installing a vulnerable version of `mise`.
 
-#### Trust the `mise.toml` file
+**Step 1: Install and activate `mise`**
+
+To run an installation of `mise` you **must use** the following script. We want a local, static version of the script to reduce the likelihood of installing a vulnerable version of `mise`.
+
+```bash
+./src/improvements/scripts/install-mise.sh
+```
```
mds1, line 39: Can we run mise installation as part of the root `just install` command, to work towards the UX being that `just install` sets up all dependencies and now you can execute tasks?

blmalone: but mise installs just?

CONTRIBUTING.md
```diff
@@ -28,10 +28,54 @@ working inside of the `optimism` directory.
...
+
+Verify the installation:
+
+```bash
+~/.local/bin/mise --version
+```
```
mds1, line 45: Similarly could we do this check automatically as part of `just install`?

CONTRIBUTING.md
mds1: oh true lol, maybe it needs to be a single `bash ./setup.sh` script then, but same idea — would be cool if we can make that work

mds1: but in general it does seem our installation/setup scripts need more cleanup than just in this file, because in the README "clone superchain-ops" comes after "install mise", even though you need to clone superchain-ops to run the mise script

image

blmalone: It's more like:

1. Run install mise
2. Add the mise eval state to the correct place on each users machine based on their shell and source it
3. mise install
4. just install

`mise` requires the user to manually add the eval line to their profile. It doesn't do this step automatically. This means the setup script would have to add this to the users profile. At this point, it sounds like a docker container with all of this setup would be the ideal flow.

blmalone: See the log output in then screenshots provided for the simple script. 
CleanShot 2025-05-08 at 11 56 15@2x
CleanShot 2025-05-08 at 11 56 25@2x

blmalone: This solution is 2 commands instead of the 1 you're talking about with `./setup.sh`:

CleanShot 2025-05-08 at 12 06 10@2x

src/improvements/README.md
```diff
@@ -57,7 +63,7 @@ SIMULATE_WITHOUT_LEDGER=1 just --dotenv-path $(pwd)/.env --justfile ../../../nes
 SIMULATE_WITHOUT_LEDGER=1 just --dotenv-path $(pwd)/.env --justfile ../../../single.just simulate
 ```
 
-4. Fill out the `README.md` and `VALIDATION.md` files.
+5. Fill out the `README.md` and `VALIDATION.md` files.
```
blmalone, line 66: TODO: Blaine add in a short section that explains adding state overrides to your tasks. This now happens for every tasks so it'll be useful to add .

also add in these small fixes:
Make sure to create VALIDATION.md file automatically via just new task
Give better error if someone tries to run a single task as nested.
Make README.md file that’ created from just new task be better. 
Remove timing section from README and any other sections that add extra steps.
Remove extra spacing from printed markdown title: 7-L60C59)  (ProxyAdmin) - Chain ID: 130

mds1 (approved):

PR #965, opened 2025-05-08
Add balance change whitelist

JosepBove:
Add the ability to configure contracts that will be accepted if their balance changes


src/improvements/tasks/MultisigTask.sol
```diff
@@ -294,7 +305,8 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager {
         childMultisig = optionalChildMultisig;
 
         config.allowedStorageKeys = _taskStorageWrites();
-        config.allowedStorageKeys.push(config.safeAddressString);
+        config.allowedStorageKeys.push(safeAddressString());
```
JosepBove, line 308:
```suggestion
        config.allowedStorageKeys.push(config.safeAddressString);
```

PR #971, opened 2025-05-09
Enable setting safeAddressString in config

maurelian:


src/improvements/tasks/MultisigTask.sol
```diff
@@ -311,6 +313,19 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager {
         vm.label(address(this), "MultisigTask");
     }
 
+    /// @notice Get the safe address string from the config file.
+    /// If the string is not found, use the value from the template.
+    function getOrReadSafeAddressString(string memory taskConfigFilePath) internal view returns (string memory) {
```
maurelian, line 318: I could have just override `safeAddressString()` with this logic, but that felt a bit sneaky to do.

blmalone, line 318: Makes sense. nit: was finding it difficult to know what it was doing from the name alone. Would something like `loadSafeAddressString` be better? Idk, `getOrRead` sounds like the same thing to me. Will leave it up to you.

blmalone (approved):

blmalone (approved):
lgtm

blmalone (approved):
lgtm, I like that we can now make the safe address string configurable!

PR #972, opened 2025-05-09
feat: TransferOwners Template

blmalone:
TransferL1Owners.sol template which performs 3 owner changes:
- L1PAO
- DGF
- DWETH


src/improvements/template/TransferOwners.sol
```diff
@@ -0,0 +1,166 @@
...
+    address public STORAGE_SETTER;
+
+    /// @notice PermissionedDWETH address.
+    address public permissionedWETH;
+
+    /// @notice PermissionlessDWETH address.
+    address public permissionlessWETH;
+
+    /// @notice Stores the chain information after setup.
+    SuperchainAddressRegistry.ChainInfo internal _activeChainInfo;
```
mds1, line 34:
```suggestion
    SuperchainAddressRegistry.ChainInfo internal activeChainInfo;
```

src/improvements/template/TransferOwners.sol
```diff
@@ -0,0 +1,166 @@
...
+        }
+    }
+
+    /// @notice Sets the initialized slot to zero for the given proxy using the StorageSetter contract.
+    function setInitializedToZero(ProxyAdmin _proxyAdmin, address _proxy) internal {
+        bytes32 ZERO = bytes32(uint256(0));
+        _proxyAdmin.upgradeAndCall(
+            payable(_proxy), STORAGE_SETTER, abi.encodeCall(StorageSetter.setBytes32, (ZERO, ZERO))
+        );
+    }
```
mds1, line 116: Based on the storage layouts of the DGF and DWETH this logic should be ok, but we might want to add safety checks here to be safe. The initializer is split into two bytes:
- First byte is `_initialized`
- Second byte is `_initializing`

We can assert that byte 0 in the slot is 1 and byte 2 in the slot is 0, and the rest of the data in the slot should be. In other words the full value of the slot should be `0x0000000000000000000000000000000000000000000000000000000000000001` otherwise we are deleting other data

src/improvements/template/TransferOwners.sol
```diff
@@ -0,0 +1,166 @@
...
+        permissionlessWETH = _getDWETH("PermissionlessWETH", _activeChainInfo.chainId);
+
+        // Get the current implementation of the DisputeGameFactory before the storage setter is set.
+        address originalDGFImpl = getEIP1967Impl(dgfProxy);
+        // Set initialized slot to zero.
+        setInitializedToZero(proxyAdmin, dgfProxy);
+        // Reinitialize the DisputeGameFactory with the new owner.
+        proxyAdmin.upgradeAndCall(
+            payable(dgfProxy), originalDGFImpl, abi.encodeCall(DisputeGameFactory.initialize, (newOwner))
+        );
```
mds1, line 82: DGF is ownable so we can just do a `transferOwnership` call and remove the need for the storage setter

src/improvements/template/TransferOwners.sol
```diff
@@ -0,0 +1,166 @@
...
+    /// @notice New owner address. This is unaliased.
+    address public newOwner;
+
+    /// @notice StorageSetter address.
+    address public STORAGE_SETTER;
+
+    /// @notice PermissionedDWETH address.
+    address public permissionedWETH;
+
+    /// @notice PermissionlessDWETH address.
```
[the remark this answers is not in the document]
blmalone: done https://github.com/ethereum-optimism/superchain-ops/pull/972/commits/8e6a1f5c1a32e401946e174e0002e3e02f2ed76b

src/improvements/template/TransferOwners.sol
blmalone: done https://github.com/ethereum-optimism/superchain-ops/pull/972/commits/8e6a1f5c1a32e401946e174e0002e3e02f2ed76b

src/improvements/template/TransferOwners.sol
```diff
@@ -0,0 +1,166 @@
...
+        newOwner = abi.decode(vm.parseToml(toml, ".newOwner"), (address));
+
+        // Only allow one chain to be modified at a time with this template.
+        SuperchainAddressRegistry.ChainInfo[] memory _parsedChains =
+            abi.decode(vm.parseToml(toml, ".l2chains"), (SuperchainAddressRegistry.ChainInfo[]));
+        require(_parsedChains.length == 1, "Must specify exactly one chain id to transfer ownership for");
+        _activeChainInfo = _parsedChains[0]; // Store the ChainInfo struct
+
+        STORAGE_SETTER = abi.decode(vm.parseToml(toml, ".addresses.StorageSetter"), (address));
+        require(address(STORAGE_SETTER).code.length > 0, "Incorrect StorageSetter - no code at address");
```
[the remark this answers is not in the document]
blmalone: done https://github.com/ethereum-optimism/superchain-ops/pull/972/commits/8e6a1f5c1a32e401946e174e0002e3e02f2ed76b

src/improvements/template/TransferOwners.sol
blmalone: because these are all ownable, I've opted to not do this anymore as discussed.

src/improvements/template/TransferOwners.sol
blmalone: done https://github.com/ethereum-optimism/superchain-ops/pull/972/commits/8e6a1f5c1a32e401946e174e0002e3e02f2ed76b

src/improvements/template/TransferOwners.sol
```diff
@@ -0,0 +1,166 @@
...
+    function _upgradeDWETH(ProxyAdmin _proxyAdmin, address _dweth, address _newOwner) internal {
+        if (_dweth != address(0)) {
+            address superchainConfig = DelayedWETH(_dweth).config();
+            address originalImpl = getEIP1967Impl(_dweth);
+            setInitializedToZero(_proxyAdmin, _dweth);
+            _proxyAdmin.upgradeAndCall(
+                payable(_dweth), originalImpl, abi.encodeCall(DelayedWETH.initialize, (_newOwner, superchainConfig))
+            );
+        }
+    }
```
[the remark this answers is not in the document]
blmalone: done https://github.com/ethereum-optimism/superchain-ops/pull/972/commits/8e6a1f5c1a32e401946e174e0002e3e02f2ed76b

src/libraries/AccountAccessParser.sol
```diff
@@ -541,6 +541,7 @@ library AccountAccessParser {
     ) internal view noGasMetering {
         console.log("");
         string memory line = unicode"━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━";
+        printCopyHelper("below");
```
blmalone, line 544: I moved this here because I actually want users to copy any value transfers into the VALIDATIONS file too. 
CleanShot 2025-05-15 at 13 59 49@2x

This change is minimally invasive and I'm happy to include it in this PR.

src/improvements/template/TransferOwners.sol
```diff
@@ -0,0 +1,137 @@
...
+        address dgfProxy = superchainAddrRegistry.getAddress("DisputeGameFactoryProxy", activeChainInfo.chainId);
+        permissionedWETH = _getDWETH("PermissionedWETH", activeChainInfo.chainId);
+        permissionlessWETH = _getDWETH("PermissionlessWETH", activeChainInfo.chainId);
+
+        // Transfer ownership of the DisputeGameFactory to the new owner.
+        performOwnershipTransfer(dgfProxy, newOwner);
+
+        // Transfer ownership of the PermissionedWETH to the new owner.
+        if (permissionedWETH != address(0)) {
+            performOwnershipTransfer(permissionedWETH, newOwner);
```
mds1, line 73: `performOwnershipTransfer` has a check so that it's a no-op if `_target == address(0)`, so we can remove either this check or the other check. I think it's a bit safer to remove the check in `performOwnershipTransfer` so that it forces us to explicitly call out when we expect the target might be empty (like you do here)

src/improvements/template/TransferOwners.sol
```diff
@@ -0,0 +1,137 @@
...
+        } else {
+            console.log("PermissionedWETH not found on chain %s not performing transfer", activeChainInfo.chainId);
```
mds1, line 75: nit
```suggestion
            console.log("PermissionedWETH not found on chain %s, not performing transfer", activeChainInfo.chainId);
```

src/improvements/template/TransferOwners.sol
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+        // Transfer ownership of the PermissionlessWETH to the new owner.
+        if (permissionlessWETH != address(0)) {
+            performOwnershipTransfer(permissionlessWETH, newOwner);
+        } else {
+            console.log("PermissionlessWETH not found on chain %s not performing transfer", activeChainInfo.chainId);
```
mds1, line 82: same nit
```suggestion
            console.log("PermissionlessWETH not found on chain %s, not performing transfer", activeChainInfo.chainId);
```

src/improvements/template/TransferOwners.sol
```diff
@@ -0,0 +1,137 @@
...
+        ProxyAdmin proxyAdmin = ProxyAdmin(superchainAddrRegistry.getAddress("ProxyAdmin", activeChainInfo.chainId));
+        address dgfProxy = superchainAddrRegistry.getAddress("DisputeGameFactoryProxy", activeChainInfo.chainId);
+        permissionedWETH = _getDWETH("PermissionedWETH", activeChainInfo.chainId);
+        permissionlessWETH = _getDWETH("PermissionlessWETH", activeChainInfo.chainId);
+
+        // Transfer ownership of the DisputeGameFactory to the new owner.
+        performOwnershipTransfer(dgfProxy, newOwner);
+
+        // Transfer ownership of the PermissionedWETH to the new owner.
+        if (permissionedWETH != address(0)) {
```
mds1, line 72: There should always be a permissioned WETH, if they are not on permissioned fault proofs I think we should have this task fail

src/improvements/template/TransferOwners.sol
```diff
@@ -0,0 +1,137 @@
...
+
+interface Ownable {
+    function transferOwnership(address newOwner) external;
+}
+
+interface DisputeGameFactory {
+    function owner() external view returns (address);
+}
+
+interface DelayedWETH {
```
mds1, line 135: naming convention nit
```suggestion
interface IOwnable {
    function transferOwnership(address newOwner) external;
}

interface IDisputeGameFactory {
    function owner() external view returns (address);
}

interface IDelayedWETH {
```

src/improvements/template/TransferOwners.sol
smartcontracts, line 72: This will break after U16 - in U16 DelayedWETH no longer has an owner, it uses the proxy admin owner

src/improvements/template/TransferOwners.sol
```diff
@@ -0,0 +1,137 @@
...
+            abi.decode(vm.parseToml(toml, ".l2chains"), (SuperchainAddressRegistry.ChainInfo[]));
+        require(_parsedChains.length == 1, "Must specify exactly one chain id to transfer ownership for");
+        activeChainInfo = _parsedChains[0]; // Store the ChainInfo struct
+    }
+
+    /// @notice Builds the actions for transferring ownership of the DisputeGameFactory, DWETH contracts and ProxyAdmin.
+    function _build() internal override {
+        ProxyAdmin proxyAdmin = ProxyAdmin(superchainAddrRegistry.getAddress("ProxyAdmin", activeChainInfo.chainId));
+        address dgfProxy = superchainAddrRegistry.getAddress("DisputeGameFactoryProxy", activeChainInfo.chainId);
+        permissionedWETH = _getDWETH("PermissionedWETH", activeChainInfo.chainId);
```
smartcontracts, line 65: Warning: need to make sure these addresses are accurate. I remember that previously these would go out of sync in the registry sometimes, idk if that has since been fixed

src/improvements/template/TransferOwners.sol
```diff
@@ -0,0 +1,137 @@
...
+        // Only allow one chain to be modified at a time with this template.
+        SuperchainAddressRegistry.ChainInfo[] memory _parsedChains =
+            abi.decode(vm.parseToml(toml, ".l2chains"), (SuperchainAddressRegistry.ChainInfo[]));
+        require(_parsedChains.length == 1, "Must specify exactly one chain id to transfer ownership for");
+        activeChainInfo = _parsedChains[0]; // Store the ChainInfo struct
+    }
+
+    /// @notice Builds the actions for transferring ownership of the DisputeGameFactory, DWETH contracts and ProxyAdmin.
+    function _build() internal override {
+        ProxyAdmin proxyAdmin = ProxyAdmin(superchainAddrRegistry.getAddress("ProxyAdmin", activeChainInfo.chainId));
```
smartcontracts, line 63: What about the Guardian and Challenger roles? Are those handled separately?

src/improvements/template/TransferOwners.sol
blmalone: Superchain-ops does onchain address discovery, so it grabs these from the live chain every time.

src/improvements/template/TransferOwners.sol
```diff
@@ -0,0 +1,137 @@
...
+        address dgfProxy = superchainAddrRegistry.getAddress("DisputeGameFactoryProxy", activeChainInfo.chainId);
+        permissionedWETH = _getDWETH("PermissionedWETH", activeChainInfo.chainId);
+        permissionlessWETH = _getDWETH("PermissionlessWETH", activeChainInfo.chainId);
```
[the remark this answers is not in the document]
blmalone: I'm going to remove these global variables.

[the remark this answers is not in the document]
blmalone: done https://github.com/ethereum-optimism/superchain-ops/pull/972/commits/c6f2d36a0cb93ce2e613dc8c6565ba88f1850d7c

src/improvements/template/TransferOwners.sol
blmalone: done https://github.com/ethereum-optimism/superchain-ops/pull/972/commits/c6f2d36a0cb93ce2e613dc8c6565ba88f1850d7c

src/improvements/template/TransferOwners.sol
blmalone: Makes sense. @mds1 I'm thinking of making this more generic, i.e. if the DelayedWETH contract has an owner function then attempt the owner transfer. If not, then it must be post U16 and skip the owner transfer.

mds1: that sounds good to me

src/improvements/template/TransferOwners.sol
mds1: good points, I would say:
- assert that their superchain config is already the standard one / standard guardian — I don't think we have any use cases for this template for chains that don't have the standard superchain config
- we should check with devrel what the plan is for the challengers to see if they need to include it here, I forget what the standard config requires (it used to be the foundation challenger but I think we loosened it to also allow a 1 of N setup where one of them is the foundation challenger)

src/improvements/template/TransferOwners.sol
blmalone: Added fix here: https://github.com/ethereum-optimism/superchain-ops/pull/972/commits/bbb89f7df5312da167d2beb858954c85e8cac3af

src/improvements/template/TransferOwners.sol
blmalone: From what I can tell, they are on the standard SuperchainConfig according to this: https://github.com/ethereum-optimism/superchain-registry/blob/d82a61168fd1d7ef522ed8e213ce23c853031495/superchain/configs/sepolia/unichain.toml#L62. 

Assuming this is the standard one: https://github.com/ethereum-optimism/superchain-registry/blob/main/superchain/configs/sepolia/superchain.toml#L3 

Should this be a check in the template?

blmalone: What are you thoughts on this approach? https://github.com/ethereum-optimism/superchain-ops/pull/972/commits/e4891e1c47b19bc168a0ab3dc836c7a86d43f99c

I'm assuming that OP Portal addresses always remain static and that they always use the standard config SuperchainConfig.

blmalone: this commit is probably clearer: https://github.com/ethereum-optimism/superchain-ops/pull/972/commits/dbe826c062c3c4c860a77f2109437a81d7d00135

src/improvements/template/SystemConfigGasParams.sol
```diff
@@ -45,8 +45,7 @@ contract SystemConfigGasParams is L2TaskBase {
         super._templateSetup(taskConfigFilePath);
 
         string memory tomlContent = vm.readFile(taskConfigFilePath);
-        SuperchainAddressRegistry.ChainInfo[] memory _chains =
-            abi.decode(vm.parseToml(tomlContent, ".l2chains"), (SuperchainAddressRegistry.ChainInfo[]));
+        SuperchainAddressRegistry.ChainInfo[] memory _chains = superchainAddrRegistry.getChains();
```
blmalone, line 48: we don't need to manually read the TOML, this is done for us.

src/improvements/template/TransferL2PAOFromL1.sol
```diff
@@ -57,8 +57,7 @@ contract TransferL2PAOFromL1 is L2TaskBase {
         aliasedNewOwner = AddressAliasHelper.applyL1ToL2Alias(newOwnerToAlias);
 
         // Only allow one chain to be modified at a time with this template.
-        SuperchainAddressRegistry.ChainInfo[] memory _chains =
-            abi.decode(vm.parseToml(toml, ".l2chains"), (SuperchainAddressRegistry.ChainInfo[]));
+        SuperchainAddressRegistry.ChainInfo[] memory _chains = superchainAddrRegistry.getChains();
```
blmalone, line 60: same as above

src/improvements/template/TransferOwners.sol
```diff
@@ -0,0 +1,181 @@
...
+    using stdToml for string;
+    using LibString for string;
+
+    /// @notice OP Mainnet chain info.
+    SuperchainAddressRegistry.ChainInfo internal opMainnetChainInfo =
+        SuperchainAddressRegistry.ChainInfo({chainId: 10, name: "OP Mainnet"});
+
+    /// @notice OP Sepolia chain info.
+    SuperchainAddressRegistry.ChainInfo internal opSepoliaChainInfo =
+        SuperchainAddressRegistry.ChainInfo({chainId: 11155420, name: "OP Sepolia Testnet"});
```
mds1, line 27: not a blocker, but could use `getChain` from `StdChains` to avoid hardcoding these

src/improvements/template/TransferOwners.sol
```diff
@@ -0,0 +1,181 @@
...
+        IDelayedWETH permissionlessWETH = _getDWETH("PermissionlessWETH", activeChainInfo.chainId);
+
+        // Transfer ownership of the DisputeGameFactory to the new owner.
+        performOwnershipTransfer(address(disputeGameFactory), newOwner);
+
+        // Check if PermissionedWETH exists and is ownable. If it is, transfer ownership to the new owner.
+        if (_isDWETHOwnable(permissionedWETH) && address(permissionedWETH) != address(0)) {
+            performOwnershipTransfer(address(permissionedWETH), newOwner);
+        } else {
+            console.log("PermissionedWETH not found on chain %s, not performing transfer", activeChainInfo.chainId);
```
mds1, line 84: nit, out of date comment
```suggestion
            console.log("PermissionedWETH not found or not ownable on chain %s, not performing transfer", activeChainInfo.chainId);
```

src/improvements/template/TransferOwners.sol
```diff
@@ -0,0 +1,181 @@
...
+        }
+
+        // Check if PermissionlessWETH exists and is ownable. If it is, transfer ownership to the new owner.
+        if (_isDWETHOwnable(permissionlessWETH) && address(permissionlessWETH) != address(0)) {
+            performOwnershipTransfer(address(permissionlessWETH), newOwner);
+        } else {
+            console.log("PermissionlessWETH not found on chain %s, not performing transfer", activeChainInfo.chainId);
```
mds1, line 91:
```suggestion
            console.log("PermissionlessWETH not found or not ownable on chain %s, not performing transfer", activeChainInfo.chainId);
```

src/improvements/SuperchainAddressRegistry.sol
```diff
@@ -210,6 +214,13 @@ contract SuperchainAddressRegistry is StdChains {
     // ======== Superchain address discovery functions ========
     // ========================================================
 
+    /// @notice After instantiation of this contract, you can continue to discover new chains by calling this function.
+    /// This makes addresses on these chains available by using the getAddress function.
+    function discoverNewChain(ChainInfo memory chain) public {
```
mds1, line 219: We should update the tests for this contract to also write a new chain before reading to assert we don't accidentally overwrite or modify anything when calling this

src/improvements/template/TransferOwners.sol
```diff
@@ -0,0 +1,181 @@
...
+    /// @notice Performs an ownership transfer for the given target. If the target is address(0) we will not perform
+    /// the transfer.
+    function performOwnershipTransfer(address _target, address _newOwner) internal {
+        IOwnable(_target).transferOwnership(_newOwner);
+    }
+
+    /// @notice Validates the SuperchainConfig address against the OP Mainnet or OP Sepolia chain.
+    function _validateSuperchainConfig(address _superchainConfig) internal {
+        // 'block.chainId' will be set to whatever the network the current rpc url is pointing to.
+        if (block.chainid == getChain("mainnet").chainId) {
```
mds1, line 145: nit, not a blocker, could dedupe some logic here if you want, e.g.
```solidity
        ChainInfo memory chainInfo;
        string memory chainName;

        if (block.chainid == getChain("mainnet").chainId) {
            chainInfo = opMainnetChainInfo;
            chainName = "OP Mainnet";
        } else if (block.chainid == getChain("sepolia").chainId) {
            chainInfo = opSepoliaChainInfo;
            chainName = "OP Sepolia";
        } else {
            revert("Unsupported chain id");
        }

        superchainAddrRegistry.discoverNewChain(chainInfo);
        address expectedSuperchainConfig = superchainAddrRegistry.getAddress("SuperchainConfig", chainInfo.chainId);
        require(
            _superchainConfig == expectedSuperchainConfig,
            string.concat("SuperchainConfig does not match ", chainName, "'s SuperchainConfig")
        );
```

src/improvements/template/boilerplate/OPCMTaskBase.template.sol
```diff
@@ -81,8 +81,11 @@ contract OPCMTaskBaseTemplate is OPCMTaskBase {
     /// - `Multicall3Delegatecall`:
     ///   If the template inherits from `OPCMTaskBase`, it uses the `Multicall3Delegatecall` contract.
     ///   In this case, calls to the target **must** use `delegatecall`, e.g.:
-    ///   `(bool success,) = OPCM.delegatecall(abi.encodeWithSelector(IOPContractsManager.upgrade, opChainConfigs));`
+    ///   `(bool success,) = OPCM.delegatecall(abi.encodeWithSelector(IOPContractsManager.upgrade, opChainConfigs));
+    /// WARNING: Any state written to in this function will be reverted after the build function has been run.
+    /// Do not rely on setting global variables in this function.
```
mds1, line 86: good safety feature to track is to assert this. Implementation could be something like:
- `vm.dumpState()` -> filter for storage in `address(this)`
- do the build method
- dump state again after build and `keccak256(initStateOfThis)) == keccak256(endStateOfThis)`

src/improvements/SuperchainAddressRegistry.sol
blmalone: done: https://github.com/ethereum-optimism/superchain-ops/pull/972/commits/7cf7034df87b699639fa18fb0fe8509ca64ee264

src/improvements/template/TransferOwners.sol
blmalone: done: https://github.com/ethereum-optimism/superchain-ops/pull/972/commits/7cf7034df87b699639fa18fb0fe8509ca64ee264

src/improvements/template/TransferOwners.sol
blmalone: resolving.

src/improvements/template/TransferOwners.sol
```diff
@@ -0,0 +1,181 @@
...
+        storageWrites[2] = "PermissionedWETH";
+        storageWrites[3] = "PermissionlessWETH";
+        return storageWrites;
+    }
+
+    /// @notice Sets up the template with the new owner from a TOML file.
+    function _templateSetup(string memory _taskConfigFilePath) internal override {
+        super._templateSetup(_taskConfigFilePath);
+        string memory toml = vm.readFile(_taskConfigFilePath);
+        newOwner = abi.decode(vm.parseToml(toml, ".newOwner"), (address));
```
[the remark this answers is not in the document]
blmalone: done: https://github.com/ethereum-optimism/superchain-ops/pull/972/commits/7cf7034df87b699639fa18fb0fe8509ca64ee264

src/improvements/template/TransferOwners.sol
blmalone: done: https://github.com/ethereum-optimism/superchain-ops/pull/972/commits/7cf7034df87b699639fa18fb0fe8509ca64ee264

mds1 (approved):

mds1 (dismissed):

mds1 (approved):

PR #976, opened 2025-05-13
refactor: circle ci cleanup and slack alert 

blmalone:
A bunch of old jobs can be removed from CirclCI and I'm adding a slack notification to the template regression tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,7 +40,7 @@ parameters:
     default: "__not_set__"
 
 commands:
-  notify-failures-on-develop:
+  notify-failures-on-main:
     description: "Notify Slack"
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -400,6 +352,8 @@ jobs:
           name: Make sure all templates can be simulated.
           command: |
             (cd src/improvements && just simulate-all-templates)
+      - notify-failures-on-main:
```
Ethnical, line 355: NIT: Just to make sure when you meant: 
>  I'm adding a slack notification to the template regression tests.
This is the only notification you want to add here?

Ethnical (approved):

Issue #851, opened 2025-04-08
Add support to call 'just approve' and 'just execute' without a ledger


maurelian:
Similar to how 'just simulate' supports SIMULATE_WITHOUT_LEDGER, we should add support to run 'just approve' and 'just execute' without requiring a ledger device for testing purposes.

PR #989, opened 2025-05-15
[Improvements] Sign & Approve & execute with private-key inputs for facilitators

Ethnical:
This PR introduce:
- A private-key input each time facilitator when to approve and execute in every file (this is way helpful in some cases of testing). 

https://github.com/user-attachments/assets/b7f6edc5-a995-4879-90b2-2b0f8c975f68

To do that I have added this function `get-keystore-private-key` that will be called each time sometimes set the flag `ledger=false`. By default this would be set to `true` and most of the signers will not notice any changes. 

To execute with a private-key instead of a ledger you can do now:  
de98f58c5a802db039a1b5ebd1b814e07b3b122d7726133d17726ded9a320107

To execute with a ledger as usual nothing changes we are using the same command no need to updates the runbook. 
```
just --dotenv-path $(pwd)/.env   --justfile ../../../nested.just sign council 0
```

```bash
# Helper function to handle keystore selection and private key extraction
get-keystore-private-key:
  #!/usr/bin/env bash
  echo "Enter your foundry keystore path (e.g. ~/.foundry/keystores/sep-test-private-key):" >&2
  keystorePath=$keystorePathBase$(cd ~/.foundry/keystores/ && fzf --header="Searching your keystore in $PWD" --prompt="Keystore for signing:")
  cd ${taskPath}
  echo "Keystore path: ${keystorePath}" >&2
  # Expand the home directory in the path
  keystorePath=$(echo ${keystorePath} | sed "s|^~|$HOME|")
  signer_private_key=$(cast wallet pk --keystore ${keystorePath})
  sender=$(cast wallet address --private-key ${signer_private_key})
  if [ -z "${signer_private_key}" ]; then
    echo "Error: Failed to get private key for keystore ${keystorePath}" >&2
    exit 1
  fi
  echo "${signer_private_key}"
```

This is safer, since a private-key leak could happen when we manipulate them often, so less manipulation and copy paste better is it! 

close https://github.com/ethereum-optimism/superchain-ops/issues/851


nested.just
```diff
@@ -253,20 +253,34 @@ approve whichSafe hdPath='0':
...
+      --broadcast \
+      --sender ${sender} \
+      --sig "approveJson(string,address,bytes)" \
+      ${bundlePath} \
+      ${safe} \
+      ${signatures}
+  else
+    echo "Ledger is false. Please enter your private key:"
+    read -s PRIVATE_KEY_INPUT
+    sender=$(cast wallet address --private-key ${PRIVATE_KEY_INPUT})
```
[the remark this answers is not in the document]
Ethnical: Yeah I like it too! But I think for now keeping the private-key directly is sufficient for this PR.
Next time I can maybe do something that is checking with an hardcoded keystore name?
like `safety-key` and if this is existing then you decrypt this one and execute?

src/improvements/single.just
```diff
@@ -73,35 +77,60 @@ sign hdPath='0':
...
+
+    echo "⏳ Task signing in progress. Some tasks take longer than others..."
+    forge build
+    # Using the eip712sign within the repo folder since eip712sign was installed there in ./justfile.
+    $(git rev-parse --show-toplevel)/bin/eip712sign --ledger --hd-paths "m/44'/60'/{{hdPath}}'/0/0" -- \
+    forge script ${script} \
+      --rpc-url ${rpcUrl} \
+      --sig "simulateRun(string)" ${config}
+  else
+    cd ${TASK_PATH}
```
maurelian, line 96: why do we need this in this branch but not the other?

src/improvements/single.just
```diff
@@ -73,35 +77,60 @@ sign hdPath='0':
...
+    forge script TaskManager --rpc-url ${rpcUrl} --sig "requireSignerOnSafe(address,string)" $signer $TASK_PATH
+
+    echo "⏳ Task signing in progress. Some tasks take longer than others..."
+    forge build
+    # Using the eip712sign within the repo folder since eip712sign was installed there in ./justfile.
+    $(git rev-parse --show-toplevel)/bin/eip712sign --private-key ${signer_private_key#0x} -- \
+    forge script ${script} \
+      --rpc-url ${rpcUrl} \
+      --sig "simulateRun(string)" ${config}
+  fi
```
maurelian, line 109: there is a lot of duplication in these two different paths that I think we can avoid. 

Here is one way of doing it. I think we should use an approach like this throughout the PR in order to minimize the diff and keep it more maintainable.

```
  # Get signer info based on ledger or keystore
  if {{ledger}}; then
    signer=$(cast wallet address --ledger --mnemonic-derivation-path "m/44'/60'/{{hdPath}}'/0/0")
    echo "Signing with: ${signer}"
    echo ""
    sign_args="--ledger --hd-paths m/44'/60'/{{hdPath}}'/0/0"
  else
    cd ${TASK_PATH}
    signer_private_key=$(just --justfile={{justfile()}} get-keystore-private-key)
    signer=$signer_private_key
    sign_args="--private-key ${signer_private_key#0x}"
  fi

  # Verify signer is owner on safe
  forge script TaskManager --rpc-url ${rpcUrl} --sig "requireSignerOnSafe(address,string)" $signer $TASK_PATH

  echo "⏳ Task signing in progress. Some tasks take longer than others..."
  forge build

  # Using the eip712sign within the repo folder since eip712sign was installed there in ./justfile.
  $(git rev-parse --show-toplevel)/bin/eip712sign $sign_args -- \
  forge script ${script} \
    --rpc-url ${rpcUrl} \
    --sig "simulateRun(string)" ${config}
```

src/improvements/single.just
maurelian: What I'm confused about though is: 
1. We only do this `cd` when in the `else` case, but not the `if {{ledger}}` case.
2. I don't see it in the previous code for this diff. Maybe the `config=${TASK_PATH}/config.toml` on line 70 above is handling it already?

Ethnical: Yes since there is a call to `just` for example you need to be in the correct folder: 

```
signer_private_key=$(just --justfile=${JUSTFILE_PATH} get-keystore-private-key)
```

src/improvements/single.just
Ethnical: agreed addressed this here: 
in commit: f11da19e724860176c9896439e0aac8c4c083c4a & 2056b3604dcc839de260807edf6048e8406a64a3

PR #1002, opened 2025-05-16
fix: Owner address should be Owner addresses for stacked simulations

blmalone:
You should be able too pass through an array when specifying owner addresses to simulate as. This is a bug that the code is not doing this.


JosepBove (dismissed):
LGTM

PR #1004, opened 2025-05-16
feat: allow task devs to supply custom addresses not in the superchain-registry

blmalone:
This is a first version of an approach that allows task developers to supply custom addresses to the superchain-ops repo that do not exist in the superchain registry. This is useful for many reasons including upgrading devnets that do not exist in the superchain-registry.


blmalone:
**Note to self:**  
Instead of adding a new variable to the `ChainInfo` struct for `.l2Chains`, we should introduce a new TOML property called `.l2ChainsLocal`.

The `.l2Chains` and `.l2ChainsLocal` properties must **not** coexist in the same `config.toml` file.  
- If `.l2Chains` is present, the task will perform onchain discovery as usual.  
- If `.l2ChainsLocal` is present, the task will expect each L2 chain entry to include a file path pointing to a JSON file (matching the structure of SCR's `addresses.json`). The task will read all addresses from the file and skip onchain discovery.

This approach avoids breaking changes and keeps the `config.toml` API clean.

PR #1014, opened 2025-05-20
feat: remove optimism portal resource params state diff for normalization hash

blmalone:
This is needed for the Unichain mainnet L2PAO task: https://github.com/ethereum-optimism/superchain-ops/pull/1007


src/libraries/AccountAccessParser.sol
```diff
@@ -334,10 +337,28 @@ library AccountAccessParser {
...
+        if (_diff.slot == OPTIMISM_PORTAL_RESOURCE_PARAMS_SLOT) {
+            uint256 newValUint = uint256(_diff.newValue);
+            IResourceMetering.ResourceParams memory newValue = IResourceMetering.ResourceParams({
+                prevBaseFee: uint128(newValUint),
+                prevBoughtGas: uint64(newValUint >> 128),
+                prevBlockNum: uint64(newValUint >> (128 + 64))
+            });
+            // If the current block number is equal to the new values prevBlockNum, then we should remove this state change.
+            return block.number == newValue.prevBlockNum;
+        }
```
mds1, line 358: Took me a minute to figure out what is going on here: All 3 resource param values are packed into one slot, so we decode them into their components. But the only one we end up caring about is `prevBlockNum`, so I think this is equivalent — is this right?
```solidity
if (_diff.slot == OPTIMISM_PORTAL_RESOURCE_PARAMS_SLOT) {
    // Extract prevBlockNum from the packed value. 
    uint256 prevBlockNum = uint64(uint256(_diff.newValue) >> (128 + 64));
    // If the current block number is equal to the new values prevBlockNum, then we should remove this
    // state change because it means we have a nondeterministic change based on block number at simulation time
    return block.number == newValue.prevBlockNum;
}
```

blmalone: Ah, that's much cleaner, good call!

blmalone: done: https://github.com/ethereum-optimism/superchain-ops/pull/1014/commits/bf00689f0c305f60a7eca128a3a1126ac4c71571

mds1 (approved):

PR #1018, opened 2025-05-21
add docs on how to support base's extra level of nesting

mds1:
This is how we can support Base's extra level of nesting without significant code changes


src/improvements/tasks/sep/016-U16-opcm-upgrade-v400-base/README.md
```diff
@@ -21,16 +21,77 @@ which reads the inputs from the [`config.toml`](./config.toml) file.
...
+- BaseNested is `0x646132A1667ca7aD00d36616AFBA1A28116C770A`
+- Base is `0x6AF0674791925f767060Dd52f7fB20984E8639d8` (on Sepolia, the Base and FND safes are the same)
+- SC is `0x5dfEB066334B67355A15dc9b67317fD2a2e1f77f`
+- `SignerN` represent arbitrary signers for Safes
+
+### Facilitator Instructions
+
+1. Simulate this task from the `BaseNested` 2/2 by running `SIMULATE_WITHOUT_LEDGER=1 just --dotenv-path $(pwd)/.env --justfile ../../../nested.just simulate base_nested`
+2. In the terminal logs, directly above the Domain Hash and Message Hash, this will log the Safe Transaction Hash.
+3. Run `just new task ApproveSafeHash` to create a new task, and update the config file so this Safe Transaction Hash is used.
```
mds1, line 58: This task template does not yet exist and needs to be created

src/improvements/tasks/sep/016-U16-opcm-upgrade-v400-base/README.md
```diff
@@ -21,16 +21,77 @@ which reads the inputs from the [`config.toml`](./config.toml) file.
...
+Once that is done, your instructions as a signer are as follows:
+
+1. In this directory, run the same command as step 1 above and save off the Safe Transaction Hash.
+2. Navigate to your TODO directory, and follow the instructions to sign as normal.
+3. When performing the validation step, additionally ensure the data you are signing should match the Safe Transaction Hash you saved off in step 1.
+4. Send the signature to your ceremony facilitator.
+
 ## Simulation
 
+TODO these simulation commands don't seem to work?
```
mds1, line 93: Need to get this working, so far I tried updating `config.toml` to look like this:

```toml
[addresses]
OPCM = "0x44c191ce5ce35131e703532af75fa9ca221e2398" # Freshly deployed on Sepolia
StandardValidatorV400 = "0x482160a136b7aA13d05441cDB6f09a61AF36b935" # Freshly deployed on Sepolia

# Signers for Base sepolia
l1pao="0x0fe884546476dDd290eC46318785046ef68a0BA9" # 2/2 between fnd and base_nested
fnd="0x6AF0674791925f767060Dd52f7fB20984E8639d8"
base_nested="0x646132A1667ca7aD00d36616AFBA1A28116C770A" # 2/2 between Base and SC
base="0x6AF0674791925f767060Dd52f7fB20984E8639d8" # Same as FND
sc="0x5dfEB066334B67355A15dc9b67317fD2a2e1f77f"
```

and then running the standard simulation commands. But in the foundry trace I see `ProxyAdmin::upgradeAndCall` fails with `Ownable: caller is not the owner`

mds1: ok, we can actually just use `just simulate-stack sep` so this should not be an issue

src/improvements/tasks/sep/016-U16-opcm-upgrade-v400-base/README.md
smartcontracts: Might be easier for task N to be the approve hash txn, task N+1 is the actual transaction

src/improvements/tasks/sep/016-U16-opcm-upgrade-v400-base/README.md
```diff
@@ -21,16 +21,77 @@ which reads the inputs from the [`config.toml`](./config.toml) file.
...
+where:
+
+- L1PAO is `0x0fe884546476dDd290eC46318785046ef68a0BA9`
+- FND is `0x6AF0674791925f767060Dd52f7fB20984E8639d8`
+- BaseNested is `0x646132A1667ca7aD00d36616AFBA1A28116C770A`
+- Base is `0x6AF0674791925f767060Dd52f7fB20984E8639d8` (on Sepolia, the Base and FND safes are the same)
+- SC is `0x5dfEB066334B67355A15dc9b67317fD2a2e1f77f`
+- `SignerN` represent arbitrary signers for Safes
+
+### Facilitator Instructions
```
maurelian, line 54: Is this the Facilitator or Task Creator's job?

src/improvements/tasks/sep/016-U16-opcm-upgrade-v400-base/README.md
```diff
@@ -21,16 +21,77 @@ which reads the inputs from the [`config.toml`](./config.toml) file.
...
 
+### Base and Security Council Signer Instructions
+
+If you are a base or security council signer, you will actually sign in this task directory: TODO.
+
+Your ceremony facilitator will have performed the first three steps listed in the [Facilitator Instructions](#facilitator-instructions) section.
+
+Once that is done, your instructions as a signer are as follows:
+
+1. In this directory, run the same command as step 1 above and save off the Safe Transaction Hash.
```
maurelian, line 86: Not clear to me what 'step 1 above' refers to

src/improvements/tasks/sep/016-U16-opcm-upgrade-v400-base/README.md
```diff
@@ -21,16 +21,77 @@ which reads the inputs from the [`config.toml`](./config.toml) file.
...
+2. Navigate to your TODO directory, and follow the instructions to sign as normal.
+3. When performing the validation step, additionally ensure the data you are signing should match the Safe Transaction Hash you saved off in step 1.
```
maurelian, line 88: Do these signers need to validate the diff in this task, or just in the approve hash task?

blmalone:
I think this is a solid first step toward fully supporting the Base nested safe.

It’s important to emphasize that signers on both the SC and Base safes must first validate the state diffs for the task executing on the L1PAO. Once that’s verified, they can safely rely on the resulting safe transaction hash when executing the Base nested safe task (which calls `approveHash`).

If I recall correctly, this new template should extend `SimpleTaskBase`, and as you've already outlined, all required addresses should be defined under the `[addresses]` table in the `config.toml` file.

src/improvements/tasks/sep/016-U16-opcm-upgrade-v400-base/README.md
```diff
@@ -21,16 +21,77 @@ which reads the inputs from the [`config.toml`](./config.toml) file.
...
+┌▽───────────▽┐                │
+BaseNested    │                │
+└┬────────────┘                │
+┌▽─────────────────────────────▽┐
+│L1PAO                          │
+└┬──────────────────────────────┘
+┌▽─────────┐
+│ProxyAdmin│
+└──────────┘
+```
```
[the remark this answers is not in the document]
mds1: I copied this diagram from the [Base code](https://github.com/base/contracts/blob/cab46f4c34f11e22640ec3073aa6f0b46cdaa1b7/script/universal/MultisigScript.sol#L92) for reference, but I like your mermaid diagram and have used that instead in 7d9995a3

src/improvements/tasks/sep/016-U16-opcm-upgrade-v400-base/README.md
mds1: In 7d9995a3 I added a sentence that says "The facilitator, or someone acting on behalf of the facilitator, must perform the following steps:". I can imagine the task developer (e.g. us) creating the core task, and someone else (e.g. Base's facilitator or a dev) creating their approve hash task, or us doing it for them

src/improvements/tasks/sep/016-U16-opcm-upgrade-v400-base/README.md
mds1: Is 7d9995a3 more clear?

src/improvements/tasks/sep/016-U16-opcm-upgrade-v400-base/README.md
mds1: Yes good point (cc @blmalone who mentioned this below also). Clarified in 7d9995a3

smartcontracts (approved):

PR #1023, opened 2025-05-21
Enforce roundtrip encoding on state overrides

maurelian:


src/improvements/tasks/StateOverrideManager.sol
```diff
@@ -154,8 +154,21 @@ abstract contract StateOverrideManager is CommonBase {
...
+            // This is a hacky form of type safety to make up for the lack of it in the toml parser.
+            bytes memory reencoded = abi.encode(storageOverrides);
+            require(
+                reencoded.length == tomlOverrides.length,
+                "StateOverrideManager: Failed to reencode overrides, ensure all values are encoded as bytes32 strings"
+            );
+            require(
+                keccak256(reencoded) == keccak256(tomlOverrides),
+                "StateOverrideManager: Failed to reencode overrides, ensure all values are encoded as bytes32 strings"
+            );
```
mds1, line 171: We can delete the first `require` that just checks length because the second check is stricter right?

src/improvements/tasks/StateOverrideManager.sol
```diff
@@ -154,8 +154,21 @@ abstract contract StateOverrideManager is CommonBase {
...
+            // Reencode the overrides back to bytes and ensure that the roundtrip encoding is the same as the original.
+            // This is a hacky form of type safety to make up for the lack of it in the toml parser.
+            bytes memory reencoded = abi.encode(storageOverrides);
+            require(
+                reencoded.length == tomlOverrides.length,
+                "StateOverrideManager: Failed to reencode overrides, ensure all values are encoded as bytes32 strings"
+            );
+            require(
+                keccak256(reencoded) == keccak256(tomlOverrides),
+                "StateOverrideManager: Failed to reencode overrides, ensure all values are encoded as bytes32 strings"
```
mds1, line 170: I feel like defaulting to decimal numbers is more readable, what do you think of this
```suggestion
                "StateOverrideManager: Failed to reencode overrides, ensure any decimal numbers are not in quotes"
```

mds1 (approved):

PR #1034, opened 2025-05-28
u16 stack sim ci

maurelian:


src/improvements/tasks/sep/013-U16-enable-dpm/config.toml
```diff
@@ -0,0 +1,7 @@
+# this is the file used to determine the network configuration
+templateName = "EnableDeputyPauseModuleTemplate"
+
+newModule = "0xc10dac07d477215a1ebebae1dd0221c1f5d241d2"
+safeAddressString = "GuardianSafe"
+foundationSafeString = "FoundationUpgradeSafe"
```
maurelian, line 6: just noting that this is correct: we're moving away from the FUS as the deputy

PR #1037, opened 2025-05-29
feat: add script to check migration result

smartcontracts:
Adds a new script that can be used to check the result of running the interop migration. Script was originally added to OPCM but because there's no dedicated U17 OPCM.


foundry.toml
```diff
@@ -14,6 +14,7 @@ remappings = [
   '@openzeppelin/contracts/=lib/openzeppelin-contracts/contracts',
   '@base-contracts/=lib/base-contracts/',
   '@eth-optimism-bedrock/=lib/optimism/packages/contracts-bedrock/',
+  '@op/=lib/optimism/packages/contracts-bedrock/',
```
Inphi, line 17: nit: an `@eth-optimism-bedrock` monorepo mapping already exists. Let's use that.

Inphi (approved):

script/CheckMigrate.s.sol
```diff
@@ -0,0 +1,85 @@
...
+    function anchorStateRegistry() external view returns (address);
+    function systemConfig() external view returns (ISystemConfig);
+    function guardian() external view returns (address);
+}
+
+/// @title CheckMigrate
+/// @notice Checks that the interop migration correctly performed basic migration steps.
+///         This script is used to verify that the migration was successful and that the
+///         contracts were correctly updated.
+contract CheckMigrate is Script {
```
mds1, line 23: Is the intended use to inherit from this in the template task that calls `opcm.migrate` and then call the `run` method? If so, maybe we just scaffold out that task template instead and inline this directly into the validation method

foundry.toml
smartcontracts: I added it because `eth-optimism-bedrock` is so unnecessarily long lol

script/CheckMigrate.s.sol
smartcontracts: Sort of, the reason it isn't being used like that is so that we can use this script when we run devnets that don't use `superchain-ops` tasks

mds1: Got it, a nit, but can we just add some natspec to make the expected usages of this script (devnet vs. superchain-ops task) self-documenting?

blmalone:
@smartcontracts is this something we will want to merge? If it's nearly over the line, I'm happy to review.

PR #954, opened 2025-05-05
Metal Sepolia to permissioned FPs

Wazabie:
Description
Adds a task to get Metal Sepolia to use permissioned fault proofs.

Adding initial files.

Only the input file is correct, the others need to be adjusted for this task.

PR #953, opened 2025-05-05
Mode Sepolia to permissioned FPs

Wazabie:
Description
Adds a task to get Mode Sepolia to use permissioned fault proofs.

Adding initial files.

Only the input file is correct, the others need to be adjusted for this task.

PR #1038, opened 2025-05-29
Metal Sepolia to permissioned FPs fix

Wazabie:
This PR supersedes #954 to have libs/submodules in a clean state.

Task's files are not changed compared to #954 besides message hashes being updated after hardcoding nonces


ajsutton (dismissed):
LGTM from a proofs perspective, but the simulate_sequence job is failing so someone with more familiarity with that job likely needs to chime in with what's going on there.

mds1 (dismissed):

mds1:
The simulate_sequence job is dependent on https://github.com/ethereum-optimism/superchain-ops/pull/953 being merged, since that task should execute first in the sequence. I have added #953 and this PR to the merge queue, which I suspect will pass

mds1 (approved):

PR #1043, opened 2025-05-30
Allow L1PAO to transfer weth contracts it is not the owner() of

maurelian:
Certain op chains were deployed with an incorrect configuration, such that the owner of the DelayedWeth contracts was the SystemConfigOwner rather than the L1PAO. This means that transferring ownership would require two tasks if it were done solely using
`weth.transferOwnership().

To work around that issue, we modify this template so that it uses the StorageSetter to set the owner slot.

### DIFF
--- a/src/improvements/template/TransferOwners.sol
+++ b/src/improvements/template/TransferOwners.sol
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT
 pragma solidity 0.8.15;
 
-import {ProxyAdmin} from "@eth-optimism-bedrock/src/universal/ProxyAdmin.sol";
+import {IProxyAdmin} from "@eth-optimism-bedrock/interfaces/universal/IProxyAdmin.sol";
 import {VmSafe} from "forge-std/Vm.sol";
 import {stdToml} from "forge-std/StdToml.sol";
 import {LibString} from "solady/utils/LibString.sol";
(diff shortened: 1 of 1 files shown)


src/improvements/template/TransferOwners.sol
```diff
@@ -138,6 +142,28 @@ contract TransferOwners is L2TaskBase {
         IOwnable(_target).transferOwnership(_newOwner);
     }
 
+    /// @notice Performs an ownership transfer by writing the new owner directly to the owner slot.
+    function performWethOwnershipTransfer(address _target, address _newOwner) internal {
+        _writeToProxy(_target, bytes32(uint256(51)), bytes32(uint256(uint160(_newOwner))));
```
mds1, line 147: Slot 0 is not the owner on the latest DelayedWETH , we should have a semver check for safety (maybe checking for the `owner()` method could work as well)

mds1: Nevermind, I see this is wrapped in a `_isDWETHOwnable` check

maurelian: I could move that check within this function to ensure they are tightly coupled, if so then I think the name would need to be improved, maybe `performTransferIfOwnable`?

mds1: That seems like a good idea to me, and I like the function name

maurelian: ef841a75b947

PR #1051, opened 2025-06-04
Fix(sim-sequence) by introducing a env flag

Ethnical:
This PR fix the sim-sequence to introduce a new flag to remove the check of the override.
And also fix BASE with the new addresses.


src/improvements/tasks/StateOverrideManager.sol
```diff
@@ -77,18 +78,20 @@ abstract contract StateOverrideManager is CommonBase {
...
-                        string.concat(
-                            "StateOverrideManager: User-defined nonce (",
-                            vm.toString(userDefinedNonce),
-                            ") is less than current actual nonce (",
-                            vm.toString(currentActualNonce),
-                            ") for contract: ",
-                            vm.toString(safeAddress)
-                        )
-                    );
+                    if (Utils.isFeatureEnabled("DISABLE_OVERRIDE_NONCE_CHECK")) {
```
blmalone: Can you put a comment in here describing where this env var would be set and why?

Ethnical: Done in commit: [4e9a407](https://github.com/ethereum-optimism/superchain-ops/pull/1051/commits/4e9a4074e888e8aa33c9781bba85e7a77fd32ad0)

blmalone: i don't see the comment in that commit.

src/improvements/nested.just
```diff
@@ -194,9 +197,11 @@ simulated-run hdPath='0':
     "0x"
 
 
-approvehash_in_anvil2 whichSafe parentSafe:
+sign_and_approve_in_anvil whichSafe parentSafe:
   #!/usr/bin/env bash
-  #set -x
+
+  # set -x
+  export DISABLE_OVERRIDE_NONCE_CHECK=1
```
maurelian, line 204: so, we're losing this safety entirely?

src/improvements/tasks/StateOverrideManager.sol
Ethnical: Oh you are right @blmalone I don't know what I have done here haha!

Ethnical: Ok, I assume that I mess with my git history anyway this is the real commit ae18754

maurelian (approved):

PR #1052, opened 2025-06-04
fix: normalized hashes fixed for asr proposal

blmalone:
Making sure the normalized hashes for all U16 tasks remain consistent.


blmalone:
closing in favor of later pr

PR #1069, opened 2025-06-09
fix: Better Logging for RPC URL

blmalone:
Need to print the foundry profile to make sure we're actually using the correct RPC url.


blmalone:
Closing as we won't want to merge this incase we log creds to terminal.

PR #1071, opened 2025-06-10
chore: fix the hdpath

Ethnical:
**Description**

**Tests**

**Additional context**

**Metadata**


blmalone:
Was there a bug in the hdPath?

blmalone (approved):

PR #1072, opened 2025-06-10
feat(stack-signing): Added signing functionality to stacked tasks.

blmalone:
Right now, stacked simulations do not support signing. This is a problem because it makes it hard to sign tasks that depend on each other. 
If we have the following dependency graph: `a --> b --> c`. `a` must be executed first and `c` last. To sign for `b`, `a` must be executed. But if they're in the same signing window, `a` won't be executed before `b`. We typically collect all the signatures first before executing any tasks in a signing ceremony.

This PR tries to reuse existing signing flows and integrates it with the StackedSimulator code.


src/improvements/justfile
```diff
@@ -163,20 +163,6 @@ simulate-all-templates:
         exit 1
     fi
 
-simulate-verify-task TASK="":
```
blmalone, line 166: This isn't used and I don't think we're going to use it going forward.

src/improvements/justfile
```diff
@@ -224,7 +196,45 @@ simulate-stack NETWORK="" TASK="" OWNER_ADDRESSES="[]":
...
+    root_dir=$(git rev-parse --show-toplevel)
+    root_just_file="${root_dir}/src/improvements/justfile"
+    
+    echo -e "\033[42m                                         \033[0m"
+    echo -e "\033[42m       superchain-ops sign task          \033[0m"
+    echo -e "\033[42m                                         \033[0m"
+
+    show_usage() {
+        echo
+        echo "Usage: just sign <network> <task> <owner-safe-name>"
```
blmalone: make owner-safe-name optional

src/libraries/MultisigTaskPrinter.sol
```diff
@@ -92,19 +92,24 @@ library MultisigTaskPrinter {
...
-    function printEncodedTransactionData(bytes memory dataToSign) internal pure {
+    function printEncodedTransactionData(bytes memory dataToSign) internal view {
         // NOTE: Do not change the vvvvvvvv and ^^^^^^^^ lines, as the eip712sign tool explicitly
         // looks for those specific lines to identify the data to sign.
         printTitle("DATA TO SIGN");
-        console.log("vvvvvvvv");
-        console.logBytes(dataToSign);
-        console.log("^^^^^^^^\n");
+        // 'PRINT_DATA_TO_SIGN' is set when the task is the last task in the stack.
+        if (Utils.isFeatureEnabled("PRINT_DATA_TO_SIGN")) {
```
blmalone: Can I change this to `SUPPRESS_DATA_TO_SIGN_PRINT`, and have it always print unless it's set to `true`. The only situation where it'll not be true is in stacked signing when it's not the last task.

src/improvements/justfile
```diff
@@ -259,3 +268,98 @@ test:
...
+_get-signer-args hdPath='0' ledger='true':
+    #!/usr/bin/env bash
+    set -euo pipefail
+    root_dir=$(git rev-parse --show-toplevel)
+    if {{ledger}}; then
+        hdpaths="m/44'/60'/{{hdPath}}'/0/0"
+        signer=$(cast wallet address --ledger --mnemonic-derivation-path $hdpaths)
+        signer_args="--ledger --hd-paths $hdpaths"
+    else
+        cd ${TASK_PATH}
```
blmalone: make sure this flow works.

maurelian (approved):

PR #1077, opened 2025-06-11
Fix Metal/Mode hash messages

Wazabie:
Updating the Metal and Mode FP upgrade hashes to the correct ones

### DIFF
--- a/tasks/eth/metal-002-fp-upgrade/VALIDATION.md
+++ b/tasks/eth/metal-002-fp-upgrade/VALIDATION.md
@@ -31,12 +31,12 @@ Additionally, please verify that for each contract:
 > ### Security Council
 >
 > - Domain Hash: `0xdf53d510b56e539b90b369ef08fce3631020fbf921e3136ea5f8747c20bce967`
-> - Message Hash: `0x6e4b8104d1fa100896d3c6be742be00b5fb1c923ed7696fa8213692977922aa5`
+> - Message Hash: `0x27d70ccc7da26274fa4470df0426343f1979199b593135af157db9815e54ff80`
 >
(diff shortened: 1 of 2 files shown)


tasks/eth/metal-002-fp-upgrade/VALIDATION.md
```diff
@@ -31,12 +31,12 @@ Additionally, please verify that for each contract:
...
 >
 > - Domain Hash: `0xdf53d510b56e539b90b369ef08fce3631020fbf921e3136ea5f8747c20bce967`
-> - Message Hash: `0x6e4b8104d1fa100896d3c6be742be00b5fb1c923ed7696fa8213692977922aa5`
+> - Message Hash: `0x27d70ccc7da26274fa4470df0426343f1979199b593135af157db9815e54ff80`
 >
 > ### Optimism Foundation
 >
 > - Domain Hash: `0xa4a9c312badf3fcaa05eafe5dc9bee8bd9316c78ee8b0bebe3115bb21b732672`
-> - Message Hash: `0x2e9b9c8942a41bb7417c475995a68eccf31980e45f57260f6d1a81c4b459eb77`
+> - Message Hash: `0x27d70ccc7da26274fa4470df0426343f1979199b593135af157db9815e54ff80`
```
mds1, line 39: Interestingly, the SC and FND safes have the same nonce right now. Therefore they end up with the same message hash, because the inputs to the message hash are:
```solidity
to,
value,
data,
operation,
safeTxGas,
// Payment info
baseGas,
gasPrice,
gasToken,
refundReceiver,
// Signature info
nonce
```
This is ok, because the domain separator ensures sigs can't be replayed across safes

PR #1084, opened 2025-06-13
SetDisputeGameImpl Template + test task

Wazabie:
This PR includes the SetDisputeGameImpl template and its test task


src/improvements/template/SetDisputeGameImpl.sol
```diff
@@ -0,0 +1,111 @@
...
+import {stdToml} from "forge-std/StdToml.sol";
+
+import {L2TaskBase} from "src/improvements/tasks/types/L2TaskBase.sol";
+import {SuperchainAddressRegistry} from "src/improvements/SuperchainAddressRegistry.sol";
+import {Action} from "src/libraries/MultisigTypes.sol";
+
+import {DisputeGameFactory} from "lib/optimism/packages/contracts-bedrock/src/dispute/DisputeGameFactory.sol";
+import {GameTypes} from "lib/optimism/packages/contracts-bedrock/src/dispute/lib/Types.sol";
+
+import {IFaultDisputeGame} from "lib/optimism/packages/contracts-bedrock/interfaces/dispute/IFaultDisputeGame.sol";
```
blmalone: Notice other templates (e.g. TransferOwners.sol) do not import interfaces from the monorepo and instead define them statically at the bottom on the file. Can you also do that with this template? This way, if we update the monorepo dependency and the interfaces change, your template won't break.

src/improvements/template/SetDisputeGameImpl.sol
```diff
@@ -0,0 +1,111 @@
...
+import {IPermissionedDisputeGame} from
```
blmalone: Same as the comment above.

src/improvements/template/SetDisputeGameImpl.sol
```diff
@@ -0,0 +1,111 @@
...
+
+    /// @notice Sets up the template with implementation configurations from a TOML file.
+    function _templateSetup(string memory taskConfigFilePath) internal override {
+        super._templateSetup(taskConfigFilePath);
+        string memory toml = vm.readFile(taskConfigFilePath);
+
+        GameImplConfig[] memory configs = abi.decode(toml.parseRaw(".gameImpls.configs"), (GameImplConfig[]));
+        for (uint256 i = 0; i < configs.length; i++) {
+            cfg[configs[i].chainId] = configs[i];
+        }
```
blmalone, line 57: At this point we probably want to add some validation to the FDG addresses. e.g. We shouldn't allow the zero address being set. Also we should do some sanity checking on the address that it's actually a FDG. We should lean on the proofs team here to get their input on what checks we would need.

src/improvements/template/SetDisputeGameImpl.sol
```diff
@@ -0,0 +1,111 @@
...
+            GameImplConfig memory c = cfg[chainId];
+
+            address dgf = superchainAddrRegistry.getAddress("DisputeGameFactoryProxy", chainId);
+            DisputeGameFactory factory = DisputeGameFactory(dgf);
+
+            // Assert FDG (CANNON) implementation matches TOML
+            assertEq(address(factory.gameImpls(GameTypes.CANNON)), c.fdgImpl);
+
+            // Assert PDG (PERMISSIONED_CANNON) implementation matches TOML
+            assertEq(address(factory.gameImpls(GameTypes.PERMISSIONED_CANNON)), c.pdgImpl);
```
blmalone: Same as the comment above, can we lean on the proofs team here to make sure we're performing enough checks on the resulting upgrade. I think being overly conservative here on the checks we want is a good first place to start.

src/improvements/template/SetDisputeGameImpl.sol
```diff
@@ -0,0 +1,111 @@
...
+        }
+    }
+
+    /// @notice Override to return a list of addresses that should not be checked for code length.
+    function getCodeExceptions() internal pure override returns (address[] memory) {
+        address[] memory codeExceptions = new address[](1);
+        codeExceptions[0] = address(0);
+        return codeExceptions;
+    }
```
blmalone:
```suggestion
    function getCodeExceptions() internal pure override returns (address[] memory) {
        return new address[](0);
    }
```

blmalone: I don't think you need to have the zero address as a code exception.

src/improvements/template/SetDisputeGameImpl.sol
Wazabie: Agreed on address validation. Actually we do have a task for Arena-Z and Swell where we need to reset the FDG implementation address to 0.

src/improvements/template/SetDisputeGameImpl.sol
Wazabie: As we need to reset FDG implementation for Arena-Z and Swell to the zero address, would we need this zero address as a code exception?

src/improvements/template/SetDisputeGameImpl.sol
sbvegan: Yeah we should allow the zero address for the FDG when running permissioned proofs. We're actually going to zero these out for Arena Z and Swell based on the proofs team guidance. We'll just add it back when they go permissionless.

blmalone: ok, sounds like we need 👍🏻

src/improvements/template/SetDisputeGameImpl.sol
Inphi: It would be good to sanity check the new game implementations. Can be done before or after at `_validate`.  At the very least, add the following checks:
- `fdgimpl.gameType() == CANNON`
- `pdgImpl.gameType() == PERMISSIONED_CANNON`
- check the `game.l2ChainId()`.

Then depending on the reason for setting the game implementation, we'll add a bunch of different checks. There are two main reasons for setting a game implementation:
1. Updating the prestate (80% of the time)
2. 3. Updating the VM (10% of the time usually in response to a new Go runtime version introduced)
3. Adding a new implementation that did not exist (9% of the time)
4. Changing the proposer/challenger of the PDG.

I'd focus on the first case. The other two cases are special enough that the task writer should handle additional checks for themselves.
If the user specifies that they're updating the prestate, then the following additional checks should be done:
- ensure that for both fdg and pdg implementations, the initial contract state is unaltered _except_ for the absolute prestate
- That includes: `l2ChainId()`, `version()`, `anchorStateRegistry()`, `maxGameDepth()`, `splitDepth()`, `maxClockDuration()`, `clockExtension()`, `vm()` for all game types.
- And for the permissioned dispute game: `proposer()` and `challenger()`

If the user is updating the VM, in addition to changing the prestate, then the above  "updating the prestate" checks apply, except for the `vm()` check.

src/improvements/template/SetDisputeGameImpl.sol
```diff
@@ -0,0 +1,211 @@
...
+            uint256 chainId = chains[i].chainId;
+            GameImplConfig memory c = cfg[chainId];
+
+            address dgf = superchainAddrRegistry.getAddress("DisputeGameFactoryProxy", chainId);
+            IDisputeGameFactory factory = IDisputeGameFactory(dgf);
+
+            // Set FDG (CANNON) implementation if TOML is different from current on-chain
+            address currentFDG = address(factory.gameImpls(CANNON));
+            if (currentFDG != c.fdgImpl) {
+                factory.setImplementation(CANNON, c.fdgImpl);
```
Inphi, line 73: Let's also enforce that `c.fdgImpl` and `c.pdgImpl` is non-zero, if an existing dispute game implementation exists.
Removing a dispute game implementation in effect should be prohibited.

src/improvements/template/SetDisputeGameImpl.sol
```diff
@@ -0,0 +1,211 @@
...
+
+            address dgf = superchainAddrRegistry.getAddress("DisputeGameFactoryProxy", chainId);
+            IDisputeGameFactory factory = IDisputeGameFactory(dgf);
+
+            // Always check that DisputeGameFactory points to the expected new implementations from TOML config file
+            require(address(factory.gameImpls(CANNON)) == c.fdgImpl, "FDG implementation mismatch");
+            require(address(factory.gameImpls(PERMISSIONED_CANNON)) == c.pdgImpl, "PDG implementation mismatch");
+
+            // Always check basic invariants on any nonzero FDG new implementation
+            if (c.fdgImpl != address(0)) {
```
Inphi, line 115: If non-zero, also check that `factory.initBonds(CANNON) != 0`.

src/improvements/template/SetDisputeGameImpl.sol
```diff
@@ -0,0 +1,211 @@
...
+                IFaultDisputeGame newFdg = IFaultDisputeGame(c.fdgImpl);
+                require(newFdg.gameType() == CANNON, "FDG: gameType not CANNON");
+                require(newFdg.l2ChainId() == chainId, "FDG: l2ChainId mismatch");
+            }
+
+            // Always check basic invariants on any nonzero PDG new implementation
+            if (c.pdgImpl != address(0)) {
```
Inphi, line 123: If non-zero, also check that `factory.initBonds(PERMISSIONED_CANNON) != 0`.

src/improvements/template/SetDisputeGameImpl.sol
```diff
@@ -0,0 +1,211 @@
...
+
+            // -- FDG detailed check (only for impl->impl upgrade, not 0->impl or impl->0) --
+            address prevFdgAddr = c.prevFdgImpl;
+            address newFdgAddr = c.fdgImpl;
+            if (prevFdgAddr != address(0) && newFdgAddr != address(0) && prevFdgAddr != newFdgAddr) {
+                IFaultDisputeGame prevFdg = IFaultDisputeGame(prevFdgAddr);
+                IFaultDisputeGame newFdg = IFaultDisputeGame(newFdgAddr);
+
+                // Check if prestate or vm changed
+                bool prestateChanged = prevFdg.anchorStateRegistry() != newFdg.anchorStateRegistry();
```
Inphi:
```suggestion
                bool prestateChanged = prevFdg.absolutePrestate() != newFdg.absolutePrestate();
```

src/improvements/template/SetDisputeGameImpl.sol
```diff
@@ -0,0 +1,211 @@
...
+                bool vmChanged = address(prevFdg.vm()) != address(newFdg.vm());
+
+                // All other fields must match
+                bool othersMatch = prevFdg.maxGameDepth() == newFdg.maxGameDepth()
+                    && prevFdg.splitDepth() == newFdg.splitDepth()
+                    && prevFdg.maxClockDuration() == newFdg.maxClockDuration()
+                    && prevFdg.clockExtension() == newFdg.clockExtension();
```
Inphi:
```suggestion
                    && prevFdg.clockExtension() == newFdg.clockExtension()
                    && prevFdg.anchorStateRegistry() == newFdg.anchorStateRegistry();
```

src/improvements/template/SetDisputeGameImpl.sol
```diff
@@ -0,0 +1,211 @@
...
+
+                // Check if prestate or vm changed
+                bool prestateChanged = prevPdg.anchorStateRegistry() != newPdg.anchorStateRegistry();
+                bool vmChanged = address(prevPdg.vm()) != address(newPdg.vm());
+
+                // All other fields must match
+                bool othersMatch = prevPdg.maxGameDepth() == newPdg.maxGameDepth()
+                    && prevPdg.splitDepth() == newPdg.splitDepth()
+                    && prevPdg.maxClockDuration() == newPdg.maxClockDuration()
+                    && prevPdg.clockExtension() == newPdg.clockExtension() && prevPdg.proposer() == newPdg.proposer()
```
Inphi, line 172:
```suggestion
                    && prevPdg.clockExtension() == newPdg.clockExtension() && prevPdg.proposer() == newPdg.proposer()
                    && prevPdg.anchorStateRegistry () == newPdg.anchorStateRegistry()
```

src/improvements/template/SetDisputeGameImpl.sol
```diff
@@ -0,0 +1,211 @@
...
+
+            // -- PDG detailed check (only for impl->impl upgrade, not 0->impl or impl->0) --
+            address prevPdgAddr = c.prevPdgImpl;
+            address newPdgAddr = c.pdgImpl;
+            if (prevPdgAddr != address(0) && newPdgAddr != address(0) && prevPdgAddr != newPdgAddr) {
+                IPermissionedDisputeGame prevPdg = IPermissionedDisputeGame(prevPdgAddr);
+                IPermissionedDisputeGame newPdg = IPermissionedDisputeGame(newPdgAddr);
+
+                // Check if prestate or vm changed
+                bool prestateChanged = prevPdg.anchorStateRegistry() != newPdg.anchorStateRegistry();
```
Inphi:
```suggestion
                bool prestateChanged = prevPdg.absolutePrestate() != newPdg.absolutePrestate();
```

src/improvements/template/SetDisputeGameImpl.sol
Wazabie: @Inphi we actually need to use this template to reset the FDG implementation for Arena-Z and Swell, see [this](https://github.com/ethereum-optimism/superchain-ops/pull/1084#discussion_r2152726004) comment from @sbvegan

Inphi: That makes sense. I understand that the reset for Arena/Swell is a special case because those networks have a misconfigured cannon game implementation. If that's the case, could we add an if condition for their respective chainIDs to skip the non-zero enforcement?

src/improvements/template/SetDisputeGameImpl.sol
```diff
@@ -0,0 +1,234 @@
...
+            }
+
+            // Set PDG (PERMISSIONED_CANNON) implementation if TOML is different from current on-chain
+            address currentPDG = address(factory.gameImpls(PERMISSIONED_CANNON));
+            if (currentPDG != c.pdgImpl) {
+                factory.setImplementation(PERMISSIONED_CANNON, c.pdgImpl);
+            }
+
+            // Set FDG bond if not already set or needs update
+            if (fdgBond != 0 && factory.initBonds(CANNON) != fdgBond) {
```
Inphi: nit: since fdgBond is always 0.08 ether, might as well simplify to:
```suggestion
            if (c.fdgImpl != 0 && factory.initBonds(CANNON) != fdgBond) {
```

We also only want to set the initial bond for the specific game types that are affected.

src/improvements/template/SetDisputeGameImpl.sol
```diff
@@ -0,0 +1,234 @@
...
+                factory.setInitBond(CANNON, fdgBond);
+            }
+
+            // Set PDG bond if not already set or needs update
+            if (pdgBond != 0 && factory.initBonds(PERMISSIONED_CANNON) != pdgBond) {
```
Inphi: nit:
```suggestion
            if (c.pdgImpl != 0 && factory.initBonds(PERMISSIONED_CANNON) != pdgBond) {
```

src/improvements/template/SetDisputeGameImpl.sol
```diff
@@ -0,0 +1,234 @@
...
+        for (uint256 i = 0; i < configs.length; i++) {
+            cfg[configs[i].chainId] = configs[i];
+        }
+    }
+
+    /// @notice Write the calls that you want to execute for the task.
+    function _build() internal override {
+        // Default values for FDG and PDG bonds
+        uint256 fdgBond = 0.08 ether;
+        uint256 pdgBond = 0.08 ether;
```
Inphi:
```suggestion
        uint256 constant fdgBond = 0.08 ether;
        uint256 constant pdgBond = 0.08 ether;
```

Inphi (approved):

test/tasks/example/sep/014-set-dispute-game-impl/config.toml
```diff
@@ -0,0 +1,14 @@
+l2chains = [
+  { name = "Ink Sepolia", chainId = 763373 }
+]
+
+templateName = "SetDisputeGameImpl"
+
+[[gameImplsConfig]]
```
blmalone:
```suggestion
[[gameImplConfig]]
```

src/improvements/template/SetDisputeGameImpl.sol
```diff
@@ -0,0 +1,232 @@
...
+        storageWrites[0] = "DisputeGameFactoryProxy";
+        return storageWrites;
+    }
+
+    /// @notice Sets up the template with implementation configurations from a TOML file.
+    function _templateSetup(string memory taskConfigFilePath) internal override {
+        super._templateSetup(taskConfigFilePath);
+        string memory toml = vm.readFile(taskConfigFilePath);
+
+        GameImplConfig[] memory configs = abi.decode(toml.parseRaw(".gameImplsConfig"), (GameImplConfig[]));
```
blmalone:
```suggestion
        GameImplConfig[] memory configs = abi.decode(toml.parseRaw(".gameImplConfig"), (GameImplConfig[]));
```

blmalone (approved):

PR #1079, opened 2025-06-12
refactor(multisigtask): Introduce safe array

blmalone:
### TL;DR

Refactored the MultisigTask contract to set ourselves up for supporting multiple safes at different levels of nesting. This is one PR of many that will be refactoring MultisigTask.sol.

Base have [MultisigScript.sol](https://github.com/base/contracts/blob/main/script/universal/MultisigScript.sol#L218) which uses the same pattern. It's a clean approach to be able to handle many layers of safe nesting. 

### What changed?

- Modified `_taskSetup()` to accept an array of addresses instead of a single address
- Added `_appendOwnerSafe()` helper function to add the root/parent safe to the end of the safe array.
- Improved `_overrideState()` to work with multiple safes and track original nonces.
- Updated the signature of various methods to support the new multi-safe approach
- Consolidated state override application within the setup process
- Added TODOs for future improvements to fully support arbitrary nesting levels

PR #1085, opened 2025-06-16
refactor(MultisigTask): remove child state variables

blmalone:
## Context

This PR removes 3 state variables from `MultisigTask.sol`. To achieve this, the functions within `MultisigTask.sol` now pass around `allSafes`, `allCalldatas` and `allOriginalNonces`. This change permeates the entire code base and causes many tests to have to be updated. 

Some functions have been made internal that were once public (in `MultisigTask.sol`). This has made the diff a little tricker to read. 

This PR is part of a larger refactor plan to support arbitrary levels of nesting. You can see that this is part of a graphite stack. While going through this PR I (@blmalone) may say "_I've addressed this in a later PR <link-to-pr>_". 

### Testing
It's key that all tests still pass with these changes. I may be worth spending some time trying to understand it the changes to `*.t.sol` files are valid and not asserting on different conditions than they were before this PR.

### Note

Please pair with @blmalone on this PR. It's pretty large and the diff was somewhat unavoidable given how intertwined these state vars are in the current codebase.


blmalone:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/ethereum-optimism/superchain-ops/1085?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#1080** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/superchain-ops/1080?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1085** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/superchain-ops/1085?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/ethereum-optimism/superchain-ops/1085?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1079** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/superchain-ops/1079?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/improvements/tasks/MultisigTask.sol
```diff
@@ -675,75 +754,86 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager, TaskManage
...
 
         vm.label(AddressRegistry.unwrap(addrRegistry), "AddrRegistry");
         vm.label(address(this), "MultisigTask");
     }
 
+    /// @notice This function builds a series of nested transactions where each safe in the chain
+    /// must approve the transaction of the next safe, creating a left-to-right execution
+    /// dependency. The rightmost safe executes the actual actions, while all preceding
+    /// safes generate approval transactions for their successor.
+    function _calldatas(Action[] memory _actions, address[] memory _allSafes, uint256[] memory _originalNonces)
```
blmalone: this function was taken from base contracts: https://github.com/base/contracts/blob/main/script/universal/MultisigScript.sol#L335

src/improvements/tasks/MultisigTask.sol
```diff
@@ -349,17 +353,15 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager, TaskManage
...
+        _validate(accountAccesses, actions, rootSafe);
 
         checkStateDiff(accountAccesses);
     }
 
     /// @notice Get the multicall address for the given safe. Override to return required multicall address.
-    function _getMulticallAddress(address safe) internal view virtual returns (address) {
-        // Some child contracts may override this function and return a different multicall address
-        // based on the safe address (e.g. whether it's the parent or child multisig).
+    function _getMulticallAddress(address safe, address[] memory) internal view virtual returns (address) {
```
blmalone: takes `allSafes` as an arg so that child contracts can determine if `safe` is root or not.

src/improvements/tasks/MultisigTask.sol
```diff
@@ -408,6 +421,24 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager, TaskManage
         return (accountAccesses, txHash_);
     }
 
+    /// @notice Get the calldata to be executed by the root safe.
+    function getMulticall3Calldata(Action[] memory actions) internal view virtual returns (bytes memory data) {
```
blmalone: made this internal and moved it.

src/improvements/tasks/MultisigTask.sol
```diff
@@ -426,11 +457,38 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager, TaskManage
         }
     }
 
-    /// @notice Prank as the multisig.
-    /// Override to prank with delegatecall flag set to true
-    /// in case of OPCM tasks.
-    function _prankMultisig() internal virtual {
-        vm.startPrank(parentMultisig);
+    /// @notice Validate that the calldatas are in the correct order.
+    function validateCalldatas(bytes[] memory _datas, address[] memory _allSafes, uint256[] memory _originalNonces)
```
blmalone: this function is an important sanity check on all the generated calldatas.

src/improvements/tasks/MultisigTask.sol
```diff
@@ -463,51 +521,22 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager, TaskManage
...
         bool accessorNotRegistry = access.accessor != AddressRegistry.unwrap(addrRegistry);
         bool isCall = (access.kind == VmSafe.AccountAccessKind.Call && access.depth == topLevelDepth);
         bool isTopLevelDelegateCall =
             (access.kind == VmSafe.AccountAccessKind.DelegateCall && access.depth == topLevelDepth);
-        bool accessorIsParent = (access.accessor == parentMultisig);
-        return accountNotRegistryOrVm && accessorNotRegistry && (isCall || isTopLevelDelegateCall) && accessorIsParent;
-    }
-
-    /// @notice Composes a description string for the given access using the provided operation string.
-    function _composeDescription(VmSafe.AccountAccess memory access, string memory opStr)
```
blmalone, line 494: This function has been deleted. Not need for it to be a separate function. In a later PR this MultisigTask.sol file is refactored further.

src/improvements/tasks/MultisigTask.sol
```diff
@@ -620,28 +649,31 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager, TaskManage
     /// @notice To show the full transaction trace in Tenderly, we build custom calldata
     /// that shows both the child multisig approving the hash, as well as the parent multisig
     /// executing the task. This is only used when simulating a nested multisig.
-    function getNestedSimulationMulticall3Calldata(Action[] memory actions) internal view returns (bytes memory data) {
+    function getNestedSimulationMulticall3Calldata(address[] memory allSafes, bytes[] memory allCalldatas)
```
blmalone: This function is used for the tenderly simulation link only. All the changes are based purely on the fact that the args have changed.

src/improvements/tasks/MultisigTask.sol
```diff
@@ -682,95 +718,55 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager, TaskManage
...
+        IGnosisSafe _rootSafe; // TODO parentMultisig should be of type IGnosisSafe
+        (addrRegistry, _rootSafe, multicallTarget) = _configureTask(_taskConfigFilePath);
 
-        parentMultisig = address(_parentMultisig);
-        if (_childSafes.length > 0) childMultisig = _childSafes[0]; // TODO: remove this when childMultisig state variable is removed.
         // Appends the root safe. The earlier a safe address appears in the array, the deeper its level of nesting.
-        allSafes_ = Solarray.extend(_childSafes, Solarray.addresses(parentMultisig));
+        allSafes_ = Solarray.extend(_childSafes, Solarray.addresses(address(_rootSafe)));
         validateSafes(allSafes_);
+        root = address(_rootSafe);
```
blmalone, line 660: Keeping this state variable. May remove it in a later PR. Removing this now makes the diff even bigger.

src/improvements/tasks/MultisigTask.sol
```diff
@@ -829,19 +821,29 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager, TaskManage
...
-                string memory desc = _composeDescription(accesses[i], opStr);
+                (string memory opStr, Enum.Operation op) = GnosisSafeHashes.getOperationDetails(accesses[i].kind);
 
                 validActions[index] = Action({
                     value: accesses[i].value,
                     target: accesses[i].account,
                     arguments: accesses[i].data,
                     operation: op,
-                    description: desc
+                    description: string(
```
blmalone, line 753: this code is added in from the delete `_composeDescription` function above.

src/improvements/tasks/MultisigTask.sol
```diff
@@ -682,95 +718,55 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager, TaskManage
...
-            return IGnosisSafe(safe).nonce();
+            calldatas_[i - 1] = _generateApproveCalldata(targetSafe, callToApprove, 0, _originalNonces[i], _allSafes);
         }
     }
 
     /// ==================================================
     /// =============== Private Functions ================
     /// ==================================================
 
-    function _getOperationDetails(VmSafe.AccountAccessKind kind)
```
blmalone, line 758: Moved to `GnosisSafeHashes.sol`

src/improvements/tasks/MultisigTask.sol
```diff
@@ -879,106 +876,181 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager, TaskManage
...
         }
 
         MultisigTaskPrinter.printTenderlySimulationData(
-            targetAddress, finalExec, msg.sender, getStateOverrides(parentMultisig, childMultisig)
+            targetAddress, finalExec, msg.sender, getStateOverrides(rootSafe, childSafe)
         );
     }
 
     /// @notice Prints all relevant hashes to sign as well as the tenderly simulation link.
     function printSafe(
```
blmalone, line 859: This function is deleted in a later PR: https://github.com/ethereum-optimism/superchain-ops/pull/1090

src/libraries/Utils.sol
```diff
@@ -48,4 +48,16 @@ library Utils {
         }
         return false;
     }
+
+    /// @notice Helper function to get the safe, call data, and original nonce for a given index.
+    function getSafeData(
+        address[] memory allSafes,
+        bytes[] memory allCalldatas,
+        uint256[] memory allOriginalNonces,
+        uint256 index
```
alcueca: Just a thought, but if these three arrays are supposed to be aligned, shouldn't we instead have a {safe, calldata, originalNonce} struct, and an array of such structs, so that they don't get misaligned by error?

src/improvements/tasks/MultisigTask.sol
mds1: nit, since this is vendored from Base, we should preserve their function name + include a permalink to the commit this code was pulled from

src/libraries/Utils.sol
mds1: +1 was also going to suggest passing around an array of structs with this data

src/improvements/tasks/MultisigTask.sol
```diff
@@ -30,17 +30,13 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager, TaskManage
     using AccountAccessParser for VmSafe.AccountAccess;
     using StdStyle for string;
 
-    /// @notice Parent nonce used for generating the safe transaction.
-    uint256 public nonce;
+    /// @notice The root safe address for the task
+    address public root;
```
mds1: will users (i.e. task developers) ever need to use this variable? A more clear name might be useful
```suggestion
    address public rootSafe;
```

blmalone: I kept this variable in because the diff gets even larger when I remove it and it's trivial to remove. I can do it in a later PR. Agreed with the name change for now though many variable may shadow this one if I change the name.

src/improvements/tasks/MultisigTask.sol
blmalone: done: https://github.com/ethereum-optimism/superchain-ops/pull/1085/commits/ae0fb79cccaac8c586a8cbff53bbd31e69e3b1d5#diff-0fc3bbbfb7c338051eba7e4d380f646df613586e3973f5a605734ab254ccf1a1R740

src/libraries/Utils.sol
blmalone: good call, done: https://github.com/ethereum-optimism/superchain-ops/pull/1085/commits/ae0fb79cccaac8c586a8cbff53bbd31e69e3b1d5#diff-0fc3bbbfb7c338051eba7e4d380f646df613586e3973f5a605734ab254ccf1a1R740

src/improvements/tasks/MultisigTask.sol
alcueca: Is the root safe the same as the parent multisig?

src/improvements/tasks/MultisigTask.sol
```diff
@@ -73,30 +63,31 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager, TaskManage
...
+        address rootSafe = payload.safes[payload.safes.length - 1];
+        uint256 rootSafeNonce = payload.originalNonces[payload.originalNonces.length - 1];
 
-        (VmSafe.AccountAccess[] memory accountAccesses, bytes32 txHash) = execute(signatures, actions);
+        (VmSafe.AccountAccess[] memory accountAccesses, bytes32 txHash) = execute(signatures, payload);
 
-        // validate all state transitions
-        validate(accountAccesses, actions);
-        print(actions, accountAccesses, false, txHash);
+        validate(accountAccesses, actions, rootSafe, rootSafeNonce);
```
alcueca: We should pass the `payload` to `validate`, and let `validate` decide what it is that it needs from it.

src/improvements/tasks/MultisigTask.sol
```diff
@@ -124,49 +114,53 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager, TaskManage
...
     }
 
     /// @notice Child multisig approves the task to be executed.
-    function approve(address _childMultisig, bytes memory signatures, Action[] memory actions) public {
-        bytes memory approveCalldata = generateApproveMulticallData(actions);
-        bytes32 hash = keccak256(getEncodedTransactionData(_childMultisig, approveCalldata));
-        signatures = Signatures.prepareSignatures(_childMultisig, hash, signatures);
-
-        execTransaction(_childMultisig, MULTICALL3_ADDRESS, 0, approveCalldata, Enum.Operation.DelegateCall, signatures);
+    function approve(bytes memory signatures, TaskPayload memory payload) public {
```
alcueca: There is nothing obviously related to approving a transaction in this function anymore. Maybe there is some generalization that can be done.

src/improvements/tasks/MultisigTask.sol
```diff
@@ -682,95 +696,43 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager, TaskManage
...
         internal
-        returns (address[] memory allSafes_)
+        returns (TaskPayload memory payload_, Action[] memory actions_)
     {
-        require(parentMultisig == address(0), "MultisigTask: already initialized");
+        require(root == address(0), "MultisigTask: already initialized");
         templateConfig.safeAddressString = loadSafeAddressString(MultisigTask(address(this)), _taskConfigFilePath);
-        IGnosisSafe _parentMultisig; // TODO parentMultisig should be of type IGnosisSafe
-        (addrRegistry, _parentMultisig, multicallTarget) = _configureTask(_taskConfigFilePath);
+        IGnosisSafe _rootSafe; // TODO parentMultisig should be of type IGnosisSafe
```
alcueca: Outdated comment

src/improvements/tasks/MultisigTask.sol
alcueca: +1 on refactoring `root` out in a later PR, though.

blmalone: Yes, rootSafe is the same as parentMultisig. Trying to move away from parent, child terminology.

src/improvements/tasks/MultisigTask.sol
blmalone: In the latest version I've refactored a bunch of code. The entrypoint functions are further consolidated so this should be much clearer.

src/improvements/tasks/MultisigTask.sol
blmalone: fixed: https://github.com/ethereum-optimism/superchain-ops/pull/1085/commits/c2f6aac9136b5741524cf1dd5d102f2618a9bcb1

src/improvements/tasks/MultisigTask.sol
blmalone: Fixed here: https://github.com/ethereum-optimism/superchain-ops/pull/1085/commits/c2f6aac9136b5741524cf1dd5d102f2618a9bcb1 and here: https://github.com/ethereum-optimism/superchain-ops/pull/1085/commits/360612459835f7fef28049bb89512f00a2050ca0

src/improvements/tasks/MultisigTask.sol
blmalone: Changed here: https://github.com/ethereum-optimism/superchain-ops/pull/1085/commits/3bb6b18f103c9a7a3c9d5be68f94400bb223183c

src/improvements/tasks/MultisigTask.sol
```diff
@@ -58,14 +58,28 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager, TaskManage
...
+    /// @notice Simulates the nested safe transaction of the task.
+    /// This works by printing the 'data to sign' for the nested safe which is then passed to the eip712sign binary for signing.
+    function simulate(string memory taskConfigFilePath, address _childMultisig)
+        public
+        returns (VmSafe.AccountAccess[] memory, Action[] memory, bytes32, bytes memory)
+    {
+        return _runTask(taskConfigFilePath, "", _childMultisig, true);
+    }
+
+    /// @notice Simulates the nested safe transaction of the task without logging.
```
alcueca: That we need to supress the return values from `simulate` if we want to extract 'data to sign' cleanly is quite obscure and should be documented better.

Maybe this function should be called `printDataToSign`, or `simulateAndPrintDataToSign`. I prefer the former, since it is not that relevant to the caller that the transaction will be simulated.

src/improvements/tasks/MultisigTask.sol
alcueca: Maybe just leave a `// TODO:` so that it is not completely forgotten

src/improvements/tasks/MultisigTask.sol
```diff
@@ -68,105 +58,108 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager, TaskManage
...
-        bytes32 hash = keccak256(getEncodedTransactionData(_childMultisig, approveCalldata));
-        signatures = Signatures.prepareSignatures(_childMultisig, hash, signatures);
-
-        execTransaction(_childMultisig, MULTICALL3_ADDRESS, 0, approveCalldata, Enum.Operation.DelegateCall, signatures);
-    }
-
-    /// @notice Executes the task with the given signatures.
-    function execute(bytes memory signatures, Action[] memory actions)
+    /// @notice Executes a task for a given safe. The 'safeIndex' is the index of the safe in the payload.safes array.
+    function execute(bytes memory signatures, TaskPayload memory payload, uint256 safeIndex)
```
alcueca: It is quite confusing that the execution path for `execute` in nested.just is:
```
execute(taskConfigFilePath, signatures) public
    _runTask(...) internal
        execute(signatures, payload, safeIndex) public
```

Maybe they should be renamed as:
```
executeFromPath(taskConfigFilePath, signatures) public
executeFromPayload(payload, signatures, safeIndex) public
```
Note I moved arguments around, so that they are first what ot execute, then the signatures, then modifier arguments.

src/improvements/tasks/MultisigTask.sol
```diff
@@ -208,15 +201,15 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager, TaskManage
 
     /// @notice Build the task actions for all l2chains in the task.
     /// Contract calls must be performed in plain solidity.
-    function build() public returns (Action[] memory actions) {
-        require(parentMultisig != address(0), "Must set address registry for multisig address to be set");
-
+    function build(address rootSafe) public returns (Action[] memory actions) {
```
alcueca, line 202: I find it weird that you are passing `rootSafe` as an argument when you have `root` as a state variable with the same value. There is a chance there for both value to diverge without being immediately noticed.

I would suggest that you introduce this argument in the same PR as you remove `root` as a state variable.

src/improvements/tasks/MultisigTask.sol
```diff
@@ -237,43 +230,35 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager, TaskManage
...
-    function getEncodedTransactionData(address safe, bytes memory data)
-        public
-        view
-        returns (bytes memory encodedTxData)
-    {
+    function getEncodedTransactionData(
+        address safe,
+        bytes memory data,
+        uint256 value,
+        uint256 originalNonce,
```
alcueca, line 246: Same as below. You could just call it `nonce` now.

src/libraries/MultisigTypes.sol
```diff
@@ -67,3 +67,11 @@ struct TaskConfig {
     bool isNested;
     address task; // MultisigTask address
 }
+
+/// @notice This struct contains all the data needed to execute a task.
+/// All safes involved in the task must be represented in this struct.
+struct TaskPayload {
+    address[] safes;
+    bytes[] calldatas;
+    uint256[] originalNonces;
```
alcueca, line 76: Why is this `originalNonces`, and not just `nonces`?

src/improvements/tasks/MultisigTask.sol
```diff
@@ -237,43 +230,35 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager, TaskManage
...
-        });
-
-        IMulticall3.Call3Value[] memory calls = new IMulticall3.Call3Value[](1);
-        calls[0] = call;
-        return abi.encodeWithSignature("aggregate3Value((address,bool,uint256,bytes)[])", calls);
+    function getHash(
+        bytes memory callData,
+        address safe,
+        uint256 value,
+        uint256 originalNonce,
```
alcueca, line 235: Why is this called originalNonce, and not just nonce?

alcueca: I reckon it is because before the nonce to encode with was a state variable, so it needed to be stored with more context.

src/improvements/tasks/MultisigTask.sol
```diff
@@ -853,132 +773,191 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager, TaskManage
...
+            finalExec = _getNestedSimulationMulticall3Calldata(allSafes, allCalldatas);
+            childSafe = allSafes[0];
         } else {
-            targetAddress = parentMultisig;
+            targetAddress = rootSafe;
             finalExec = _execTransactionCalldata(
-                parentMultisig,
-                getMulticall3Calldata(actions),
+                targetAddress,
+                allCalldatas[allCalldatas.length - 1],
```
alcueca: Wouldn't this be `allCalldatas[0]`?

src/improvements/tasks/MultisigTask.sol
```diff
@@ -853,132 +773,191 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager, TaskManage
...
+        address rootSafe = payload.safes[payload.safes.length - 1];
+        accountAccesses.decodeAndPrint(rootSafe, txHash);
 
-        return printSafe(actions, accountAccesses, isSimulate, txHash);
+        return printSafe(accountAccesses, isSimulate, txHash, payload);
     }
 
     /// @notice Print the Tenderly simulation payload with the state overrides.
-    function printTenderlySimulationData(Action[] memory actions) internal view {
+    function _printTenderlySimulationData(address[] memory allSafes, bytes[] memory allCalldatas) internal view {
```
alcueca: I'm leaning a bit towards passing `payload` to this function, instead of disaggregating it, and the same with `_getNestedSimulationMulticall3Calldata`, so that there is no chances of future implementations having mismatching safe and calldata arrays.

src/improvements/tasks/MultisigTask.sol
alcueca: What if `allCalldatas.length` is zero?

test/tasks/SingleMultisigTask.t.sol
```diff
@@ -246,17 +253,25 @@ contract SingleMultisigTaskTest is Test {
...
-        bytes memory dataToSign = multisigTask.getEncodedTransactionData(multisigTask.parentMultisig(), callData);
-        address multisig = multisigTask.parentMultisig();
+
+        address[] memory allSafes = MultisigTaskTestHelper.getAllSafes(multisigTask.root());
+        uint256[] memory allOriginalNonces = MultisigTaskTestHelper.getAllOriginalNonces(allSafes);
+        bytes[] memory allCalldatas = multisigTask.transactionDatas(actions, allSafes, allOriginalNonces);
+
+        address rootSafe = allSafes[allSafes.length - 1];
+        bytes memory rootSafeCalldata = allCalldatas[allCalldatas.length - 1];
+        uint256 rootSafeNonce = allOriginalNonces[allOriginalNonces.length - 1] - 1; // The task has already run so we decrement the nonce by 1.
```
alcueca: I think that you could do with a `_getRootSafeData() internal returns(address rootSafe, uint256 rootSafeNonce, bytes rootSafeCalldata)` function.

test/tasks/SingleMultisigTask.t.sol
```diff
@@ -246,17 +253,25 @@ contract SingleMultisigTaskTest is Test {
...
         multisigTask.processTaskActions(actions);
-        bytes memory callData = multisigTask.getMulticall3Calldata(actions);
-        bytes memory dataToSign = multisigTask.getEncodedTransactionData(multisigTask.parentMultisig(), callData);
-        address multisig = multisigTask.parentMultisig();
+
+        address[] memory allSafes = MultisigTaskTestHelper.getAllSafes(multisigTask.root());
+        uint256[] memory allOriginalNonces = MultisigTaskTestHelper.getAllOriginalNonces(allSafes);
+        bytes[] memory allCalldatas = multisigTask.transactionDatas(actions, allSafes, allOriginalNonces);
+
+        address rootSafe = allSafes[allSafes.length - 1];
```
alcueca: Shouldn't this be also `multisigTask.root()`?

src/improvements/tasks/MultisigTask.sol
alcueca: Looking deeper into this, only `approve` and `_runTask` call `execute(signatures, payload, safeIndex)`.

Maybe `_runTask` can be merged with `execute(signatures, payload, safeIndex)` and have all the calls from the just files end in the same resulting function.

alcueca:
All considered, I see no problem merging this. All my concerns where about style, and I'm conscious that this is one of many refactoring PRs that need to be executed in sequence, so not everything needs to be fixed in this one. Once the comments are addressed and CI passes I'll be happy to approve.

test/tasks/SingleMultisigTask.t.sol
blmalone: It's the same address. The last element in allSafes will always be the root. We're going to remove the root state var in a later PR anyway.

src/libraries/MultisigTypes.sol
blmalone: I think `originalNonces` adds a little more information - they're the nonces of each of the safes _before_ the simulation. Base contracts use the same term: https://github.com/base/contracts/blob/main/script/universal/MultisigScript.sol#L224

src/improvements/tasks/MultisigTask.sol
blmalone: Yea, I was originally going to remove the root state variable in this PR then decided it was going to have a larger blast radius. 

I carefully made the change to pass`rootSafe` as a function param to the templates. This changes 3 functions:
1. `_templateSetup`
2. `_build`
3. `_validate`

All of which are being given the correct root safe address via the `MultisigTask.sol` file.

I'm confident this change is benign and so removing it now to just re-add it back in is adding more work. For that reason I'll keep it in unless there's there's a pressing issue with it.

src/improvements/tasks/MultisigTask.sol
blmalone: Same as the other comment around bases naming convention.

src/improvements/tasks/MultisigTask.sol
blmalone: Line 837 is `if (allSafes.length > 1) {`, if that's > 1 then so too will `allCalldatas`. I agree with your comment below, we should pass payload here to make it more obvious. Length validations were performed on the payload so we should operate on it instead of individual arrays.

src/improvements/tasks/MultisigTask.sol
blmalone: Fixed: https://github.com/ethereum-optimism/superchain-ops/pull/1085/commits/f60fe1e8349abf0dadd0b96b0a441985c6887efe

src/improvements/tasks/MultisigTask.sol
blmalone: Made clearer here: https://github.com/ethereum-optimism/superchain-ops/pull/1085/commits/f60fe1e8349abf0dadd0b96b0a441985c6887efe

test/tasks/SingleMultisigTask.t.sol
blmalone: I have something like that already, I've made this commit to use it: https://github.com/ethereum-optimism/superchain-ops/pull/1085/commits/6ac78866aa91c98ba83f9e3cb2504771c836721e

alcueca:
I cloned the repo again and the tests pass, let's merge.

alcueca (dismissed):

mds1 (approved):

PR #1092, opened 2025-06-18
refactor(GnosisSafeHashes): moving functions into GnosisSafeHahses.

blmalone:
### TL;DR

Refactored Gnosis Safe transaction encoding logic by moving functions from `MultisigTask` to the `GnosisSafeHashes` library.

### What changed?

- Moved `getEncodedTransactionData` and `_execTransactionCalldata` from `MultisigTask` to `GnosisSafeHashes` library, renaming the latter to `encodeExecTransactionCalldata`
- Updated all references to these functions throughout the codebase
- Renamed some private functions in `MultisigTask` to follow the underscore prefix convention (e.g., `printNestedData` → `_printNestedData`)
- Added comprehensive tests for the newly moved functions in `GnosisSafeHashes.t.sol`
- Fixed parameter naming in the library to follow consistent conventions


src/libraries/GnosisSafeHashes.sol
```diff
@@ -137,8 +137,8 @@ library GnosisSafeHashes {
     }
 
     // Helper function to get minimum of two values
-    function min(uint256 a, uint256 b) internal pure returns (uint256) {
-        return a < b ? a : b;
+    function min(uint256 _a, uint256 _b) internal pure returns (uint256) {
+        return _a < _b ? _a : _b;
```
alcueca, line 141: Why this change? It looked fine to me.

alcueca (approved):
Mostly moving two functions to the GnosisSafeHashes library

src/libraries/GnosisSafeHashes.sol
blmalone: We have a lose styling guide that says we should prefix params of internal/private functions with `_` ([here](https://github.com/ethereum-optimism/optimism/blob/ab50df748916e058d2d5c203549ac61eeaec4953/packages/contracts-bedrock/book/src/contributing/style-guide.md?plain=1#L65)).

PR #1130, opened 2025-07-15
feat: removing CI config for old rehearsals

blmalone:


security-council-rehearsals/justfile
```diff
@@ -1,13 +0,0 @@
-set dotenv-load
```
Ethnical, line 1: I assume this is intended here and we don't want also to archive this justfile in the `archive` folder right?

.circleci/config.yml
```diff
@@ -95,67 +95,6 @@ jobs:
...
-            cd security-council-rehearsals
-            just setup r2-remove-signer tmp-ci
-            cd *r2-remove-signer-tmp-ci
-            export SIMULATE_WITHOUT_LEDGER=1
-            just simulate
-            just get-owners
-            just prepare
-            just simulate # simulate again to make sure the json is still valid
-
-  just_simulate_sc_rehearsal_4:
```
Ethnical, line 139: 1. Interesting that here we gap the `sc_rehearsal_3` seems to not be present in the file. 
2. I see only the removal of the `just_simulate_sc_rehearsal_*` I wonder now how we will simulate this not in the CI or there is another PR for introducing the simulation again with the new superchain-opsv2? Is this going to use the `template-regression_tests` located here -> https://github.com/ethereum-optimism/superchain-ops/blob/33c0e991df7615caf3420a47514cd9df8414b59e/.circleci/config.yml#L213

Ethnical (approved):
Nits: I left few comments that if you can answer just before merging would be nice :) 
Otherwise looking to me!

.circleci/config.yml
blmalone: yes, the templates are automatically simulated using `template_regression_tests`. The amount of code used for these rehearsals has decreased, we shouldn't need to simulated them individually like this anymore.

security-council-rehearsals/justfile
blmalone: Yea I figured I'd just keep the records of the rehearsals completed and delete the source code. We not longer need it.

Ethnical: Got it thank you!

Ethnical (approved):

PR #1132, opened 2025-07-15
feat: Updating Security Council Rehearsals (step 3 - nested upgrade) 

blmalone:
### TL;DR

Refactored the deployment script for nested rehearsals and added comprehensive documentation for the third security council rehearsal step.

### What changed?

- Renamed `DeployRehearsal4.s.sol` to `DeployNestedRehearsal.sol` and simplified it by:
  - Removing the L1ERC721Bridge deployment
  - Removing dependency on Bedrock's Deployer
  - Focusing only on deploying ProxyAdmin and OptimismPortalProxy
  - Improving code organization and readability

- Added a new `step-3-nested-upgrade` directory with:
  - Detailed README with step-by-step instructions
  - Environment configuration file
  - Justfile for deployment
  - Supporting images for documentation

- Updated references in security council rehearsal documentation to use `eth` directory instead of `sep`

- Added a new `SetEIP1967Implementation.sol` template for proxy implementation upgrades

- Improved the MultisigTask architecture by:
  - Separating storage access setup into dedicated methods
  - Making validation methods non-view to allow state modifications

### How to test?

Follow the instructions in the new `security-council-rehearsals/step-3-nested-upgrade/README.md` as both the facilitator and signer.


src/improvements/tasks/MultisigTask.sol
```diff
@@ -901,6 +905,12 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager, TaskManage
     /// State overrides are not applied yet. Keep this in mind when performing various pre-simulation assertions in this function.
     function _templateSetup(string memory taskConfigFilePath, address rootSafe) internal virtual;
 
+    /// @notice Sets the allowed storage accesses.
```
blmalone, line 908: These were introduced ultimately to support being able to set a variable in _templateSetup and then use that variable in `_taskStorageWrites()`. See the new ordering of the functions above and the `SetEIP1967Impl` template for an example of this in action.

security-council-rehearsals/step-1-welcome/README.md
```diff
@@ -204,18 +204,18 @@ using keys owned by the Facilitator, to make it a M-of-2M+1 multisig.
 For example, if there are 3 participants, we recommend creating a 3 of 7
 multisig with 4 additional keys owned by the Facilitator.
 
-#### 2. Create a new task in the `sep` directory:
```
Ethnical, line 207: Curious here, if there is still the instruction for `sep` since we replace all them with `eth`

blmalone: The rehearsals aren't meant to occur on sepolia. That's why I've changed the instructions to only reference mainnet. It's trivial to try it out on sepolia though.

security-council-rehearsals/step-3-nested-upgrade/README.md
```diff
@@ -0,0 +1,286 @@
...
+
+## [For Facilitator ONLY] How to prepare and execute the rehearsal
+
+### [Before the rehearsal] Prepare the rehearsal
+
+#### 1. Create and configure the safes
+
+To simulate our production Safe configuration (i.e. 2-of-2 between the Security Council and the Optimism Foundation), we need to create 2 more Safe contracts. We are assuming you have already created the Security Council Safe in the [previous rehearsal](../step-1-welcome/README.md). You should create the safes in the following order:
+1. Fake Foundation Safe: You can set the owner on this safe to your EOA and the threshold to 1.
+2. Owner Safe: This safe will have two owners (fake security council and the fake foundation safe). It's important they're both Safe contracts. The simulation will fail if they're not.
```
alcueca:
```suggestion
2. Owner Safe: This safe will have two owners (fake security council and the fake foundation safe). It's important they're both Safe contracts. The simulation will fail if they're not. Set the threshold to 2.
```

src/improvements/justfile
```diff
@@ -115,6 +115,10 @@ simulate-all-templates:
     root_dir=$(git rev-parse --show-toplevel)
     forge build
     simulation_count=0
+    # When we spawn too many concurrent processes, the simulation fails with 'solc exited with signal: 9 (SIGKILL)'.
+    # The batch size should be low enough to not cause issues, but high enough to not make the simulation too slow.
+    batch_size=13
```
Ethnical, line 120: Interesting, to know how did you conclude that 13 will be the correct number to avoid the SIGKILL here.

Ethnical (approved):
SGTM, I have check mostly the file that are used in Prod. 
Since the rest would be used during drills (for this part we should be able to caught bug when we will run the code with the SC leader).

src/improvements/justfile
blmalone: trial and error

PR #1137, opened 2025-07-18
Ethni/new unpause

Ethnical:
**Description**

**Tests**

**Additional context**

**Metadata**


src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
```diff
@@ -0,0 +1,94 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import {VmSafe} from "forge-std/Vm.sol";
+import {stdToml} from "forge-std/StdToml.sol";
+
+import {L2TaskBase} from "src/improvements/tasks/types/L2TaskBase.sol";
+import {SuperchainAddressRegistry} from "src/improvements/SuperchainAddressRegistry.sol";
```
blmalone: import not used

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
```diff
@@ -0,0 +1,94 @@
...
+
+import {IOptimismPortal2} from "lib/optimism/packages/contracts-bedrock/interfaces/safe/IDeputyGuardianModule.sol";
```
blmalone: import not used.

src/improvements/template/UnPauseSuperchainConfig.sol
```diff
@@ -1,64 +0,0 @@
-// SPDX-License-Identifier: MIT
```
blmalone: Does this need to be delete? Might this not be used for chains pre U16? Also does this have a test task that exists that'll fail now it's deleted?

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
```diff
@@ -0,0 +1,94 @@
...
+
+    function unpause(address _identifier) external;
+
+    function paused() external view returns (bool);
+
+    function paused(address _identifier) external view returns (bool);
+}
+
+/// @title UnPauseSuperchainConfigTemplateAfterU16 template.
+contract UnPauseSuperchainConfigTemplateAfterU16 is L2TaskBase {
```
blmalone: Can you come up with a better name for this template? Here's some of the tips from the cli:

```
  echo -e "Enter template file name (e.g.\033[33m <template_name>.sol\033[0m)."
  echo "    - Make the name generic enough so that other developers know it is reusable."
  echo "    - Follow single responsibility: one task per template (see: TransferL1PAO and TransferL2PAO)."
  echo "    - Ideally, the name should start with a verb like 'Transfer', 'Update', 'Set'."
  echo "    - Avoid using 'Template' in the name."
  echo -e "\033[33mTip: Name with future reuse in mind.\033[0m"
```

Perhaps something like UnpauseSuperchainConfigVXXX? Where VXXX is the op-contracts version that the chain is on or something?

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
```diff
@@ -0,0 +1,94 @@
...
+
+import {L2TaskBase} from "src/improvements/tasks/types/L2TaskBase.sol";
+import {SuperchainAddressRegistry} from "src/improvements/SuperchainAddressRegistry.sol";
+
+import {IOptimismPortal2} from "lib/optimism/packages/contracts-bedrock/interfaces/safe/IDeputyGuardianModule.sol";
+
+import {Action} from "src/libraries/MultisigTypes.sol";
+
+interface IETHLockbox {
+    // function systemConfig() external view returns (ISystemConfig);
```
blmalone: delete?

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
```diff
@@ -0,0 +1,94 @@
...
+    /// @notice Returns the string identifier for the safe executing this transaction.
+    function safeAddressString() public pure override returns (string memory) {
+        return "SecurityCouncil";
+    }
+
+    /// @notice Returns string identifiers for addresses that are expected to have their storage written to.
+    function _taskStorageWrites() internal pure override returns (string[] memory) {
+        string[] memory storageWrites = new string[](2);
+        storageWrites[0] = "SuperchainConfig";
+        storageWrites[1] = safeAddressString();
```
blmalone: Is this because of the LivenessGuard?

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
```diff
@@ -0,0 +1,94 @@
...
+
+    function paused(address _identifier) external view returns (bool);
+}
+
+/// @title UnPauseSuperchainConfigTemplateAfterU16 template.
+contract UnPauseSuperchainConfigTemplateAfterU16 is L2TaskBase {
+    using stdToml for string;
+
+    /// @notice Identifier of the eth_lockbox used in the SuperchainConfig loaded from TOML
+    address public identifier;
```
blmalone: can you give an example of what this identifier would be?

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
```diff
@@ -0,0 +1,94 @@
...
+import {stdToml} from "forge-std/StdToml.sol";
+
+import {L2TaskBase} from "src/improvements/tasks/types/L2TaskBase.sol";
+import {SuperchainAddressRegistry} from "src/improvements/SuperchainAddressRegistry.sol";
+
+import {IOptimismPortal2} from "lib/optimism/packages/contracts-bedrock/interfaces/safe/IDeputyGuardianModule.sol";
+
+import {Action} from "src/libraries/MultisigTypes.sol";
+
+interface IETHLockbox {
```
blmalone: moving the interfaces down to the bottom of the file is cleaner imo.

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
```diff
@@ -0,0 +1,94 @@
...
+        storageWrites[1] = safeAddressString();
+        return storageWrites;
+    }
+
+    /// @notice Sets up the template with implementation configurations from a TOML file.
+    function _templateSetup(string memory _taskConfigFilePath, address _rootSafe) internal override {
+        super._templateSetup(_taskConfigFilePath, _rootSafe);
+        string memory file = vm.readFile(_taskConfigFilePath);
+        identifier = vm.parseTomlAddress(file, ".identifier"); // Get the identifier of the eth_lockbox from the TOML file
+        sc = ISuperchainConfig(superchainAddrRegistry.get("SuperchainConfig"));
```
blmalone: should you be checking if the chain is paused before continuing? Are there anymore checks you've want to perform? e.g. is it a valid identifier?

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    /// @notice Write the calls that you want to execute for the task.
+    function _build(address) internal override {
+        // 1. Load the SuperchainConfig contract.
+        sc = ISuperchainConfig(superchainAddrRegistry.get("SuperchainConfig"));
```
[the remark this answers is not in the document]
blmalone: like graphite says, this is redundant code.

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
Ethnical: solve in commit: b5581ba3e1f65cafaed871a94733e15975b70c44

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
Ethnical: solve in commit: b5581ba3e1f65cafaed871a94733e15975b70c44

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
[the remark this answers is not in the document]
Ethnical: solved in commit: 31755cb699116806c5d3fb61789d9633f0bcd74a

src/improvements/template/UnPauseSuperchainConfig.sol
Ethnical: I don't understand this comment if you can elaborate more there?

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
Ethnical: solved in commit: d7e4822de2dc7a2b6104f54b4499a455ab0f79b6

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
Ethnical: Yeah, but I realized that we don't need to put the safe somehow, in the latest commit there is no safe but only the guardian this is likely removal I will do the test and see

src/improvements/template/UnPauseSuperchainConfig.sol
blmalone: The `UnPauseSuperchainConfig.sol` template is being deleted in this PR, why?

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
Ethnical: yeah fixed here 259b39c8a613524a4fa7c5e4d7230e468c06f2db

blmalone: ok, I'd imagine if you're using the security council mainnet address in this task you'll need the Livenessguard in there like this: https://github.com/ethereum-optimism/superchain-ops/blob/main/src/improvements/template/GnosisSafeRemoveOwner.sol#L45

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
Ethnical: I did here in this commit f39e83b22d987b8ef48a0a162ea13dd4308aa428

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
Ethnical: Got moved at the bottom!
solved in 7a272c9e17e9fd590d254a14bf98eca178b4fb19

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
Ethnical: You think I should do add still the SC into? 
For me that make sense but somehow here it seems to be by default?

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
Ethnical: I did in previous commit

src/improvements/template/UnPauseSuperchainConfig.sol
```diff
@@ -1,64 +0,0 @@
...
-
-    /// @notice Write the calls that you want to execute for the task.
-    function _build(address) internal override {
-        // Load the DeputyGuardianModule contract.
-        IDeputyGuardianModule dgm = IDeputyGuardianModule(superchainAddrRegistry.get("DeputyGuardianModule"));
-        dgm.unpause(); // Unpause the SuperchainConfig contract through the DeputyGuardianModule.
-    }
-
-    /// @notice This method performs all validations and assertions that verify the calls executed as expected.
-    function _validate(VmSafe.AccountAccess[] memory, Action[] memory, address) internal view override {
```
Ethnical: @blmalone this template is using old code I don't even know this compiling in the past here, like see together with `_validate`

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
blmalone: Not sure I'm following what you mean. I'm saying, if this is simulated with the mainnet security council address, you'll likely get an error. Your test task is in the sepolia directory right now.

src/improvements/template/UnPauseSuperchainConfig.sol
Ethnical: The idea is to remove the previous one since this would be deprecated once we perform the U16 upgrade.

blmalone: Okay so there's no chance we will ever use it again? If that's the case then it makes sense to remove it.

src/improvements/template/UnPauseSuperchainConfig.sol
blmalone: this function signature for `_validate` isn't old code, it's correct?

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
blmalone: I don't see the change

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
Ethnical: Oh that's precisely my point! 
this is the liveness module on sep for the SC SEP-> https://sepolia.etherscan.io/address/0xEB3eF34ACF1a6C1630807495bCC07ED3e7B0177e#code
this is the livenessguard on  sep  -> https://sepolia.etherscan.io/address/0xc26977310bC89DAee5823C2e2a73195E85382cC7#readContract

I wonder how this is possible that the storage access is not reverting since I didn't add the the livenessModule logic into the storage access. 
This should even revert on sepolia no?

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
```diff
@@ -0,0 +1,88 @@
...
+        string[] memory storageWrites = new string[](1);
+        storageWrites[0] = "SuperchainConfig";
+        return storageWrites;
+    }
+
+    /// @notice Sets up the template with implementation configurations from a TOML file.
+    function _templateSetup(string memory _taskConfigFilePath, address _rootSafe) internal override {
+        super._templateSetup(_taskConfigFilePath, _rootSafe);
+        string memory file = vm.readFile(_taskConfigFilePath);
+        identifier = vm.parseTomlAddress(file, ".identifier"); // Get the identifier of the eth_lockbox from the TOML file
```
blmalone: is there ever a case where you would want one task to unpause >1 identifier? If so, should this be an array?

Ethnical: I don't think we want to do that either we used the superchain-wide-unpause or either you choose one interop-set to unpause specific.

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
Ethnical: The contract is not called with `template` anymore see this line https://github.com/ethereum-optimism/superchain-ops/blob/7a272c9e17e9fd590d254a14bf98eca178b4fb19/src/improvements/template/UnPauseSuperchainConfigAfterU16.sol#L12

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
blmalone: Ah, never mind, the simualtion executes on the Guardian not the SecurityCouncil. So you won't see the LivenessGuard state changes locally, only in Tenderly.

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
blmalone: I see, maybe a nit but the `AfterU16` part could be seen as ambiguous. U16 is just a name we're giving to this interop upgrade but we might have another U16 in the future. Can you be more specific an add the version like I mentioned above?

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
blmalone: Resolving.

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
blmalone: sounds good, resolving.

src/improvements/template/UnPauseSuperchainConfig.sol
Ethnical: Oh you are right, I was likely on the incorrect file. My bad for this you can ignore this comment haha starting to be tired!

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
Ethnical: Oooh okay! I will fix this Monday! Thank you didn't see it haha

src/improvements/template/UnPauseSuperchainConfig.sol
JosepBove: I guess the theoretically there is a possibility to have a chain thats still on an older version right?

Ethnical: >I guess the theoretically there is a possibility to have a chain thats still on an older version right?

I don't see how since we will upgrade our SuperchainConfig contract on L1. 
This can still occurs with chains that we didn't key-handover.  

I guess this would more confusing to keep it with a name similar if we wants to keep it we should likely rename it but I almost sure this template will never serve (we can keep it at first, and remove it later in the future anyway if never used).

JosepBove: Yeah my worries is for people that do not upgrade (yet)

JosepBove: Seeing it's not a trouble to keep it here I keep it, we can always remove it on the future

Ethnical: Kept both addressed in this commit: 3b25ce72c6ec6cbb9f3ca98cbe7f315b4269ea0a

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
Ethnical: Addressed in this commit: 49b770664acb7dcfe6df54893832a6f8faa9c3ae

src/improvements/template/UnPauseSuperchainConfigv400.sol
```diff
@@ -0,0 +1,88 @@
+// SPDX-License-Identifier: MIT
```
blmalone: nit: file name should be `UnpauseSuperchainConfigV400.sol` (lowercase 'p' and upper case V as used in the OPCM templates).

src/improvements/template/UnPauseSuperchainConfig.sol
```diff
@@ -13,7 +13,7 @@ import {L2TaskBase} from "src/improvements/tasks/types/L2TaskBase.sol";
 import {SuperchainAddressRegistry} from "src/improvements/SuperchainAddressRegistry.sol";
 import {Action} from "src/libraries/MultisigTypes.sol";
 
-/// @title UnPauseSuperchainConfig
+/// @title UnPauseSuperchainConfig before contract version 4.0.0. After version 4.0.0 please refer to the template UnpauseSuperchainConfigv400.
 contract UnPauseSuperchainConfig is L2TaskBase {
```
blmalone: nit: contract name should be `UnpauseSuperchainConfig.sol` 
```suggestion
contract UnpauseSuperchainConfig is L2TaskBase {
```

src/improvements/template/UnPauseSuperchainConfigAfterU16.sol
Ethnical: Nice catch! 
I have addressed this here: 5cc48eeabc9e3f34b0db8245e93bdebe8edecd2c
This will check that the superchain config is paused.

src/improvements/template/UnPauseSuperchainConfigv400.sol
Ethnical: Thanks! addressed here: 1fe4463b2f549128aa2e7dfd1aacade9ff0b3dbb and this 46a502df59d8f680b2813db2ff056fa4b66b2cc8

src/improvements/template/UnPauseSuperchainConfig.sol
Ethnical: rename here: 620dfb2789c0455f30865b8c96403a1617a10431

src/improvements/template/UnpauseSuperchainConfigV400.sol
```diff
@@ -0,0 +1,92 @@
...
+
+import {Action} from "src/libraries/MultisigTypes.sol";
+
+/// @title UnpauseSuperchainConfigV400 template designed to unpause the SuperchainConfig contract after the U16 upgrade.
+contract UnpauseSuperchainConfigV400 is L2TaskBase {
+    using stdToml for string;
+
+    /// @notice Identifier of the eth_lockbox used in the SuperchainConfig
+    /// @dev Address (20 bytes) that has an associated lockbox contract, loaded from TOML
+
```
blmalone:
```suggestion
```

src/improvements/template/UnpauseSuperchainConfigV400.sol
```diff
@@ -0,0 +1,92 @@
...
+        storageWrites[0] = "SuperchainConfig";
+        return storageWrites;
+    }
+
+    /// @notice Sets up the template with implementation configurations from a TOML file.
+    function _templateSetup(string memory _taskConfigFilePath, address _rootSafe) internal override {
+        super._templateSetup(_taskConfigFilePath, _rootSafe);
+        string memory file = vm.readFile(_taskConfigFilePath);
+        identifier = vm.parseTomlAddress(file, ".identifier"); // Get the identifier of the eth_lockbox from the TOML file
+        // 1. Load the SuperchainConfig contract.
```
blmalone:
```suggestion
        // Load the SuperchainConfig contract.
```

src/improvements/template/UnpauseSuperchainConfigV400.sol
```diff
@@ -0,0 +1,92 @@
...
+        sc = ISuperchainConfig(superchainAddrRegistry.get("SuperchainConfig"));
+    }
+
+    /// @notice Write the calls that you want to execute for the task.
+    function _build(address) internal override {
+        // 2. Make sure that the SuperchainConfig is paused before unpausing for the identifier.
+        assertEq(sc.paused(identifier), true, "ERR99: SuperchainConfig should be Paused for the identifier provided.");
```
blmalone: What's `ERR99`?

src/improvements/template/UnpauseSuperchainConfigV400.sol
```diff
@@ -0,0 +1,92 @@
...
+        super._templateSetup(_taskConfigFilePath, _rootSafe);
+        string memory file = vm.readFile(_taskConfigFilePath);
+        identifier = vm.parseTomlAddress(file, ".identifier"); // Get the identifier of the eth_lockbox from the TOML file
+        // 1. Load the SuperchainConfig contract.
+        sc = ISuperchainConfig(superchainAddrRegistry.get("SuperchainConfig"));
+    }
+
+    /// @notice Write the calls that you want to execute for the task.
+    function _build(address) internal override {
+        // 2. Make sure that the SuperchainConfig is paused before unpausing for the identifier.
```
blmalone: This should be 1 right? I'd probably just remove these numbers entirely.
```suggestion
        // 1. Make sure that the SuperchainConfig is paused before unpausing for the identifier.
```

src/improvements/template/UnpauseSuperchainConfigV400.sol
```diff
@@ -0,0 +1,92 @@
...
+        assertEq(sc.paused(identifier), true, "ERR99: SuperchainConfig should be Paused for the identifier provided.");
+
+        // 3. UnPause the SuperchainConfig contract through the identifier.
```
blmalone:
```suggestion
        // 2. UnPause the SuperchainConfig contract through the identifier.
```

src/improvements/template/UnpauseSuperchainConfigV400.sol
```diff
@@ -0,0 +1,92 @@
...
+        sc.unpause(identifier);
+    }
+
+    /// @notice This method performs all validations and assertions that verify the calls executed as expected.
+    function _validate(VmSafe.AccountAccess[] memory, Action[] memory, address) internal view override {
+        // Validate that the SuperchainConfig contract is unpaused.
+        // 4. check that the Superchain Config is not paused anymore with the identifier provided.
+        assertEq(
+            sc.paused(identifier), false, "ERR100: SuperchainConfig should be unpaused for the identifier provided."
```
blmalone: Why does it start at ERR99 and increment from there? I think there are other patterns where we use codes throughout the code base but it's normally in test files iirc.

src/improvements/template/UnpauseSuperchainConfigV400.sol
```diff
@@ -0,0 +1,92 @@
...
+        assertEq(sc.paused(identifier), true, "ERR99: SuperchainConfig should be Paused for the identifier provided.");
+
+        // 3. UnPause the SuperchainConfig contract through the identifier.
+        sc.unpause(identifier);
+    }
+
+    /// @notice This method performs all validations and assertions that verify the calls executed as expected.
+    function _validate(VmSafe.AccountAccess[] memory, Action[] memory, address) internal view override {
+        // Validate that the SuperchainConfig contract is unpaused.
+        // 4. check that the Superchain Config is not paused anymore with the identifier provided.
```
blmalone: ?
```suggestion
        // 1. check that the Superchain Config is not paused anymore with the identifier provided.
```

src/improvements/template/UnpauseSuperchainConfigV400.sol
```diff
@@ -0,0 +1,92 @@
...
+        assertEq(
+            sc.paused(identifier), false, "ERR100: SuperchainConfig should be unpaused for the identifier provided."
+        );
+        if (
+            identifier != address(0) // If the identifier is 0 this indicates a superchain-wide pause so we don't need to check the lockbox associated
+        ) {
+            assertEq(
+                IETHLockbox(identifier).paused(),
+                false,
+                "ERR103: ETHLockbox should be unpaused for the identifier provided."
```
blmalone: This one jumps to 103. I'm not sure you need these error codes.

src/improvements/template/UnpauseSuperchainConfigV400.sol
```diff
@@ -0,0 +1,92 @@
...
+            );
+        }
+    }
+
+    /// @notice Override to return a list of addresses that should not be checked for code length.
+    function _getCodeExceptions() internal view virtual override returns (address[] memory) {
+        address[] memory codeExceptions = new address[](0);
+        return codeExceptions;
+    }
```
blmalone:
```suggestion
    function _getCodeExceptions() internal view virtual override returns (address[] memory) {}
```

blmalone (commented):
Can you also add a test into `Regression.t.sol` for this new template? Also, do you need to update the superchain-registry as part of this PR?

test/tasks/example/sep/019-unpause-superchainConfig/config.toml
```diff
@@ -0,0 +1,13 @@
...
+templateName = "UnpauseSuperchainConfigV400"
+
+safeAddressString = "SecurityCouncil"
+identifier="0x0000000000000000000000000000000000000000"
+
+
+[stateOverrides]
+# Override the state of the superchain wide pause to 1 to unpause during the task. 
+"0xC2Be75506d5724086DEB7245bd260Cc9753911Be" = [ # Slot calculation by cast index address 0x0000000000000000000000000000000000000000 1
+     {key = "0xa6eef7e35abe7026729641147f7915573c7e97b47efa546f5f6e3230263bcb49", value = "0x000000000000000000000000000000000000000000000000ffffffffff87b31e"} # 0xffffffffff87b31e represent the max timestamp possible - 1 to make sure this always paused.
```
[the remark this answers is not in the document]
Ethnical: I prefer to keep the previous one.

test/tasks/example/sep/019-unpause-superchainConfig/config.toml
```diff
@@ -0,0 +1,13 @@
+l2chains = [{name = "OP Sepolia Testnet", chainId = 11155420}] # e.g. [{name = "OP Mainnet", chainId = 10}]
+
+templateName = "UnpauseSuperchainConfigV400"
+
+safeAddressString = "SecurityCouncil"
+identifier="0x0000000000000000000000000000000000000000"
+
+
+[stateOverrides]
+# Override the state of the superchain wide pause to 1 to unpause during the task.
```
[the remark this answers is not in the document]
Ethnical: I have fixed this in this 66f2cde05ab801af7c4dc2c74238858bc9c92c2a

src/improvements/template/UnpauseSuperchainConfigV400.sol
```diff
@@ -0,0 +1,92 @@
...
+
+    address public identifier;
+
+    /// @notice SuperchainConfig SC contract instance
+    ISuperchainConfig public sc;
+
+    /// @notice Returns the string identifier for the safe executing this transaction.
+    function safeAddressString() public pure override returns (string memory) {
+        return "Guardian";
+    }
```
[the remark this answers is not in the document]
Ethnical: I have removed the `safeAddress` from the `config.toml` in this PR: 82824b63129dc2e8631db9954bdab2e30556b230

src/improvements/template/UnpauseSuperchainConfigV400.sol
```diff
@@ -0,0 +1,88 @@
...
+                IETHLockbox(identifier).paused(),
+                false,
+                "ERR102: ETHLockbox should be unpaused for the identifier provided."
+            );
+        }
+    }
+
+    /// @notice Override to return a list of addresses that should not be checked for code length.
+    function _getCodeExceptions() internal view virtual override returns (address[] memory) {}
+}
```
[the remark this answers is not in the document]
Ethnical: I feel this is more clear to not do that.

src/improvements/template/UnpauseSuperchainConfigV400.sol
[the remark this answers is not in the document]
Ethnical: I don't want to display this in the error section since this is passed from the identifier in the `config.toml`

src/improvements/template/UnpauseSuperchainConfigV400.sol
```diff
@@ -0,0 +1,88 @@
...
+        string memory file = vm.readFile(_taskConfigFilePath);
+        identifier = vm.parseTomlAddress(file, ".identifier"); // Get the identifier of the eth_lockbox from the TOML file
+        // Load the SuperchainConfig contract.
+        sc = ISuperchainConfig(superchainAddrRegistry.get("SuperchainConfig"));
+    }
+
+    /// @notice Write the calls that you want to execute for the task.
+    function _build(address) internal override {
+        // Make sure that the SuperchainConfig is paused before unpausing for the identifier.
+        assertEq(sc.paused(identifier), true, "ERR100: SuperchainConfig should be Paused for the identifier provided.");
```
blmalone: Can you move this check into the `_templateSetup`? While it's true we can add checks here, we've typically added pre-checks into the `_templateSetup` function and kept the `_build` function for any `call` or `delegate` call we want to make.

src/improvements/template/UnpauseSuperchainConfigV400.sol
```diff
@@ -0,0 +1,88 @@
...
+
+        /// UnPause the SuperchainConfig contract through the identifier.
```
blmalone:
```suggestion
        /// Unpause the SuperchainConfig contract through the identifier.
```

src/improvements/template/UnpauseSuperchainConfigV400.sol
blmalone:
```suggestion
        assertEq(sc.paused(identifier), true, "ERR100: SuperchainConfig should be paused for the identifier provided.");
```

Ethnical: thank you I just did in the last commit from my phone.
Let's see if it's pass the checks (maybe the formatting would be broken).

PR #1138, opened 2025-07-18
perf(decode): Small performance improvement for decoding state diffs

blmalone:
No need to call `getStateDiffFor` twice, we can call it once and reuse later.

### DIFF
--- a/src/libraries/AccountAccessParser.sol
+++ b/src/libraries/AccountAccessParser.sol
@@ -207,27 +207,28 @@ library AccountAccessParser {
         // The order of 'uniqueAccounts' informs the order that the account state diffs get processed.
         address[] memory uniqueAccounts = getUniqueWrites(_accountAccesses, _sort);
         uint256 totalDiffCount = 0;
+        StateDiff[][] memory accountDiffs = new StateDiff[][](uniqueAccounts.length);
         // Count the total number of net state diffs.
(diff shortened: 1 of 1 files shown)


src/libraries/AccountAccessParser.sol
```diff
@@ -207,27 +207,28 @@ library AccountAccessParser {
         // The order of 'uniqueAccounts' informs the order that the account state diffs get processed.
         address[] memory uniqueAccounts = getUniqueWrites(_accountAccesses, _sort);
         uint256 totalDiffCount = 0;
+        StateDiff[][] memory accountDiffs = new StateDiff[][](uniqueAccounts.length);
```
Ethnical, line 210: Sgtm! One question just the `StateDiff[][]` was added but before seems it was dynamic size.
Just want to make sure this intended and this can't lead to out of bound. 
otherwise looks good!

blmalone: ah, before we had:
```solidity
for (uint256 i = 0; i < uniqueAccounts.length; i++) {
            StateDiff[] memory accountDiffs = getStateDiffFor(_accountAccesses, uniqueAccounts[i], _sort);
```
so `getStateDiff` was called `uniqueAccounts.length` times. 

This means, we want an array of `StateDiff[]` that's `uniqueAccounts.length` in size.

PR #1149, opened 2025-07-25
refactor(MultisigTask): Remove root address state var

blmalone:
Further refactoring of MultisigTask to remove the root address state variable. We want to remove this to reduce complexity of the code. The more functional, the easier it is to follow the codes flow.


src/improvements/tasks/MultisigTask.sol
```diff
@@ -897,7 +900,7 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager, TaskManage
     /// @notice By default, any value written to storage that looks like an address is expected to
     /// have code. Sometimes, accounts without code are expected, and this function allows you to
     /// specify a list of those addresses.
-    function _getCodeExceptions() internal view virtual returns (address[] memory);
+    function _getCodeExceptions(address _rootSafe) internal view virtual returns (address[] memory);
```
alcueca: It is weird to have to pass the root safe here, and has implications for all templates, just to facilitate GnosisSafeRemoveOwner. Isn't there any other solution for that particular template?

blmalone: I'm open to other suggestions. My opinion is that generally the less state vars the better. This could be an array of addresses to add to the exceptions to make it more generic. Right now it tasks the root safe because some templates need it to look up owners on the safe.

PR #1150, opened 2025-07-25
feat(multi-level-nesting): Support Multi-level Nesting

blmalone:
This PR adds the ability for signing with GnosisSafes at an arbitrary level of depth. Right now, superchain-ops can only sign up to a safe depth of 1 e.g. Foundation safe on the L1PAO. 

This feature will be useful for op-stack chains who have a different safe architecture to the standard OP-Mainnet safe architecture. 

# Testing Steps

@Ethnical and @blmalone executed 3 sepolia tasks which tested the system end-to-end. Specifically,
1. Single
2. Nested
3. The new nested nested setup (e.g. base)


src/improvements/ci.just
```diff
@@ -0,0 +1,114 @@
+# Justfile commands used in CI
```
blmalone, line 1: Basically most of this file is simply moved from `justfile`. The idea was to reduce the size of justfile and keep only user invoked commands there. Anything ci relies on we can put in `ci.just`.

test/tasks/NestedMultisigTask.t.sol
```diff
@@ -159,6 +164,10 @@ contract NestedMultisigTaskTest is Test {
         );
     }
 
+    function testNestedNestedExecuteWithSignatures() public {
+        // TODO: @blmalone implement this test.
```
blmalone: @blmalone need to implement this test.

Ethnical:
Would be nice to add when there is a waiting transaction the line: 

"Please check your hardware wallet and sign physically".

Ethnical:
I think even if this is fake safes etc would be nice to just bump the nonce to make the sep simulation working. 
I know we saw together by testing but would be cleaner wdyt @blmalone ?

test/tasks/NestedMultisigTask.t.sol
blmalone: done in: https://github.com/ethereum-optimism/superchain-ops/pull/1150/commits/7ce4d053ca9571abc6901ad028bc11061ec2a26e

src/improvements/tasks/sep/022-nested-test/addresses.json
Ethnical, line 1: Need to Speak about during the call, this didn't know we had this file.

Ethnical: Ok, used for creating fake chains, it will override completely the SCR when this file is created.

test/tasks/Regression.t.sol
```diff
@@ -669,10 +670,10 @@ contract RegressionTest is Test {
             // TODO: fix when we support > 1 level of nesting.
```
Ethnical, line 670: NIT: todo can be remove

Ethnical, line 670:
```suggestion
```

Ethnical (approved):
Really good work! 🔥 
🚢 it!

test/tasks/Regression.t.sol
blmalone: will do this in a follow up pr.

blmalone: will do in a follow up pr

PR #1161, opened 2025-08-12
perf(retries): Intermittent failures slowing down dev workflow, implement retries for forge scripts

blmalone:
We're seeing persistent 502 errors when we run the justfile command `simulate-all-templates` in CI. After some debugging, we believe the error is upstream in GCP. Therefore, we're going to implement a simple retry attempt on our end. We only need this for the `simulate` step in the `justfile`.


blmalone:
Moving this to draft state. Keeping this in the back burner incase we start to see more CI errors. If we do, we can merge this to see if it helps.

blmalone:
Closing as https://github.com/ethereum-optimism/superchain-ops/pull/1172 implements this feature.

PR #1165, opened 2025-08-14
refactor(psp): remove PSP docs and code

blmalone:
We're not longer using PSP's for pausing the superchain. We have an alternative, less clunky approach to doing this. These files can be removed from superchain-ops.


Ethnical (approved):
LGTM!
Bye bye PSPs you will not be missed!

PR #1173, opened 2025-08-28
feat: add L1PortalExecuteL2Call template for L2 actions through OptimismPortal

0xiamflux:
### Description
- Add L1→L2 portal execution template
  - `src/improvements/template/L1PortalExecuteL2Call.sol`
  - Uses full `IOptimismPortal2` (depositTransaction) to dispatch an L2 call from an L1 Safe
  - Config-driven params via `config.toml`: `portal`, `l2Target`, `l2Data`, `gasLimit` (uint64), `value` (default 0), `isCreation` (default false)
  - Extends `SimpleTaskBase` for L1 multisig flows; `safeAddressString()` defaults to `ProxyAdminOwner`
  - Validation: asserts exactly one portal call with expected calldata/value; allowlists storage/balance changes for `OptimismPortal`
- Add full portal interface
  - `src/improvements/interfaces/IOptimismPortal.sol` (complete `IOptimismPortal2`)
- Scaffold nested rehearsal task
  - `src/improvements/tasks/eth/rehearsals/2025-08-11-l1-to-l2-noop/`
  - `config.toml` placeholders and signer `README.md` with simulate/validate/sign steps

### Tests
- Template-level validation enforces encoded calldata and a single portal action
- Manual simulate run expected:
  - Tenderly diff: no unintended changes; shows portal call and event emission


security-council-rehearsals/rehearsal-4-noop-upgrade/README.md
```diff
@@ -0,0 +1,265 @@
+# Rehearsal 4 - No-op Governor Upgrade
```
blmalone: Why is this being put into the rehearsals directory?

foundry.toml
```diff
@@ -21,6 +21,7 @@ remappings = [
   '@lib-keccak/=lib/optimism/packages/contracts-bedrock/lib/lib-keccak/contracts/lib/',
   'ds-test/=lib/optimism/packages/contracts-bedrock/lib/forge-std/lib/ds-test/src',
   'forge-std/=lib/forge-std/src/',
+  '@optimism/=lib/optimism/'
```
blmalone: what's this update?

security-council-rehearsals/rehearsal-4-noop-upgrade/README.md
0xiamflux: GM, our goal here was to add a rehearsal task for a no-op upgrade on the Optimism Governor Proxy on L2 using a newly added template. What would be the best place to do this?

foundry.toml
0xiamflux: We needed to use the portal for the new template, would it be better if we add a standalone interface in the superchain ops repo instead?

0xiamflux: Update: Removed the unnecessary remapping [here](https://github.com/ethereum-optimism/superchain-ops/pull/1173/commits/a9b18e602e6e117bc7c8385d57d29006dbbc4591).

security-council-rehearsals/rehearsal-4-noop-upgrade/README.md
blmalone: hm, this rehearsals directory contains templates and ceremonies related to onboarding Security Council members. I don't think this is what you want to do right?

0xiamflux: The task was *to create a rehearsal upgrade ceremony for the Security Council to complete an end-to-end process of a no-op upgrade in a way that is repeatable.* 

Which would be a better suited place to put it?

blmalone: Is this simply to test this templates functionality? Or have you been in touch with the security council and agreed that this should be a new rehearsal for them to perform when onboarding a security council memeber? 

If it's the former, you don't need to add this as a rehearsal. When you create a new template, you just need to follow this guide: https://github.com/ethereum-optimism/superchain-ops/blob/main/src/improvements/doc/NEW_TEMPLATE_GUIDE.md (1. make template 2. make Regression test 3. make example task

src/improvements/template/L1PortalExecuteL2Call.sol
```diff
@@ -0,0 +1,121 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import {VmSafe} from "forge-std/Vm.sol";
+import {stdToml} from "forge-std/StdToml.sol";
+
+import {MultisigTaskPrinter} from "../../libraries/MultisigTaskPrinter.sol";
+import {Action} from "../../libraries/MultisigTypes.sol";
+
```
blmalone:
```suggestion
```

src/improvements/template/L1PortalExecuteL2Call.sol
```diff
@@ -0,0 +1,121 @@
...
+import {IOptimismPortal2} from "lib/optimism/packages/contracts-bedrock/interfaces/L1/IOptimismPortal2.sol";
+
```
blmalone:
```suggestion
```

src/improvements/template/L1PortalExecuteL2Call.sol
```diff
@@ -0,0 +1,121 @@
...
+            }
+        }
+        require(_found && _matches == 1, "expected one portal deposit action");
+        MultisigTaskPrinter.printTitle("Validated portal deposit action");
+    }
+
+    /// @notice No code exceptions required for this template.
+    function _getCodeExceptions() internal view override returns (address[] memory) {
+        return new address[](0);
+    }
```
blmalone:
```suggestion
    /// @notice no code exceptions for this template
    function _getCodeExceptions() internal view virtual override returns (address[] memory) {}
```

src/improvements/template/L1PortalExecuteL2Call.sol
```diff
@@ -0,0 +1,121 @@
...
+
+    /// @notice Build the portal deposit action. WARNING: State changes here are reverted after capture.
+    function _build(address) internal override {
+        // Record the L1 portal call with value for action extraction.
+        IOptimismPortal2(portal).depositTransaction{value: valueWei}(l2Target, valueWei, gasLimit, isCreation, l2Data);
+    }
+
+    /// @notice Validate that exactly one action to the portal with the expected calldata and value was captured.
+    function _validate(VmSafe.AccountAccess[] memory, Action[] memory _actions, address) internal override {
```
blmalone:
```suggestion
    function _validate(VmSafe.AccountAccess[] memory, Action[] memory _actions, address) internal view override {
```

src/improvements/template/L1PortalExecuteL2Call.sol
```diff
@@ -0,0 +1,121 @@
...
+        isCreation = false;
+        try vm.parseTomlBool(_toml, ".isCreation") returns (bool _b) {
+            isCreation = _b;
+        } catch {}
+    }
+
+    /// @notice Build the portal deposit action. WARNING: State changes here are reverted after capture.
+    function _build(address) internal override {
+        // Record the L1 portal call with value for action extraction.
+        IOptimismPortal2(portal).depositTransaction{value: valueWei}(l2Target, valueWei, gasLimit, isCreation, l2Data);
```
blmalone, line 112: does this need to take a value? I'm curious as I don't think we've fully tested if the value actually gets transferred.

0xiamflux: For the particular use case we had in mind when adding the template *it does not* need the value, but I think for a general purpose solution is good to have it there. What would be the proper way to fully test and document this?

security-council-rehearsals/rehearsal-4-noop-upgrade/README.md
0xiamflux: We just removed the rehearsal docs in the rehearsal folder on [this](https://github.com/ethereum-optimism/superchain-ops/pull/1173/commits/51d63c5734af455221f9e27f24085f787afb4da3) commit.

src/improvements/template/L1PortalExecuteL2Call.sol
0xiamflux: Addressed [here](https://github.com/ethereum-optimism/superchain-ops/pull/1173/commits/543b409e13a6f27d1b2ea8ce9197376f71abc4c8).

src/improvements/template/L1PortalExecuteL2Call.sol
0xiamflux: Addressed [here](https://github.com/ethereum-optimism/superchain-ops/pull/1173/commits/543b409e13a6f27d1b2ea8ce9197376f71abc4c8).

src/improvements/template/L1PortalExecuteL2Call.sol
0xiamflux: Addressed [here](https://github.com/ethereum-optimism/superchain-ops/pull/1173/commits/543b409e13a6f27d1b2ea8ce9197376f71abc4c8).

src/improvements/template/L1PortalExecuteL2Call.sol
0xiamflux: Addressed [here](https://github.com/ethereum-optimism/superchain-ops/pull/1173/commits/543b409e13a6f27d1b2ea8ce9197376f71abc4c8).

security-council-rehearsals/README.md
```diff
@@ -5,6 +5,7 @@ This directory contains templates and ceremonies related to onboarding Security
 ## Creating and signing a new rehearsal ceremony
 
 To create a new rehearsal ceremony, follow the instructions in the _Facilitator_ section of the README files for each of the following rehearsals:
+
```
mds1: nit, can undo this diff since we don't touch the rehearsals dir anywhere else here

src/improvements/template/L1PortalExecuteL2Call.sol
```diff
@@ -0,0 +1,117 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import {VmSafe} from "forge-std/Vm.sol";
+import {stdToml} from "forge-std/StdToml.sol";
+
+import {IOptimismPortal2} from "lib/optimism/packages/contracts-bedrock/interfaces/L1/IOptimismPortal2.sol";
```
mds1: Instead of importing the interface, let's define the relevant interface in this file with `interface IOptimismPortal2 { ... }`. The reason for this is so that, in the future, the template can keep working for chains running older versions of the OP stack, even if the monorepo submodule version is bumped. In other words, we don't want templates to suddenly break when we bump the submodule version (i.e. if we change the portal interface in the future) because not all chains are necessarily on the latest version at all times, so we prefer to make sure templates continue to work

src/improvements/template/L1PortalExecuteL2Call.sol
```diff
@@ -0,0 +1,117 @@
...
+
+/// @notice Template to execute an L2 call via the L1 Optimism Portal from a nested L1 Safe.
+/// Sends an L2 transaction using OptimismPortal.depositTransaction with config-driven params.
+contract L1PortalExecuteL2Call is SimpleTaskBase {
+    using stdToml for string;
+
+    // -------- Config inputs --------
+    address payable public portal; // L1 OptimismPortal address
+    address public l2Target; // L2 target address
+    bytes public l2Data; // Inner L2 calldata
```
mds1: nit, just to be more precise, assuming this is correct
```suggestion
    bytes public l2Data; // Inner L2 calldata, i.e. the calldata to execute on `l2Target`
```

src/improvements/template/L1PortalExecuteL2Call.sol
```diff
@@ -0,0 +1,117 @@
...
+
+        // Optional fields
+        valueWei = 0;
+        try vm.parseTomlUint(_toml, ".value") returns (uint256 _v) {
+            valueWei = _v;
+        } catch {}
+
+        isCreation = false;
+        try vm.parseTomlBool(_toml, ".isCreation") returns (bool _b) {
+            isCreation = _b;
```
mds1, line 102: The portal has this check so it's arguably unnecessary, but might be better UX to check and revert here if `isCreation && l2Target != address(0)`. Up to you

src/improvements/template/L1PortalExecuteL2Call.sol
```diff
@@ -0,0 +1,117 @@
...
+    /// - isCreation: bool (optional, default false)
+    function _templateSetup(string memory _taskConfigFilePath, address) internal override {
+        string memory _toml = vm.readFile(_taskConfigFilePath);
+
+        // Resolve portal from registry first if available, else read explicit field.
+        try simpleAddrRegistry.get("OptimismPortal") returns (address p) {
+            portal = payable(p);
+        } catch {
+            portal = payable(_toml.readAddress(".portal"));
+        }
```
mds1, line 80: For the revshare use case, it might be better UX (fewer playbooks, fewer signatures) to have this template support doing a deposit transaction for many L2s in the same transaction. In that case, we would want to inherit from `L2TaskBas` instead and have the config inputs be arrays corresponding to the deposit. What do you think here?

src/improvements/template/L1PortalExecuteL2Call.sol
```diff
@@ -0,0 +1,117 @@
...
+        require(portal != address(0), "portal must be set (addresses.OptimismPortal or .portal)");
+
+        l2Target = _toml.readAddress(".l2Target");
+        require(l2Target != address(0), "l2Target must be set");
+
+        // Read hex string and parse to bytes.
+        string memory _dataHex = _toml.readString(".l2Data");
+        l2Data = vm.parseBytes(_dataHex);
+        require(l2Data.length > 0, "l2Data must be set");
```
mds1, line 88: Could just do `_toml.readBytes` here

src/improvements/template/L1PortalExecuteL2Call.sol
```diff
@@ -0,0 +1,117 @@
...
+    function _build(address) internal override {
+        // Record the L1 portal call with value for action extraction.
+        IOptimismPortal2(portal).depositTransaction{value: valueWei}(l2Target, valueWei, gasLimit, isCreation, l2Data);
+    }
+
+    /// @notice Validate that exactly one action to the portal with the expected calldata and value was captured.
+    function _validate(VmSafe.AccountAccess[] memory, Action[] memory _actions, address) internal view override {
+        bytes memory _expected = abi.encodeWithSelector(
+            IOptimismPortal2.depositTransaction.selector, l2Target, valueWei, gasLimit, isCreation, l2Data
+        );
```
mds1: nit, prefer `abi.encodeCall`

src/improvements/tasks/sep/019-worldchain-l1-ownership-transfers/VALIDATION.md
mds1, line 1: This task was recently changed to CANCELLED so we should be able to undo the diff in this task's dir

src/improvements/template/L1PortalExecuteL2Call.sol
0xiamflux: I do like this, makes sense to early revert before reaching the portal.

src/improvements/template/L1PortalExecuteL2Call.sol
0xiamflux: Do you think it makes sense to have 2 separate templates one for simpler use cases and the other for multi chain calls?

src/improvements/template/L1PortalExecuteL2Call.sol
blmalone: Can you hardcode `0` for this first version? 

When performing an upgrade action i.e. `portal.depositTransaction`, this function is executed inside a Multicall3 contract. The safe delegatecall's to the Multicall3 contract. Right now, this will fail because of the require check inside the Multicall3 contract: https://github.com/mds1/multicall3/blob/main/src/Multicall3.sol#L160 (see: [stackexchange](https://ethereum.stackexchange.com/questions/102605/why-cant-you-specify-the-message-value-when-using-delegatecall))

src/improvements/template/L1PortalExecuteL2Call.sol
0xiamflux: As we continued working on the Rev Share use case, we have came to the conclusion it should have it’s own, separate template and we can keep this as a general purpose template for more simple cases. Wdyt?

src/improvements/template/L1PortalExecuteL2Call.sol
0xiamflux: Addressed in [here](https://github.com/ethereum-optimism/superchain-ops/pull/1231/commits/8bdcf19d4716a1929fff2b74bbab4899d066f5f4).

blmalone: got it, do you want to close this PR in favor of #1231?

src/improvements/template/L1PortalExecuteL2Call.sol
mds1: Having a separate template is good with me!

PR #1210, opened 2025-09-19
fix: nonce update and hash for worldchain

blmalone:
Worldchain nonce update along with hash.


Wazabie (approved):

Ethnical (approved):
Thank you, make sense to me to cancelled this one instead the bumping the nonce each time :)

PR #1231, opened 2025-09-25
feat: L2 call via portal template

0xiamflux:
### Description
- Add L1→L2 portal execution template
  - `src/improvements/template/L1PortalExecuteL2Call.sol`
  - Uses full `IOptimismPortal2` (depositTransaction) to dispatch an L2 call from an L1 Safe
  - Config-driven params via `config.toml`: `portal`, `l2Target`, `l2Data`, `gasLimit` (uint64), `isCreation` (default false)
  - Extends `SimpleTaskBase` for L1 multisig flows; `safeAddressString()` defaults to `ProxyAdminOwner`
  - Validation: asserts exactly one portal call with expected calldata/value; allowlists storage/balance changes for `OptimismPortal`
  - 
### Tests
- Template-level validation enforces encoded calldata and a single portal action
- Manual simulate run expected:
  - Tenderly diff: no unintended changes; shows portal call and event emission
- Regression tests


src/template/L1PortalExecuteL2Call.sol
```diff
@@ -0,0 +1,123 @@
...
+
+/// @notice Interface for the OptimismPortal2 contract on L1.
+interface IOptimismPortal2 {
+    function depositTransaction(address _to, uint256 _value, uint64 _gasLimit, bool _isCreation, bytes memory _data)
+        external
+        payable;
+}
+
+/// @notice Template to execute an L2 call via the L1 Optimism Portal from a nested L1 Safe.
+/// Sends an L2 transaction using OptimismPortal.depositTransaction with config-driven params.
```
blmalone, line 20: Not strictly necessary but might be nice to add the op-contracts version to the natspec here. i.e. "This template is known to work with op-contracts/X.Y.Z".

src/template/L1PortalExecuteL2Call.sol
```diff
@@ -0,0 +1,123 @@
...
+contract L1PortalExecuteL2Call is SimpleTaskBase {
```
blmalone: I was thinking this _could_ inherit from `L2TaskBase` instead of SimpleTaskBase if you anticipate bundling many L1Portal calls for many chains in the one task. The OptimismPortal added doesn't need to be added to the config.toml file then. It'll be available via something like: 

```
        OptimismPortal optimismPortal =
            OptimismPortal(superchainAddrRegistry.getAddress("OptimismPortalProxy", chains[0].chainId));
```

blmalone (commented):
It's looking good! one q about the inheritance.

src/template/L1PortalExecuteL2Call.sol
0xiamflux: I think this is a good idea. Will add it.

src/template/L1PortalExecuteL2Call.sol
0xiamflux: In the case we make it a `L2TaskBase` for bundling calls to different chains do you think it would be a problem we have a single `l2Target` for the template? i.e. it's use would be limited to contracts that share the same address across chains, right?

mds1: I do like the idea of not needing to specify the portal in the config file, removes a source of error. Having a single target / array length of one is ok, if I understood the question correctly

blmalone: Yea I agree with @mds1. Let's change this to inherit from `L2TaskBase`. Having an array of 1 is totally fine, we do this frequently.

0xiamflux: We addressed the change of type of the template [here](https://github.com/ethereum-optimism/superchain-ops/pull/1231/commits/28f403f636da975850d0900713ca6d496c218295).

src/template/L1PortalExecuteL2Call.sol
0xiamflux: Addressed [here](https://github.com/ethereum-optimism/superchain-ops/pull/1231/commits/3e491b5353c2ec5b9b9fbad8c36eb2cf9e178eda).

src/template/L1PortalExecuteL2Call.sol
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    /// @notice Parse config and initialize template variables.
+    /// Expected TOML keys:
+    /// - portal: address (L1 OptimismPortal) OR addresses.OptimismPortal in [addresses]
+    /// - l2Target: address (L2 target address)
+    /// - l2Data: hex string (e.g. 0x1234...)
+    /// - gasLimit: uint (will be cast to uint64)
+    /// - value: uint (optional, default 0)
+    /// - isCreation: bool (optional, default false)
```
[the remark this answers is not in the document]
blmalone: @0xiamflux can you remove the `value` from the natspec?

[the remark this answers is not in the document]
0xiamflux: Yeah, of course!

blmalone (approved):

PR #1237, opened 2025-10-02
Fusaka Prestate Update Template and OP Task

Wazabie:
This PR adds the template to update prestates of chains on smart contract version 4.1.0

The template was written after OPCMUpdatePrestateV300 - the only change is the hardcoded checks adjusted for V410

The PR also adds: 
- an example test task (based off of OP Sepolia
- the updated Regression test, to include the above example task
- the OP Sepolia actual prestate update task


src/tasks/sep/034-op-ink-sep-fusaka-prestate/config.toml
```diff
@@ -0,0 +1,37 @@
...
+[[opcmUpgrades]]
+chainId = 763373
+absolutePrestate = "0x03eb07101fbdeaf3f04d9fb76526362c1eea2824e4c6e970bdb19675b72e4fc9"  # op-program/v1.6.1 - https://github.com/ethereum-optimism/superchain-registry/blob/47f21fb1291999cb9a46653dd92d7cd55b6e3821/validation/standard/standard-prestates.toml#L6C9-L6C75
+# The above prestate needs to be updated with the correct Fusaka Absolute Prestate
+expectedValidationErrors = ""
+
+
+[addresses]
+OPCM = "0x3bb6437aba031afbf9cb3538fa064161e2bf2d78" # version 3.2.0 https://github.com/ethereum-optimism/superchain-registry/blob/40526b1288534f6b84b7aae21d13c0b5f5b12f47/validation/standard/standard-versions-sepolia.toml#L23
+StandardValidatorV410 = "0x7b4d2a02d5fa6c7c98d835d819956ebb876ff439" # Sepolia https://github.com/ethereum-optimism/optimism/blob/f79ed8b9c9cbdbf8bb492074f3f98da7f072e21a/op-validator/pkg/validations/addresses.go#L30
```
mbaxter: The [link here](https://github.com/ethereum-optimism/optimism/blob/f79ed8b9c9cbdbf8bb492074f3f98da7f072e21a/op-validator/pkg/validations/addresses.go#L30) does not match the V410 standard validator.

src/tasks/sep/034-op-ink-sep-fusaka-prestate/config.toml
```diff
@@ -0,0 +1,37 @@
...
+
+[stateOverrides]
```
mbaxter, line 24: Why do we need the state overrides here?

src/tasks/sep/034-op-ink-sep-fusaka-prestate/config.toml
```diff
@@ -0,0 +1,37 @@
...
+name = "OP Sepolia Testnet"
+
+[[l2chains]]
+chainId = 763373
+name = "Ink Sepolia Testnet"
+
+[[opcmUpgrades]]
+chainId = 11155420
+absolutePrestate = "0x03eb07101fbdeaf3f04d9fb76526362c1eea2824e4c6e970bdb19675b72e4fc9"  # op-program/v1.6.1 - https://github.com/ethereum-optimism/superchain-registry/blob/47f21fb1291999cb9a46653dd92d7cd55b6e3821/validation/standard/standard-prestates.toml#L6C9-L6C75
+# The above prestate needs to be updated with the correct Fusaka Absolute Prestate
```
mbaxter: I'm not sure what our normal process is, but it seems like it would be better to follow-up with a PR that includes this task with the correct data.

test/tasks/example/sep/028-opcm-update-prestate-v410/config.toml
```diff
@@ -0,0 +1,27 @@
+templateName = "OPCMUpdatePrestateV410"
+
+[[l2chains]]
+chainId = 11155420
+name = "OP Sepolia Testnet"
+
+[[opcmUpgrades]]
+chainId = 11155420
+absolutePrestate = "0x03eb07101fbdeaf3f04d9fb76526362c1eea2824e4c6e970bdb19675b72e4fc9"  # op-program/v1.6.1 - https://github.com/ethereum-optimism/superchain-registry/blob/47f21fb1291999cb9a46653dd92d7cd55b6e3821/validation/standard/standard-prestates.toml#L6C9-L6C75
```
mbaxter: fyi - there is a typo in these prestates:
```suggestion
absolutePrestate = "0x03eb07101fbdeaf3f04d9fb76526362c1eea2824e4c6e970bdb19675b72e4fc8"  # op-program/v1.6.1 - https://github.com/ethereum-optimism/superchain-registry/blob/47f21fb1291999cb9a46653dd92d7cd55b6e3821/validation/standard/standard-prestates.toml#L6C9-L6C75
```

test/tasks/example/sep/028-opcm-update-prestate-v410/config.toml
```diff
@@ -0,0 +1,27 @@
...
+# The above prestate needs to be updated with the correct Fusaka Absolute Prestate
```
mbaxter: (nit) Don't need this comment for this test file

test/tasks/Regression.t.sol
```diff
@@ -418,6 +419,35 @@ contract RegressionTest is Test {
         _assertDataToSignNestedMultisig(multisigTask, actions, expectedDataToSign, MULTICALL3_ADDRESS, rootSafe);
     }
 
+    /// @notice Expected call data and data to sign generated by manually running the OPCMUpdatePrestateV410 template at block 8098229 on sepolia
```
mbaxter: Stale block number here:
```suggestion
    /// @notice Expected call data and data to sign generated by manually running the OPCMUpdatePrestateV410 template at block 9327881 on sepolia
```

src/template/OPCMUpdatePrestateV410.sol
```diff
@@ -0,0 +1,149 @@
...
+        address OPCM = tomlContent.readAddress(".addresses.OPCM");
+        OPCM_TARGETS.push(OPCM);
+        require(IOPContractsManager(OPCM).version().eq("3.2.0"), "Incorrect OPCM - expected version 3.2.0");
+        vm.label(OPCM, "OPCM");
+
+        STANDARD_VALIDATOR_V410 = IStandardValidatorV410(tomlContent.readAddress(".addresses.StandardValidatorV410"));
+        require(
+            address(STANDARD_VALIDATOR_V410).code.length > 0, "Incorrect StandardValidatorV410 - no code at address"
+        );
+        require(
```
mbaxter: Shouldn't we validate that the standard validator address comes from the OPCM?  Or better - why don't we just grab the [standard validator address from the OPCM](https://github.com/ethereum-optimism/optimism/blob/op-contracts/v4.1.0/packages/contracts-bedrock/interfaces/L1/IOPContractsManager.sol#L373) instead of configuring it?

test/tasks/example/sep/028-opcm-update-prestate-v410/config.toml
Wazabie: Hey @mbaxter this was intended, the plan here is to prepare the tasks ahead of time and update them with the correct Fusaka Prestate as soon as it is available. 

All Fusaka Prestate tasks have the same placeholder/wrong prestate`0x03eb07101fbdeaf3f04d9fb76526362c1eea2824e4c6e970bdb19675b72e4fc9`. The typo is intended because the correct prestate (ending with `8`) is already used by these chains and the task would not simulate.

mbaxter: Probably better to use something that looks obviously wrong like `0xdead000000000000000000000000000000000000000000000000000000000000`.

src/template/OPCMUpdatePrestateV410.sol
mds1: +1 let's just read it from the OPCM directly (in the OPCM version used by the prior version of this template, the standard validator getter did not exist yet)

src/template/OPCMUpdatePrestateV410.sol
```diff
@@ -0,0 +1,149 @@
...
+            upgrades[_upgrades[i].chainId] = _upgrades[i];
+        }
+
+        address OPCM = tomlContent.readAddress(".addresses.OPCM");
+        OPCM_TARGETS.push(OPCM);
+        require(IOPContractsManager(OPCM).version().eq("3.2.0"), "Incorrect OPCM - expected version 3.2.0");
+        vm.label(OPCM, "OPCM");
+
+        STANDARD_VALIDATOR_V410 = IStandardValidatorV410(tomlContent.readAddress(".addresses.StandardValidatorV410"));
+        require(
```
mds1: related to above comment, here we can just call `opcm.opcmStandardValidator()` to get this address and then we can simplify the config file. And reminder to define that solidity interface inline in this template directly:

```solidity
interface IOPCM {
  function opcmStandardValidator() external view returns (IOPContractsManagerStandardValidator);
}

File: src/template/OPCMUpdatePrestateV410.sol [hunk 9]
```diff
@@ -0,0 +1,149 @@
...
+            IStandardValidatorV410.InputV410 memory input = IStandardValidatorV410.InputV410({
+                proxyAdmin: proxyAdmin,
+                sysCfg: sysCfg,
+                absolutePrestate: expAbsolutePrestate,
+                l2ChainID: chainId
+            });
+
+            string memory errors = STANDARD_VALIDATOR_V410.validate({_input: input, _allowFailure: true});
+
+            require(errors.eq(expErrors), string.concat("Unexpected errors: ", errors, "; expected: ", expErrors));
```
mds1, line 125: We can also change `validate` to instead call `validateWithOverrides` which will let us have fewer to no expected errors in the config files, see [here](https://github.com/ethereum-optimism/optimism/blob/594bc933a38425f745b46399a3619bcdeb74965d/packages/contracts-bedrock/src/L1/OPContractsManagerStandardValidator.sol#L798) and [here](https://oplabs-pbc.slack.com/archives/C04FKKKHWBW/p1759269480877759)

src/tasks/sep/034-op-ink-sep-fusaka-prestate/README.md
```diff
@@ -0,0 +1,21 @@
...
+Simulation commands for each safe:
+```bash
+cd src/tasks/sep/034-op-ink-sep-fusaka-prestate
+SIMULATE_WITHOUT_LEDGER=1 just --dotenv-path $(pwd)/.env simulate <council|foundation>
+```
+
+Signing commands for each safe:
+```bash
+cd src/tasks/sep/034-op-ink-sep-fusaka-prestate
+just --dotenv-path $(pwd)/.env sign <council|foundation>
```
mds1: Let's also add `SKIP_DECODE_AND_PRINT=1` to the just commands, like we did here: https://github.com/ethereum-optimism/superchain-ops/pull/1241/files

The reason for this is because printing and decoding can be very slow, and multiple signers complained about this in the U16A tasks, and the decode and print step is a large reason for the slowness

src/tasks/sep/034-op-ink-sep-fusaka-prestate/config.toml
Wazabie: Yes, we'll need to make sure we use these nonces

src/tasks/sep/034-op-ink-sep-fusaka-prestate/config.toml
Wazabie: Updating the template so I'll get the validator address from OPCM

PR #1387, opened 2026-03-18
Fee vault upgrade template - FeeVaultUpgrader

Wazabie:
Fee vault upgrade template based on FeeVaultUpgrader.

It differs from the template #1381 because this one deploys new FeeVault implementation contracts to L2 and upgrades the proxies to them via upgradeAndCall + setRecipient in one atomic batch (v1.6.0 settable-recipient approach). UpdateFeeVaultRecipient in #1381 assumes implementations are already deployed and simply calls upgradeTo to point the proxies at them (immutable-recipient approach).


Wazabie:
Closing as the template is being added in #1430

PR #1396, opened 2026-04-15
Deputy pause key rotation tasks for eth and sep

JosepBove:


src/tasks/sep/074-deputy-pause-key-rotation/config.toml
```diff
@@ -0,0 +1,12 @@
+l2chains = [{name = "OP Sepolia Testnet", chainId = 11155420}]
+templateName = "DeputyPauseKeyRotationTemplate"
+
+newDeputy = "0x8D2AAe4009418Ef6D83F1F2c90D4dAc3cE2b5D4f"
+
+newDeputySignature = "0x3d3d1206bfb51e9d6db5ab1d474619cffa44907b5f2055966fa26cc85c6783f45f78cdfb0300f89763e29357c9f077f1ae83cc83a2711f51f1b4c84478287e8e1c"
```
Ethnical, line 6: Just for best practice we should separate this PRs in 2 PRs one for mainnet and one for sepolia in case we catch something on the sepolia would be nice to identify an separate PRs.

Ethnical (changes requested):
Please split this into two separate PRs:

1. **Sepolia PR** — keep `src/tasks/sep/074-deputy-pause-key-rotation/` and the `src/addresses.toml` change (if sepolia-specific)
2. **Mainnet PR** — move `src/tasks/eth/052-deputy-pause-key-rotation/` into its own PR

This way we can land and execute the sepolia rotation first, confirm everything works as expected, and then proceed with mainnet independently. If we catch an issue on sepolia, the mainnet task stays clean and unblocked by any fix-up commits.

PR #1408, opened 2026-04-28
U19 template

Wazabie:
# U19 OPCM upgrade template

A single PAO-signed transaction that delegatecalls into OPCMv2 (`op-contracts/v7.1.17`).

### 1. `OPCM.upgradeSuperchain(...)`

Called once per task. Re-initialises the shared `SuperchainConfig` proxy with the new v7.1.17 impl.

### 2. `OPCM.upgrade(UpgradeInput)`

Called once per L2 chain. The OPCM executes in the PAO Safe's context via `Multicall3DelegateCall`, then internally:

- **2a. Bumps proxy implementations** to their v7.1.17 release impls and re-runs each initialiser, for:
  `SystemConfig`, `OptimismPortal`, `L1CrossDomainMessenger` (via `AddressManager`), `L1StandardBridge`, `L1ERC721Bridge`, `OptimismMintableERC20Factory`, `DisputeGameFactory`, `DelayedWETH`, `AnchorStateRegistry`.

- **2b. Wires `SystemConfig`'s two new fields**:
  - `delayedWETH` → the freshly-deployed `DelayedWETH` proxy
  - `opcm` → the v7.1.17 OPCM address

- **2c. Deploys a new `DelayedWETH` proxy** for the new game's bonds, authorised by the `PermittedProxyDeployment = \"DelayedWETH\"` extraInstruction.

- **2d. Disables `CANNON` in the `DisputeGameFactory`**:
  - `gameImpls[CANNON] = address(0)`
  - `initBond = 0`
  - New `CANNON` games can no longer be created; existing ones still resolve to completion.

- **2e. Rewires `PERMISSIONED_CANNON`** in the DGF to its v7.1.17 impl with the supplied \`cannonPrestate\` + chain \`(proposer, challenger)\`. Done **unconditionally** on every U19 chain — installed fresh if the slot wasn't live, rewired if it was — so every chain comes out of U19 with a permissioned fallback game ready for emergency Guardian rollback.

- **2f. Installs `CANNON_KONA`** in the DGF:
  - `gameImpls[CANNON_KONA]` → v7.1.17 `FaultDisputeGame` impl
  - `gameArgs = abi.encode(cannonKonaPrestate)` (the Kona-built prestate)
  - `initBond = 0.08 ETH`

- **2g. Leaves super-root and ZK slots disabled** — `SUPER_CANNON`, `SUPER_PERMISSIONED_CANNON`, `SUPER_CANNON_KONA`, `ZK_DISPUTE_GAME` stay empty. The v7.1.17 OPCM container ships zero impls for them; they belong to a later release.

- **2h. Rotates the respected game type to `CANNON_KONA` (8)** — happens inside the `AnchorStateRegistry`'s re-initialisation in step 2a, fed by the \`overrides.cfg.startingRespectedGameType = 8\` extraInstruction. The same SSTORE the Guardian's \`setRespectedGameType\` would write, but here it's done atomically inside the upgrade tx via PAO authorisation (\`_assertOnlyProxyAdminOrProxyAdminOwner\`).

### 3. Post-execution validation

Calls \`OPCMStandardValidator.validate(...)\` — or \`validateWithOverrides(...)\` for non-standard L1PAO / Challenger like betanets — and asserts the returned error string matches \`expectedValidationErrors\` from the TOML.

### 4. State-diff invariants (inherited from `OPCMTaskBase` / `MultisigTask`)

Auto-checked, no per-task code:

- Every storage write hit either an allowlisted contract (the 15 identifiers in \`_taskStorageWrites()\`, including the \`PermissionlessWETH\` entry) or a newly deployed contract.
- The rootSafe nonce incremented by exactly 1.
- Any storage value that looks like an address either points at code or is in \`_getCodeExceptions()\` — which includes the SystemConfig EOAs (\`owner\`, \`unsafeBlockSigner\`, \`batchInbox\`, batcherHash-derived address), since v7.1.17 reinitialises SystemConfig with these as plain EOAs on betanets.

---

**One signer (PAO Safe), one atomic tx.**


donoso-eth (dismissed):
Two findings from a deeper pass on the example task:

✅ 1. Recommendation: add a gameImpls[CANNON] override to the example task
u19-beta-0 is permissioned-only, so gameImpls[CANNON=0] was already 0x0, the OPCM's disable write produces no observable delta. Pre-positioning to a placeholder makes the disable visible in the state diff:

```
# Forces a visible CANNON disable delta on chains where CANNON wasn't live.
# gameImpls[0] = keccak256(abi.encode(uint32(0), uint256(101)))
0x90dFd26A757f2f9c6516b7E6466C9Eef8eE6030a = [
    {key = "0xffdfc1249c027f9191656349feb0761381bb32c9f557e01f419fd08754bf5a1b", value = "0x000000000000000000000000ca11000000000000000000000000000000dead00"}
]
```

⚠️ 2. EthLockbox + ProtocolVersions don't appear to be upgraded
Per the v7.1.17 contracts release, both have new impls:

```
EthLockbox v1.3.1 → 0xb3a24db0... (currently 0x784d2f03... v1.2.0)
ProtocolVersions v1.1.2 → 0x15f5edb6... (currently 0x1f734b89... v1.1.1)
```
  But the state diff shows no writes to either proxy. Verified also via override probe: pre-positioning both impl slots to placeholders produced no delta,  the OPCM doesn't write to them.
  
 Happy to dismiss if this is intentional or my testing is wrong!

Wazabie:
> Two findings from a deeper pass on the example task:
> 
> ✅ 1. Recommendation: add a gameImpls[CANNON] override to the example task u19-beta-0 is permissioned-only, so gameImpls[CANNON=0] was already 0x0, the OPCM's disable write produces no observable delta. Pre-positioning to a placeholder makes the disable visible in the state diff:
> 
> ```
> # Forces a visible CANNON disable delta on chains where CANNON wasn't live.
> # gameImpls[0] = keccak256(abi.encode(uint32(0), uint256(101)))
> 0x90dFd26A757f2f9c6516b7E6466C9Eef8eE6030a = [
>     {key = "0xffdfc1249c027f9191656349feb0761381bb32c9f557e01f419fd08754bf5a1b", value = "0x000000000000000000000000ca11000000000000000000000000000000dead00"}
> ]
> ```
> 
> ⚠️ 2. EthLockbox + ProtocolVersions don't appear to be upgraded Per the v7.1.17 contracts release, both have new impls:
> 
> ```
> EthLockbox v1.3.1 → 0xb3a24db0... (currently 0x784d2f03... v1.2.0)
> ProtocolVersions v1.1.2 → 0x15f5edb6... (currently 0x1f734b89... v1.1.1)
> ```
> 
> But the state diff shows no writes to either proxy. Verified also via override probe: pre-positioning both impl slots to placeholders produced no delta, the OPCM doesn't write to them.
> 
> Happy to dismiss if this is intentional or my testing is wrong!

Thanks @donoso-eth for pointing these out. For the first comment, I think it's better to actually test with a permissionless devnet than adding the placeholder. Planning to do that in the sepolia directory after we do the permissionless switch on U19 betanet 1.

For point 2, EDIT: cross checked internally and it looks ok to not set them. On the proposal branch, if a chain already has a lockbox it keeps it otherwise it is not set. We also created the proposal branch before the PV contract was fully removed (it's being removed), so it's correct that it's deployed but not set.

donoso-eth (dismissed):

Wazabie:
## Template Review Findings — `OPCMUpgradeV700.sol`

Generated by [`superchain-ops-template-review`](https://github.com/ethereum-optimism/op-claude-marketplace) skill (Security-team rigor checklist). Live values cross-checked against a sepolia archive endpoint.

### Audit Trail

| Step | Action | Result |
|------|--------|--------|
| 1 | `forge build --deny-warnings` | PASS (0 warnings, Solc 0.8.15) |
| 3 | `./src/script/check-struct-order.sh` | PASS (`OPCMUpgrade` in `OPCMUpgradeV700.sol` ordered: `cannonKonaPrestate`, `cannonPrestate`, `chainId`, `expectedValidationErrors`, `initBond`, `startingRespectedGameType`) |
| 4 | `forge test --skip Integration` | Run twice. With public-node sepolia: 192 PASS / 66 FAIL (every failure was `historical state ... not available`, RPC archive limitation). With a sepolia archive endpoint: re-ran the V700 regression test specifically — **`testRegressionCallDataMatches_OPCMUpgradeV700_U19_Betanet` PASSES** solo (gas: 212,076,261, 25.97s). Confirms the calldata + dataToSign pinned at `Regression.t.sol:1294-1314` reproduce exactly. The remaining whole-suite failures are HTTP-429 rate-limit on the archive endpoint under parallel test load and mainnet-archive-not-available; they are environmental and CI's archive endpoint resolves them. |
| 5 | Read `OPCMUpgradeV700.sol` (512 lines) + parent `OPCMTaskBase.sol` + diff vs baseline `OPCMUpgradeV600.sol` | See findings |
| 6 | `cast call 0x44e197058fb98fb3618453b3d90cdef6f5db8297 "version()"` (Sepolia, op-labs RPC) | `"7.1.17"` — matches strict `require` at line 190 |
| 7 | `cast call 0x44e1…8297 "opcmStandardValidator()"` (op-labs RPC) | `0x03a10AAd7683E56fF57bd787cE4b81EB363E591B` |
| 8 | Run example task simulation `just simulate` (block 10808500) | Domain `0x07e034…e61e`, Message `0x6adbe0…0d85`, calldata/data-to-sign **bit-for-bit identical** to the values pinned in `Regression.t.sol:1294-1314` |

### Blocking Issues

_None._ Template is structurally sound, well-tested, and well-documented.

### Warnings

- [ ] **`OPCMUpgradeV700.sol:188` — OPCM address loaded from TOML, not hardcoded by chain ID.**  The skill's standard checklist recommends hardcoding OPCM addresses by chain ID; this template instead reads `.addresses.OPCM` from TOML and verifies `version() == "7.1.17"` after the fact. **This matches the established pattern in `OPCMUpgradeV600.sol:68`**, so it is not a regression. The docstring at lines 60-61 explicitly says it's "designed to work with chains that are NOT in the public superchain-registry," which is incompatible with strict chain-ID hardcoding. Mitigations already present: strict version pin, `OPCM_TARGETS` allowlist enforced in `OPCMTaskBase.validate`, post-exec `STANDARD_VALIDATOR` run, and `_taskStorageWrites()` allowlist. **Decision needed:** is the extra defense of a per-chain hardcode for `chainid == 11155111 → 0x44e197…8297` (and mainnet equivalent when published) worth the loss of betanet flexibility?

- [ ] **`OPCMUpgradeV700.sol:225-230` — `_isEnabled` takes an unused `IDisputeGameFactory` parameter.**  The DGF arg is never read (the only caller at line 247 passes `IDisputeGameFactory(address(0))`). Either drop the parameter or wire it up if the intent is a hook for chain-specific policy that reads live `gameImpls`. As-is, it's dead weight that suggests an incomplete design.

- [ ] **`OPCMUpgradeV700.sol:114-132` — `_taskStorageWrites()` lists 15 entries; sep/037 simulation only writes 11.**  `EthLockboxProxy`, `PermissionedWETH`, `PermissionlessWETH`, and `ProtocolVersions` did not show diffs in the betanet simulation. The wider list is documented as deliberate to support permissionless chains. **Suggestion:** add a one-line comment on `EthLockboxProxy` and `ProtocolVersions` explaining why they are listed even when not written on every chain (only `PermissionlessWETH` has that explanation today, lines 110-113).

### Passed Checks

- [x] `forge build --deny-warnings` (0 warnings)
- [x] `forge fmt --check`
- [x] Struct ordering check (alphabetical)
- [x] **All interfaces defined inline** at end of file (lines 426-512): `ISystemConfigEOAs`, `IOPContractsManagerV700`, `IOPContractsManagerStandardValidator`, `ISuperchainConfig`, `IDisputeGameFactory`, `ISystemConfig`. Only external imports are user-defined value types (`Claim`/`GameType`), forge-std utilities, solady's `LibString`, OZ's `EnumerableSet`, and local types — no monorepo interface imports.
- [x] `_templateSetup()` calls `super._templateSetup()` first (line 155)
- [x] Strict version pin `OPCM.version().eq("7.1.17")` (line 190) — matches live OPCM `0x44e197…8297`
- [x] Non-zero assertions on prestates per chain (lines 167-170)
- [x] Validator non-zero AND has code (lines 195-196)
- [x] `vm.label()` for OPCM and Validator (lines 191, 197)
- [x] Cross-chain invariant: every chain shares the same `SuperchainConfig` (lines 176-184)
- [x] Duplicate-chain config rejection (line 166)
- [x] `OPCM_TARGETS.push(address(OPCM))` (line 189), enforced in `OPCMTaskBase.validate`
- [x] `_build()` uses `delegatecall` (lines 330, 351) — correct for `OPCMTaskBase`
- [x] Build success checked with chain-aware error message (line 353)
- [x] `_validate()` uses `validateWithOverrides` only when chain L1PAO/Challenger differ from standard (lines 381-396) — same idiom as V600
- [x] `expectedValidationErrors` asserted exactly equal to actual (line 399)
- [x] `_getCodeExceptions()` returns 4 EOAs per chain (owner, unsafeBlockSigner, batchInbox, batcherHash-derived) — each justified in docstring (lines 403-422)
- [x] OPCMv2 7-element fixed game-type order [`CANNON, PERMISSIONED_CANNON, CANNON_KONA, SUPER_CANNON, SUPER_PERMISSIONED_CANNON, SUPER_CANNON_KONA, ZK_DISPUTE_GAME`] matches the OPCM's `validGameTypes` literal (lines 278-286)
- [x] `_taskBalanceChanges()` returns empty — confirmed by sim "No ETH or ERC20 transfers"
- [x] `_setAllowedStorageAccesses()` override correctly extends per-chain `SuperchainConfig` + `ProtocolVersions` allowlist for betanets (lines 144-151) — well documented
- [x] Example task exists at `test/tasks/example/sep/037-opcm-upgrade-v700/` with `.env`, `addresses.json`, `config.toml`. Regression test wired in at `Regression.t.sol:1288`.
- [x] NatSpec is exceptionally thorough — Paul/U19 thread reference, OPCMv2 source link, semantic distinction between "starting" (a SET, not a delta), and rationale for each disabled/enabled slot

### Escalation Items (informational, not blockers)

- [x] **Calls a new OPCM function pair** (`upgradeSuperchain` + `upgrade(UpgradeInput)`) — replaces V600's single `upgrade(OpChainConfig[])`. Worth Security team confirmation that the two-step pattern is the intended one for U19.
- [x] **Unusual state change pattern**: respected-game-type rotation via `extraInstructions["overrides.cfg.startingRespectedGameType"]` happens **inside** the OPCM's `AnchorStateRegistry` re-initialisation, not via the Guardian's `setRespectedGameType`. PR description and template docstring both call this out — should be confirmed with Security/Foundation since it bypasses the Guardian path for this one rotation.

### Verification Commands

```bash
forge build --deny-warnings
forge fmt --check
./src/script/check-struct-order.sh
cast call 0x44e197058fb98fb3618453b3d90cdef6f5db8297 "version()(string)" \
  --rpc-url <sepolia-archive-rpc>
```

Wazabie:
## Task Review Findings — `test/tasks/example/sep/037-opcm-upgrade-v700/`

Generated by [`superchain-ops-task-review`](https://github.com/ethereum-optimism/op-claude-marketplace) skill (Security-team rigor checklist). Live values cross-checked against a sepolia archive endpoint.

### Context

This is an **example task** (under `test/tasks/example/`), not a production task. None of the example tasks ship with `README.md` or `VALIDATION.md` (verified across the 11 OPCM example tasks). So the absence of those files is not a finding for this PR — but it would be for any production task that copies this scaffold.

### Audit Trail

| Step | Action | Result |
|------|--------|--------|
| 1 | Read `config.toml`, `.env`, `addresses.json` | 7 inputs to verify |
| 2 | Cross-check `addresses.json` ↔ `chain.yaml` (`devnets/betanets/u19-beta/u19-beta-0/chain.yaml`) | **14/14 contract addresses MATCH** |
| 3 | Cross-check role addresses ↔ `op-deployer/intent.toml` `[chains.roles]` for chain `0x190a85e3` (= `420120035`) | **6/6 roles MATCH** (`l1ProxyAdminOwner`, `systemConfigOwner`, `unsafeBlockSigner`, `batcher`, `proposer`, `challenger`) |
| 4 | `cast call OPCM.version()` on Sepolia (op-labs RPC) | `"7.1.17"` ✓ |
| 5 | `cast call OPCM.opcmStandardValidator()` (op-labs RPC) | `0x03a10AAd7683E56fF57bd787cE4b81EB363E591B` ✓ matches comment in `config.toml:43` |
| 6 | `cast call 0xe934…40ff "nonce()"` (op-labs RPC) | `74` ✓ matches state override at `config.toml:51` |
| 7 | `cast call 0xe934…40ff "getOwners()" / "getThreshold()"` | `[0x107781Bc6FA8f66B843f4216fd6D5862D3aa4fcd]`, threshold `1` (1-of-1) ✓ matches Regression.t.sol comment |
| 8 | `cast call 0xc35d…BFBF "guardian()"` | `0xe934…40ff` (Guardian == ProxyAdminOwner on this betanet) ✓ |
| 9 | Run `just simulate` from task dir | **SUCCESS** — re-confirmed against a sepolia archive endpoint. Safe Hash `0xba70c4…26f0`, Domain `0x07e034…e61e`, Message `0x6adbe0…0d85`, calldata `0x82ad56cb…`, data-to-sign `0x190107e0…0d85` — **bit-for-bit identical** to `expectedDataToSign` / `expectedCallData` in `Regression.t.sol:1294-1314`. Solo regression test `testRegressionCallDataMatches_OPCMUpgradeV700_U19_Betanet` also **PASSES** against the same archive endpoint. |
| 10 | Inspect AnchorStateRegistry diff | Slot `0x06`: `…00000001` → `…00000008` ✓ respected game type rotated PERMISSIONED_CANNON → CANNON_KONA |
| 11 | Inspect DGF initBond diff | `0x011c37937e080000` = 0.08 ETH ✓ matches `initBond = 80000000000000000` |
| 12 | New contract `0xb3BA0bb7…357D` deployed during simulation | Matches `PermittedProxyDeployment = "DelayedWETH"` extra instruction (newly-deployed DelayedWETH proxy, allowed by OPCM) |

### Blocking Issues (for production use only)

- [ ] **`config.toml:27-28` — Both prestates are `0xdead00…` placeholder values.**  Acknowledged TODO at lines 25-26. **For any real-betanet or production U19 task, these MUST be replaced** with the U19 release prestates (cannon + cannon-kona) once published, and cross-checked against:
  - The Notion Prestate Reviews page (`notion.so/oplabs/1bdf153ee16280aeadc0c101118ea969`)
  - The published `op-program` / `kona-client` release artifacts
  - `superchain-registry/standard-prestates.toml` (per the comment)

- [ ] **`config.toml:31` — `expectedValidationErrors = "OVERRIDES-L1PAOMULTISIG,OVERRIDES-CHALLENGER,SYSCON-130,PLDG-10"`.**  The two `OVERRIDES-*` errors are expected on a betanet whose L1PAO/Challenger differ from the standard validator (handled by the `validateWithOverrides` path). The `SYSCON-130` and `PLDG-10` errors are real validator failures and were captured "live during simulation against the dev v7.1.17 OPCM." Before promoting to production: enumerate exactly what each error code means (look up in `lib/optimism/.../OPContractsManagerStandardValidator.sol`), confirm each is acceptable, and re-baseline once real prestates are wired in. The comment explicitly says "Will tighten once real prestates are wired in" — so **this is tracked, just blocking for production**.

### Input Verification Table

| Input | Config Value | Authoritative Source | Match |
|-------|--------------|---------------------|-------|
| `OPCM` | `0x44e197058fb98fb3618453b3d90cdef6f5db8297` | Live `cast` on Sepolia (`version() = "7.1.17"`) | YES |
| `chainId` | `420120035` | `devnets/.../u19-beta-0/chain.yaml:chain_id` | YES |
| `cannonPrestate` | `0xdead000…` | Notion Prestate Reviews | **PLACEHOLDER (TODO)** |
| `cannonKonaPrestate` | `0xdead000…` | Notion Prestate Reviews | **PLACEHOLDER (TODO)** |
| `expectedValidationErrors` | `OVERRIDES-L1PAOMULTISIG,OVERRIDES-CHALLENGER,SYSCON-130,PLDG-10` | Captured live; not yet documented | **NEEDS DOCUMENTATION before production** |
| `initBond` | `80000000000000000` (0.08 ETH) | PR description; confirmed in DGF state diff | YES |
| `startingRespectedGameType` | `8` (CANNON_KONA) | U19 spec; corresponds to always-enabled slot 2 in `_isEnabled` | YES |
| Safe nonce override | slot `0x05` of `0xe934…40ff` = `74` | Live `cast` on Sepolia | YES |
| All 14 contracts in `addresses.json` | per-chain | `chain.yaml` | YES (14/14) |
| All 6 roles in `addresses.json` | per-chain | `intent.toml` `[chains.roles]` | YES (6/6) |

### Passed Checks

- [x] **Simulation completes successfully** — domain hash, message hash, calldata, and data-to-sign all reproduce exactly the values pinned in `Regression.t.sol:1294-1314`
- [x] `templateName = "OPCMUpgradeV700"` — matches the template file name
- [x] `fallbackAddressesJsonPath` set (required for chains not in `superchain-registry`)
- [x] **State override on Safe nonce (74) matches on-chain `nonce()` query** — verified live
- [x] Gas limit 15M (`TENDERLY_GAS=15000000`) — at the Fusaka cap. Acceptable for a single-chain task; flag if a future production U19 task batches multiple chains.
- [x] `FORK_BLOCK_NUMBER = 10808500` present in `.env` and referenced in regression test
- [x] State diff matches U19 spec: respected game type rotated 1→8 in `AnchorStateRegistry`, `CANNON_KONA` `initBond = 0.08 ETH` in DGF, `DelayedWETH` proxy newly deployed, DGF `gameImpls` written for slot 1 (`PERMISSIONED_CANNON`) and slot 8 (`CANNON_KONA`), all impl slots upgraded to v7.1.17 impls, `SuperchainConfig` re-initialised with new impl
- [x] **All 14 contract addresses in `addresses.json` match `https://github.com/ethereum-optimism/devnets/blob/main/betanets/u19-beta/u19-beta-0/chain.yaml`**
- [x] **All 6 role addresses match `https://github.com/ethereum-optimism/devnets/blob/main/betanets/u19-beta/op-deployer/intent.toml` `[chains.roles]` for chain id `0x190a85e3`**
- [x] **OPCM `0x44e197058fb98fb3618453b3d90cdef6f5db8297` returns `version() = "7.1.17"` on Sepolia (live)**
- [x] **OPCM `opcmStandardValidator()` returns `0x03a10AAd7683E56fF57bd787cE4b81EB363E591B` on Sepolia (live)**, matching the comment at `config.toml:43`
- [x] `[stateOverrides]` slot `5` (Gnosis Safe nonce) = `74` matches live on-chain nonce — pinning correct as of capture block
- [x] `startingRespectedGameType = 8` (CANNON_KONA) corresponds to the always-enabled slot 2 in `_isEnabled` — won't trigger `OPContractsManagerV2_InvalidGameConfigs`
- [x] `chainId = 420120035` matches `chain_id` in `chain.yaml` and decimal of `0x190a85e3` in `intent.toml`
- [x] `initBond = 80000000000000000` (0.08 ETH) confirmed in DGF state diff at storage slot `0x4702e508…87d5`

### User Verification Required (before approval)

- [ ] **You verify Tenderly** state diff against the simulation output. I confirmed the structural correctness (10 contracts modified + Safe nonce++, 1 newly deployed `DelayedWETH`), the AnchorStateRegistry game-type rotation (1→8), and the initBond (0.08 ETH). Open the Tenderly link from `just simulate` to verify SystemConfig's two new fields (`delayedWETH` and `opcm`) are wired correctly.
- [ ] **Confirm with Security team** that the OPCM-from-TOML pattern (vs hardcode-by-chain-ID) is acceptable for U19 production tasks given the stated betanet-compatibility goal. (Same as the matching item in the template review comment.)
- [ ] **Confirm with Foundation/Security** that rotating the respected game type via OPCM `extraInstructions` (not Guardian's `setRespectedGameType`) is the intended path for U19.

### Verification Commands

```bash
SIMULATE_WITHOUT_LEDGER=1 OVERRIDE_RPC_URL=<sepolia-archive-rpc> \
  bash -c 'cd test/tasks/example/sep/037-opcm-upgrade-v700 && \
    just --dotenv-path "$(pwd)/.env" --justfile ../../../../../src/justfile simulate'

cast call 0x44e197058fb98fb3618453b3d90cdef6f5db8297 "version()(string)" \
  --rpc-url <sepolia-archive-rpc>
cast call 0xe934Dc97E347C6aCef74364B50125bb8689c40ff "nonce()(uint256)" \
  --rpc-url <sepolia-archive-rpc>
```

> **Note on archives:** the historical fork simulation and the V700 regression test require a sepolia archive endpoint that serves block 10808500. Reproduced bit-identical hashes matching `Regression.t.sol` against the archive used here. No new findings from the re-run.

JosepBove (dismissed):

snevins-agent-2 (dismissed):
## Review: U19 CANNON / CANNON_KONA follow-up

Thanks for the thorough template and self-review. I checked the specific follow-up item Steven had queued: **CANNON must be disabled/enforced off, and permissionless chains must use CANNON_KONA**.

### What looks good

- `CANNON` is explicitly included in the 7-entry OPCMv2 game config array and disabled by `_isEnabled(...)` returning `false` for everything except `PERMISSIONED_CANNON` and `CANNON_KONA`.
- Disabled configs are emitted with `enabled=false`, `initBond=0`, and empty `gameArgs`, so slot 0 should clear via OPCM rather than being omitted.
- `CANNON_KONA` is always enabled and uses `cannonKonaPrestate`.
- The example config sets `startingRespectedGameType = 8`, and CI is green.

### Request changes

The template does **not enforce** that the respected game type is `CANNON_KONA`.

`startingRespectedGameType` is TOML-configurable and is passed directly through in `_extraInstructions(...)`. `_templateSetup(...)` validates config length/duplicates/prestates, but it does not require `parsed[i].startingRespectedGameType == CANNON_KONA`. Since `PERMISSIONED_CANNON` is also enabled, a future task could set `startingRespectedGameType = 1`; the OPCM enabled-slot validation would likely still pass, but the chain would not rotate to Kona.

Please add a setup-time invariant, e.g.:

```solidity
require(
    parsed[i].startingRespectedGameType == CANNON_KONA,
    "OPCMUpgradeV700: startingRespectedGameType must be CANNON_KONA"
);
```

Alternatively, remove the TOML field and encode `CANNON_KONA` directly in `_extraInstructions(...)`.

### Non-blocking cleanup

The top-level comment says “currently-registered pre-superroot game types are kept enabled,” but the implementation intentionally disables `CANNON` and all super/ZK slots, enabling only `PERMISSIONED_CANNON` and `CANNON_KONA`. Please update that comment so operators do not infer that live permissionless `CANNON` is preserved.

I could not run Forge locally in my Hermes environment (`forge` is not installed), but GitHub checks show `forge_build`, `forge_test`, `template_regression_tests`, and stacked simulations all passing.

src/template/OPCMUpgradeV700.sol
```diff
@@ -0,0 +1,512 @@
...
+    function _extraInstructions(uint256 chainId)
+        internal
+        view
+        returns (IOPContractsManagerV700.ExtraInstruction[] memory ix)
+    {
+        ix = new IOPContractsManagerV700.ExtraInstruction[](2);
+        ix[0] = IOPContractsManagerV700.ExtraInstruction({key: "PermittedProxyDeployment", data: bytes("DelayedWETH")});
+        ix[1] = IOPContractsManagerV700.ExtraInstruction({
+            key: "overrides.cfg.startingRespectedGameType",
+            data: abi.encode(upgrades[chainId].startingRespectedGameType)
```
stevennevins: Can we add a require on the value from the config override for startingRespectedGameType can't be CANNON, just to be extra certain since it's an important requirement

stevennevins (dismissed):

stevennevins (approved):

JosepBove (approved):

PR #1410, opened 2026-04-29
feat(v800): add ZK_DISPUTE_GAME slot to OPCM upgrade

### LINKED ISSUE #20471: fix(superchain-ops): update the super root upgrade script to use latest OPCM
OPCM now has the zk dispute game

An array of all valid game types must be provided as input

This PR updates the task template to now provide the 7 games (before 6)

And updates the OPCM we point to in the test task to the latest OPCM deployed for the devnet testing

stevennevins:
## Summary

Rename the provisional OPCM upgrade template and example fixture to V800, and update it for the new devnet OPCM at `0x15c65ee282777e4a2856fdf26d33ee4d5bab30c4`. The updated template builds a 7-entry `disputeGameConfigs` array including `ZK_DISPUTE_GAME` (raw value 10), checks the ZK game implementation slot when deciding whether to enable it, and drops the now-fixed `SCDG-SHAPE` validation error from `expectedValidationErrors`.

---

## Additional context

The target protocol upgrade version has not been formally defined yet, so the V800 name is provisional and the OPCM version assertion is intentionally commented out. `SuperRootUpgradeTest` (`test_load_data` and `test_upgrade_sepolia`) passes against a Sepolia fork.

fixes https://github.com/ethereum-optimism/optimism/issues/20471


src/template/OPCMUpgradeV700.sol
```diff
@@ -171,6 +172,7 @@ contract OPCMUpgradeV700 is OPCMTaskBase {
         if (gt == SUPER_CANNON_KONA) {
             return address(disputeGameFactory.gameImpls(GameType.wrap(CANNON_KONA))) != address(0);
         }
+        if (gt == ZK_DISPUTE_GAME) return false;
```
ajsutton: I know this is strictly never deployed now but shouldn't we actually check the gameImpls anyway? That way when this is inevitably copied to create the template for V8 where ZK is deployed somewhere it will just work instead of being a footgun.

ajsutton (dismissed):
LGTM - just the one note about future proofing.

src/template/OPCMUpgradeV700.sol
stevennevins: [a2dd901](https://github.com/ethereum-optimism/superchain-ops/pull/1410/commits/a2dd9011ad236976cc25f0f15abe13b454855819)

stevennevins:
Note on the V700/V800 naming: this PR is for a version that has not been formally defined yet, so the exact template/version name is provisional. We can rename it once the target version is finalized.

Inphi:
@stevennevins why the rename? What would be the tentative version of opcm were we to cut an rc today?

stevennevins:
> @stevennevins why the rename? What would be the tentative version of opcm were we to cut an rc today?

Basically, to avoid confusion with U19 which would be the V700 template

stevennevins:
> > @stevennevins why the rename? What would be the tentative version of opcm were we to cut an rc today?
> 
> Basically, to avoid confusion with U19 which would be the V700 template

Adrian commented on a slack thread for it and decided to include in this PR since it was small

Inphi (dismissed):

src/template/OPCMUpgradeV800.sol
```diff
@@ -144,10 +146,11 @@ contract OPCMUpgradeV700 is OPCMTaskBase {
...
         vm.etch(address(0x0002b8639730E2F4dc88Dfd5Bbd0352E5518A758), hex"01");
 
-        // OPCM from TOML; must be v7.1.15
-        opcm = IOPContractsManagerV700(tomlContent.readAddress(".addresses.OPCM"));
+        // OPCM from TOML; must be v7.1.17
+        opcm = IOPContractsManagerV800(tomlContent.readAddress(".addresses.OPCM"));
         OPCM_TARGETS.push(address(opcm));
-        require(opcm.version().eq("7.1.15"), "Incorrect OPCM");
+        // Version is intentionally not asserted while the target protocol upgrade version is still undefined.
+        // require(opcm.version().eq("7.1.17"), "Incorrect OPCM");
```
Inphi: Consider adding a `skipOPCMVersion` check option so we don't forget to take care of this.
On another note: Given that we're strict about bumping the semver major/minor version of contracts whenever there is an ABI breakage or a new release, it could suffice to check the major/minor semver part.

stevennevins: good call, added it as an optional config.toml key
[dc58b99](https://github.com/ethereum-optimism/superchain-ops/pull/1410/commits/dc58b996c97596ab3783bfa426718b890ba92381)

Inphi (approved):

JosepBove (approved):

PR #1415, opened 2026-05-05
feat(devnet-gen): add task generator for devnet OPCM upgrades

Inphi:
## Summary

Adds `just gen-devnet-task` — a Python tool under `src/script/devnet-gen/` that generates devnet ops task directories from existing Solidity templates plus a path to a devnet in the [devnets](https://github.com/ethereum-optimism/devnets) repo. Two adapters ship: `OPCMUpgradeV600` and `OPCMUpgradeV800`.

## Why

Almost every input a devnet upgrade task needs is already in the devnets repo — per-chain L1 addresses in `chain.yaml`, prestates in `op-program/prestates.json`, OPCM in `op-deployer/state.json`. Hand-copying these into a fresh `src/tasks/sep/` task is rote and error-prone; this automates it.

## Usage

```bash
just gen-devnet-task list
just gen-devnet-task info OPCMUpgradeV800
just gen-devnet-task OPCMUpgradeV600 ../devnets/alphanets/u18-alpha
just gen-devnet-task OPCMUpgradeV800 ../devnets/alphanets/u18-alpha \
    --override initBond=0 --override startingRespectedGameType=1
```


stevennevins:
just curious what your thoughts are for having this be here vs on the netchef side?

Inphi:
> just curious what your thoughts are for having this be here vs on the netchef side?

@stevennevins It could go either place. Though I don't see any advantage to putting this in netchef as it would only split focus on where the adapters should live. I'd prefer to establish a pattern where adding a new superchain-ops template also prompts the developer to consider a corresponding devnet-gen adapter for that task. That way, both the template and the devnet-gen changes can be handled in a single PR.

PR #1416, opened 2026-05-05
fix(mise): Replace deprecated alias section

Inphi:
Working on superchain-ops with a recent version of mise results in the following warning in the shell prompt:
```
 mise WARN  deprecated [alias]: [alias] is deprecated, use [tool_alias] instead in ~/optimism/superchain-ops/mise.toml      10:49:13
mise WARN  deprecated [alias]: [alias] is deprecated, use [tool_alias] instead in ~/optimism/superchain-ops/mise.toml
```

This change replaces the deprecated `[alias]` syntax with `[tool_alias]` to silence the warning.
Users on older versions of mise may need to update, but since tool_alias was introduced nearly a year ago, breakage is unlikely.

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -18,7 +18,7 @@ forge = "v1.1.0" # https://github.com/foundry-rs/foundry/releases/tag/v1.1.0
 cast = "v1.1.0"
 anvil = "v1.1.0"
 
-[alias]
+[tool_alias]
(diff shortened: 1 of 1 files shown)


stevennevins (approved):

mds1:
Closing as superseded by #1456, which migrated mise aliases to `tool_alias`.

PR #1428, opened 2026-05-12
Add interop-jnt-v2 V800 upgrade and migration tasks

stevennevins:
## Summary

- Add Sepolia task 076 for the interop-jnt-v2 OPCM V800 upgrade to SUPER_CANNON_KONA.
- Add Sepolia task 077 for the interop-jnt-v2 OPCM V800 migration to a shared SUPER_CANNON_KONA game for both devnet chains.
- Include fallback devnet address artifacts for chains 420120102 and 420120103 because these chains are not in the public superchain registry.
- Mark both tasks EXECUTED.

## Validation

- Compared every fallback address value against ethereum-optimism/devnets PR #323 at 783214f937aef1f1a3ef7e63f5e1459d58f82153, using cast to-check-sum-address for EIP-55 normalization.
- Verified every address string in both addresses.json files is EIP-55 checksummed.
- Verified both addresses.json files are valid JSON and byte-identical.
- Validated the migration config structure against test/tasks/example/sep/036-opcm-migrate-v800/config.toml.
- Ran git diff --check.

## Notes

- VALIDATION.md files updated with simulated domain/message hashes and verified against signed data


Inphi (commented):
We probably don't need to commit this task here if the sepolia L1PAO used for the interop-jnt-v2 devnet is different from the production one. the value of keeping tasks in superchain-ops is to log and coordinate signing operations. For devnets, this is less useful to do.

Happy to stamp this change if still needed. But fyi for next time.

stevennevins:
> We probably don't need to commit this task here if the sepolia L1PAO used for the interop-jnt-v2 devnet is different from the production one. the value of keeping tasks in superchain-ops is to log and coordinate signing operations. For devnets, this is less useful to do.
> 
> Happy to stamp this change if still needed. But fyi for next time.

I'm okay with closing it

PR #1439, opened 2026-05-29
fix(OPCMUpgradeV700): support standard chains in U19 upgrade template

sbvegan:
## Summary

Four fixes discovered while simulating U19 upgrades on standard superchain-registry chains for the first time (all prior `OPCMUpgradeV700` usage was betanet-only):

- **`SuperchainAddressRegistry`**: Register `EthLockboxProxy` per-chain from the registry JSON in `_fetchAndSaveInitialContracts`. The standard on-chain discovery path didn't include it, so it was missing from allowed storage accesses and the state-diff validator rejected valid OPCM upgrade writes.

- **`OPCMUpgradeV700._setAllowedStorageAccesses`**: Wrap `getAddress("ProtocolVersions", chainId)` in try/catch. Standard chains store `ProtocolVersions` on the sentinel chain (via `addresses.toml`), not per L2 chainId, so the per-chain lookup reverted unconditionally for non-betanet tasks.

- **`OPCMUpgradeV700._getCodeExceptions`**: Only include true EOAs (no code). Some chains (e.g. Ink) use a multisig Safe as `SystemConfig.owner()`. Adding a contract address to the EOA exceptions list caused `"unexpected code"` validation failures when the OPCM re-initializes `SystemConfig` and writes the owner to storage.

- **`OPCMUpgradeV700._build`**: Skip `upgradeSuperchain()` if `SuperchainConfig` is already at the v7.1.17 implementation. Chains with non-standard L1PAOs (e.g. Unichain) cannot authorize `ProxyAdmin.upgrade()` on the shared `SuperchainConfig`. In the stacked execution model a prior shared-L1PAO task runs `upgradeSuperchain` first, so the Unichain task can skip it via this idempotency check.

## Test plan

- [ ] Existing betanet U19 tasks (076, 079) still simulate correctly
- [ ] `just simulate-stack sep 086-U19-op` passes (once sep task PR is opened)
- [ ] `just simulate-stack eth 053-U19-op` passes (once mainnet task PR is opened)

### DIFF
--- a/src/SuperchainAddressRegistry.sol
+++ b/src/SuperchainAddressRegistry.sol
@@ -344,11 +344,9 @@ contract SuperchainAddressRegistry is StdChains {
 
         // EthLockboxProxy cannot be discovered on-chain via a standard getter; read from
         // the registry JSON when present (introduced in op-contracts v7.x).
-        string memory ethLockboxKey =
-            string.concat("$.", vm.toString(chain.chainId), ".EthLockboxProxy");
+        string memory ethLockboxKey = string.concat("$.", vm.toString(chain.chainId), ".EthLockboxProxy");
         if (vm.keyExistsJson(chainAddressesContent, ethLockboxKey)) {
-            address ethLockbox =
-                parseContractAddress(chainAddressesContent, chain.chainId, "EthLockboxProxy");
+            address ethLockbox = parseContractAddress(chainAddressesContent, chain.chainId, "EthLockboxProxy");
             if (ethLockbox != address(0)) saveAddress("EthLockboxProxy", chain, ethLockbox);
         }
     }
--- a/src/SuperchainAddressRegistry.sol
+++ b/src/SuperchainAddressRegistry.sol
@@ -341,14 +341,6 @@ contract SuperchainAddressRegistry is StdChains {
 
         optimismPortalProxy = getOptimismPortalProxy(l1CrossDomainMessengerProxy);
         saveAddress("OptimismPortalProxy", chain, optimismPortalProxy);
-
-        // EthLockboxProxy cannot be discovered on-chain via a standard getter; read from
-        // the registry JSON when present (introduced in op-contracts v7.x).
-        string memory ethLockboxKey = string.concat("$.", vm.toString(chain.chainId), ".EthLockboxProxy");
-        if (vm.keyExistsJson(chainAddressesContent, ethLockboxKey)) {
-            address ethLockbox = parseContractAddress(chainAddressesContent, chain.chainId, "EthLockboxProxy");
-            if (ethLockbox != address(0)) saveAddress("EthLockboxProxy", chain, ethLockbox);
-        }
     }
 
     function _saveProxyAdminEntries(ChainInfo memory chain, address systemConfigProxy) internal {
--- a/src/SuperchainAddressRegistry.sol
+++ b/src/SuperchainAddressRegistry.sol
@@ -341,6 +341,16 @@ contract SuperchainAddressRegistry is StdChains {
 
         optimismPortalProxy = getOptimismPortalProxy(l1CrossDomainMessengerProxy);
         saveAddress("OptimismPortalProxy", chain, optimismPortalProxy);
+
+        // EthLockboxProxy cannot be discovered on-chain via a standard getter; read from
+        // the registry JSON when present (introduced in op-contracts v7.x).
+        string memory ethLockboxKey =
+            string.concat("$.", vm.toString(chain.chainId), ".EthLockboxProxy");
+        if (vm.keyExistsJson(chainAddressesContent, ethLockboxKey)) {
+            address ethLockbox =
+                parseContractAddress(chainAddressesContent, chain.chainId, "EthLockboxProxy");
+            if (ethLockbox != address(0)) saveAddress("EthLockboxProxy", chain, ethLockbox);
+        }
     }
 
     function _saveProxyAdminEntries(ChainInfo memory chain, address systemConfigProxy) internal {
--- a/src/template/OPCMUpgradeV700.sol
+++ b/src/template/OPCMUpgradeV700.sol
@@ -345,6 +345,7 @@ contract OPCMUpgradeV700 is OPCMTaskBase {
         SuperchainAddressRegistry.ChainInfo[] memory chains = superchainAddrRegistry.getChains();
         address sharedSC = superchainAddrRegistry.getAddress("SuperchainConfig", chains[0].chainId);
 
+
         // ERC-1967 implementation slot: keccak256("eip1967.proxy.implementation") - 1
         bytes32 ERC1967_IMPL_SLOT = 0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc;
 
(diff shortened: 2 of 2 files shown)


Wazabie (commented):
## Template Review Findings — `OPCMUpgradeV700`

**Scope:** four template fixes for U19 standard-chain support. Target version: `op-contracts/v7.1.17` (OPCMv2). Template type: `OPCMTaskBase` (delegatecall via Multicall3DelegateCall).

> _**Updated 2026-05-29 (post-`edf7775d`):** The `SuperchainAddressRegistry` `EthLockboxProxy` block has been removed from this PR — it will land alongside the standard-chain task PR (#1440) where it can be exercised end-to-end. This PR is now purely template-level. All CI gates pass._

### Audit Trail

| Step | Command/Action | Result |
|------|----------------|--------|
| 1 | `forge build --deny-warnings` | PASS (0 warnings) |
| 3 | `./src/script/check-struct-order.sh` | PASS (OPCMUpgrade fields alphabetical) |
| 4 | `forge test --skip Integration` | 219 pass, 40 fail — all 40 are RPC historical-state-unavailable, unrelated to template |
| 5 | Read `src/template/OPCMUpgradeV700.sol` | Reviewed against checklist A–J |
| 6 | `just simulate` on `test/tasks/example/sep/037-opcm-upgrade-v700` | Local sim still reverts on stale-EthLockbox in `addresses.json` — pre-existing, unrelated to this PR. CI `template_regression_tests` passes. |

### Blocking Issues

**0 blocking issues.** All four template fixes are correct.

### Warnings (2)

1. **`src/template/OPCMUpgradeV700.sol:359`** — Staticcall success check regressed during rebase.
   Commit 59be4fd7 ("address Graphite review") added `require(implCallSuccess && implData.length >= 32, "OPCMUpgradeV700: implementations() call failed")`. Subsequent graphite-app commits (366e4372, a2b7b9e8) stripped it. Current code:
   ```solidity
   (, bytes memory implData) = address(OPCM).staticcall(abi.encodeWithSignature("implementations()"));
   address targetSCImpl = abi.decode(implData, (address));
   ```
   The version check in `_templateSetup` (`OPCM.version().eq("7.1.17")`) makes this safe in practice, but the prior commit's stated rationale ("unhelpful abi panic instead of a clear error") still applies if a misconfigured OPCM ever slips past the version assertion. Re-instate the success+length guard.

2. **`test/tasks/example/sep/037-opcm-upgrade-v700`** — Local sim reverts post-build (CI passes).
   With `DISABLE_OVERRIDE_NONCE_CHECK=1 SIMULATE_WITHOUT_LEDGER=1`, the storage-access check reverts on `0xb3BA0bb7f117Fc89Db8eaB51554389FB8191357D` — the EthLockboxProxy in use on-chain at fork block 10808500. The example's `addresses.json` lists `EthLockboxProxy: 0x4a54316b...`, which is what ends up in `_allowedStorageAccesses`; the real one is not allowlisted. Both addresses have code on Sepolia, so the JSON entry is stale rather than fictional.
   Not introduced by this PR — the betanet addresses.json drifted from on-chain state. CI's `template_regression_tests` still passes (different env / different code path), so it doesn't block merge. Worth a follow-up to refresh the example's JSON.

### Passed Checks

- **A. Interface decoupling**: all interfaces defined inline (`IOPContractsManagerV700`, `IOPContractsManagerStandardValidator`, `ISuperchainConfig`, `IDisputeGameFactory`, `ISystemConfig`, `ISystemConfigEOAs`); no monorepo interface imports.
- **B. OPCM version pin**: `require(OPCM.version().eq("7.1.17"))` at `OPCMUpgradeV700.sol:200`. OPCM address read from TOML — pre-existing pattern; not introduced by this PR.
- **C. Struct order**: `OPCMUpgrade` fields alphabetical (`check-struct-order.sh` PASS).
- **D. `_templateSetup`**: calls `super._templateSetup` first; version pin; non-zero + has-code on `sharedSC` and `STANDARD_VALIDATOR`; `vm.label` on OPCM and validator.
- **E. `_build`**: `delegatecall` (correct for `OPCMTaskBase`); both delegatecalls have `require(ok, "...")` with meaningful messages; new skip path also logs via `console.log` for audit trail.
- **F. `_validate`**: uses `validateWithOverrides` only when overrides differ from standard; iterates all chains; compares expected vs actual errors.
- **G. `_taskStorageWrites`**: includes `EthLockboxProxy` (will be allowlisted once the registry-side change lands in the standard-chain PR). `PermissionlessWETH` warning is intentional and noted in the doc comment.
- **H. `_getCodeExceptions`**: fix filters to true EOAs only; pure-Solidity copy (assembly trim removed per 9a4dc76b).
- **I. Example task**: `config.toml`, `addresses.json`, `.env` (with `FORK_BLOCK_NUMBER=10808500`) all present.
- **J. Docs/style**: NatSpec describes target version, behavior, and OPCMv2 mechanics; `forge build --deny-warnings` and `forge fmt --check` both clean.

### Per-Fix Review

**Fix 1 — `_setAllowedStorageAccesses` try/catch on `ProtocolVersions`** (`src/template/OPCMUpgradeV700.sol:156-165`)
- Per-chain lookup for betanets; `get("ProtocolVersions")` (sentinel) fallback for standard chains. `EnumerableSet.add` is a no-op on duplicates, so re-adding the sentinel across iterations is fine.
- Catch block emits `console.log` naming the chainId before taking the fallback — operator-visible diagnostic for the standard-chain path.
- ✅ Correct.

**Fix 2 — `_getCodeExceptions` filters true EOAs only** (`src/template/OPCMUpgradeV700.sol:459-472`)
- `candidates[j].code.length == 0` filter correctly drops contract addresses (Ink's Safe `SystemConfig.owner()`). Those are then handled by the normal allowed-storage-access check.
- ✅ Correct.

**Fix 3 — Skip `upgradeSuperchain` if already at target impl** (`src/template/OPCMUpgradeV700.sol:348-382`)
- Reads ERC-1967 impl slot from live proxy via `vm.load`; reads target via `OPCM.implementations()` decoded as `address` (first field).
- Verified: `Implementations.superchainConfigImpl` is the first struct field at `OPContractsManager.sol:104-117`. Decoding only the head address is safe.
- Re-simulation confirms the skip path fires correctly on the example betanet: `OPCMUpgradeV700: skipping upgradeSuperchain - SuperchainConfig proxy already points at target implementation 0xE4F9779ab53070a55db24dFAeFf9AF147c6ED550`.
- ✅ Correct (subject to Warning 1).

**Fix 4 — `EthLockboxProxy` registration** — _**moved to #1440**_ where it can be exercised end-to-end against the actual standard-chain tasks. The replacement implementation will use `try IFetcher(optimismPortalProxy).ethLockbox()` instead of reading from JSON (cleaner, matches the rest of the file's on-chain-discovery pattern, no stale-JSON risk).

### Phase 5b — Storage Writes Cross-Reference

Deferred to PR #1440 — that's where standard-chain tasks will produce real state-diffs to cross-reference against `_taskStorageWrites`.

### Escalation Checklist

- [ ] New template type — **No** (V700 variation)
- [ ] New OPCM function — **No** (`upgradeSuperchain`, `upgrade` already used)
- [ ] Cross-layer operations — **No**
- [ ] Security-critical operations — **No directly** (idempotency skip is a new pattern but justified by stacked-execution model)
- [ ] Unusual state changes — **No**
- [ ] Standard Validator failures — **N/A** (sim didn't reach validator)
- [ ] Complex Safe operations — **No**
- [ ] Unclear specification — **No**

No escalation triggers met.

### Recommendation

Approve. Follow-ups:
1. **Restore the staticcall success+length guard** at `OPCMUpgradeV700.sol:359` (lost during rebase from 59be4fd7) — can be done in this PR or a follow-up.
2. **Refresh `addresses.json`** in `test/tasks/example/sep/037-opcm-upgrade-v700` so the example task simulates end-to-end locally (CI is fine).
3. **`EthLockboxProxy` registry discovery** lands in #1440 alongside the first standard-chain task that exercises it.

The four template fixes are correct and CI-green.

---

🤖 Generated via `superchain-ops-template-review` skill.

src/SuperchainAddressRegistry.sol
```diff
@@ -341,6 +341,14 @@ contract SuperchainAddressRegistry is StdChains {
 
         optimismPortalProxy = getOptimismPortalProxy(l1CrossDomainMessengerProxy);
         saveAddress("OptimismPortalProxy", chain, optimismPortalProxy);
+
+        // EthLockboxProxy cannot be discovered on-chain via a standard getter; read from
```
smartcontracts: It should be able to be accessed via Portal?

smartcontracts (dismissed):

src/SuperchainAddressRegistry.sol
Wazabie: Right, I'll remove this change from here and add it back corrected in this [U19 sepolia tasks PR](https://github.com/ethereum-optimism/superchain-ops/pull/1440/changes#diff-ca6dec9d63928949acb60295766bd1138317fe087bb8f06ed0832477d331c636)

JosepBove (approved):

PR #1440, opened 2026-05-29
U19 upgrade tasks for Sepolia

sbvegan:
## Summary

Upgrades the public-registry Sepolia chains to **Upgrade 19** (`op-contracts/v7.1.17`)
via `OPCM.upgradeSuperchain` + `OPCM.upgrade`, using the `OPCMUpgradeV700` template
(merged separately in #1439). Two tasks:

| Task | Scope | ProxyAdminOwner |
|------|-------|-----------------|
| `099-U19-op-ink-mmz-soneium` | 6 OP-governed chains, bundled into **one** OPCM transaction | shared Sepolia L1PAO `0x1Eb2fFc9…56E2` |
| `100-U19-unichain` | Unichain Sepolia (chainId `1301`) | independent Unichain PAO `0xd363339e…9dbf` |

### Task 099 — bundled OP-governed upgrade

OP (`11155420`), Ink (`763373`), Metal (`1740`), Mode (`919`), Zora (`999999999`),
and Soneium Minato (`1946`) share the standard Sepolia L1PAO / Security Council /
Foundation Upgrade Safe, so they are bundled into a single `OPCM` call — **one nonce
consumed per safe** (not one per chain).

Post-upgrade respected-game-type split (per the Sepolia U19 thread):

- **Permissionless** → rotate `respectedGameType` to `CANNON_KONA` (8): **OP, Ink, Soneium**.
- **Permissioned** → keep `PERMISSIONED_CANNON` (1), CANNON_KONA left unwired: **Metal, Mode, Zora**.

U19 installs `CANNON_KONA` (game type 8) as the primary FPVM and disables the legacy
`CANNON` slot (`gameImpls[0] = address(0)`).

### Task 100 — Unichain

Unichain has an independent ProxyAdminOwner, so it gets its own task. **It depends on
099**: 099's `upgradeSuperchain` updates the shared `SuperchainConfig`, which Unichain's
`SystemConfig` references, so 099 must execute first. (The stacked simulator runs 099→100
in order, so `just simulate-stack sep 100-U19-unichain` already exercises this dependency.)

## Prerequisites already landed on `main`

These set the stage for 099/100:

- **#1439** — the `OPCMUpgradeV700` template.
- `086`–`096` — U19 rehearsal/corrective tasks on the OP Labs **devnets** (`karst-u19-beta-0`,
  `karst-u19-alpha-0`), which validated the upgrade flow before touching public chains.
- `097`–`098` — Osaki Sepolia L1/L2 ownership transfers. These have **already executed**,
  each consuming one nonce on the three shared Sepolia safes, so 099's `stateOverrides`
  equal the current live nonces (L1PAO 52 / Council 66 / Foundation 72; Unichain PAO 43).

## Prestate

Uses the real, reproducibly-built kona-client **v1.6.0-rc.2** `cannon64-kona` absolute
prestate `0x0337ecb3604c0b40c352e0c7711beb17a212d583f4fe956fd8d66e29ad5f9025`
(commit `d7cea91`), published in the superchain-registry
([standard-prestates.toml — `1.6.0-rc.2` / `cannon64-kona`](https://github.com/ethereum-optimism/superchain-registry/blob/ac4e48516794322e20abfe4f4c175e379b9a5d19/validation/standard/standard-prestates.toml#L460-L462))
and reviewed in the [U19 Kona prestate review](https://app.notion.com/p/oplabs/U19-Kona-37bf153ee1628067abf5ec75b5fc0609).
CANNON type-0 prestates are intentionally `0xdead…` placeholders since CANNON is disabled in U19.

## Verification

- [x] All config inputs verified against superchain-registry (OPCM `0x44e1…8297` = v7.1.17;
      6 chain IDs + PAOs; Unichain SystemConfig; kona prestate published in standard-prestates.toml).
- [x] Nonce overrides match live on-chain values.
- [x] Stacked simulations pass for both tasks; VALIDATION.md domain/message hashes reproduce
      and match (099 Council + Foundation paths, 100 Unichain).
- [x] Ordered dependency confirmed on a Tenderly Virtual TestNet: 099 upgrades SuperchainConfig
      (2.4.0 → 2.4.2), after which 100 succeeds; gas 099 ≈ 12.4M / 100 ≈ 2.3M, both under the
      16,777,216 (2²⁴) per-tx cap.


Wazabie:
Clarify ethLockbox address fetch in SuperchainAddressRegistry. (The negative-gas-refund MultisigTask change referenced earlier has been reverted out of this PR.)

Wazabie (commented):
## Template Review (per `superchain-ops-template-review` skill)

### Scope

This PR does **not modify any files under `src/template/`**. The U19 template (`OPCMUpgradeV700.sol`) was merged separately via #1439 and is already in `main`. The only framework change in this PR is `SuperchainAddressRegistry.sol`, reviewed in a dedicated comment below. (The earlier `MultisigTask.sol` fixes have since been reverted out of this PR and will land separately.)

The skill's verification commands still apply to the PR's diff — results below.

### Verification Commands

| # | Command | Result |
|---|---------|--------|
| 1 | `forge build --deny-warnings` | PASS — 0 warnings, 0 errors |
| 3 | `./src/script/check-struct-order.sh` | PASS — 18 TOML/JSON-parsed structs all alphabetically ordered |
| 4 | `forge test --match-path "test/registry/*"` | 54 passed / 8 failed — every failure is `historical state … is not available` from pruned RPC archive, not a code regression. New `testEthLockboxDiscoveredViaPortal` passes on both Mainnet and Sepolia forks. |

### Template Code Review

The skill's Phase 4 checklist (interface decoupling, OPCM hardcoding, struct ordering, `_templateSetup()`, etc.) targets template files. Since none are modified in this PR, those items are **N/A here** and should be considered closed by #1439's review.

For context, the consuming template `src/template/OPCMUpgradeV700.sol`:
- Resolves OPCM via TOML `[addresses].OPCM` rather than chain-id hardcoding. This is intentional — the contract NatSpec notes the template is "Designed to work with chains that are NOT in the public superchain-registry" and accepts a `fallbackAddressesJsonPath`. Departs from the skill's default expectation; worth confirming this design choice was reviewed in #1439.
- Pins version via `OPCM.version().eq("7.1.17")` in `_templateSetup` (`OPCMUpgradeV700.sol:205`).
- `_taskStorageWrites()` declares `EthLockboxProxy` at line 131 — this is what motivates the `SuperchainAddressRegistry` change in this PR; see the dedicated comment below.
- Comment at `OPCMUpgradeV700.sol:232-233` cites an individual as the source for a design choice (`"the explicit ask from Paul on the U19 thread"`). Pre-existing from #1439; prefer linking to the PR/thread rather than naming an individual. Out of scope for merging this PR.

### Escalation Checklist

- [ ] New template type — N/A (template already in `main`)
- [ ] New OPCM function — N/A
- [ ] Cross-layer operations — N/A
- [ ] Unusual state changes — see task-review comment; standard U19 upgrade flow

### Audit Trail

| Step | Action | Result |
|------|--------|--------|
| 1 | `gh pr view 1440 --json files` | 10 files changed; 0 under `src/template/` |
| 2 | `forge build --deny-warnings` | PASS |
| 3 | `forge fmt --check` | PASS |
| 4 | `./src/script/check-struct-order.sh` | PASS |
| 5 | `forge test --match-test testEthLockboxDiscoveredViaPortal` | PASS on Mainnet + Sepolia |

Wazabie (commented):
## Task Review (per `superchain-ops-task-review` skill)

PR contains 2 tasks targeting U19 (`op-contracts/v7.1.17`):
- `sep/090-U19-op-ink-mmz-soneium` — 6 chains bundled (OP, Ink, Metal, Mode, Zora, Soneium Minato)
- `sep/091-U19-unichain` — 1 chain (Unichain Sepolia)

### Input Verification — Task 090

| Input | Config Value | Authoritative Source | Match |
|-------|--------------|----------------------|-------|
| OPCM | `0x44e197058fb98fb3618453b3d90cdef6f5db8297` | on-chain `OPCM.version()` returns `"7.1.17"` | YES (on-chain) |
| OPCM (registry row) | same | superchain-registry `standard-versions-sepolia.toml` | **MISSING ROW** — registry currently lists this address under `v7.0.0-rc.*` only. No `op-contracts/v7.1.17` row yet. On-chain version proves the contract IS v7.1.17, but the config.toml comment cites a registry entry that doesn't exist. Either update the comment or wait for the registry to publish v7.1.17 before signing. |
| cannonKonaPrestate | `0x03cdb57a890f7e129d5abbe074f19425f94cf2b84f614aab008a96f91debae65` | superchain-registry `standard-prestates.toml` (kona-client v1.6.0-rc.1 cannon64-kona) | YES |
| cannonPrestate | `0xdead000000000000000000000000000000000000000000000000000000000000` | placeholder | **PLACEHOLDER** — PR description acknowledges this. CANNON is disabled in V700; OPCM zeros the slot regardless. Either replace with a real value or document that `0xdead…` is the intended permanent input before signing. |
| L1PAO nonce override = 50 | safe `0x1Eb2fFc903729a0F03966B917003800b145F56E2` | `cast call … "nonce()"` → `50` | YES |
| Security Council nonce override = 64 | safe `0xf64bc17485f0B4Ea5F06A96514182FC4cB561977` | `cast call …` → `64` | YES |
| Foundation nonce override = 70 | safe `0xDEe57160aAfCF04c34C887B5962D0a69676d3C8B` | `cast call …` → `70` | YES |
| Chain IDs (6) | 11155420, 763373, 1740, 919, 999999999, 1946 | superchain-registry | YES |
| StandardValidator | (read dynamically from OPCM via `opcmStandardValidator()`) | template fetches dynamically | YES — no hardcoded validator |

### Input Verification — Task 091

| Input | Config Value | Authoritative Source | Match |
|-------|--------------|----------------------|-------|
| OPCM | `0x44e197...` (same as 090) | on-chain v7.1.17 | YES |
| cannonKonaPrestate | `0x03cdb57a…ae65` | superchain-registry | YES |
| cannonPrestate | `0xdead…` | placeholder | **PLACEHOLDER** |
| Unichain L1PAO nonce override = 43 | safe `0xd363339eE47775888Df411A163c586a8BdEA9dbf` | `cast call …` → `43` | YES |
| Chain ID 1301 | Unichain Sepolia | superchain-registry | YES |

### Simulation

| Task | Command | Result |
|------|---------|--------|
| 090 (Foundation) | `just simulate-stack sep 090-U19-op-ink-mmz-soneium` | PASS — hashes extracted |
| 090 (Council) | `just simulate-stack sep 090-U19-op-ink-mmz-soneium council` | PASS — hashes extracted |
| 091 | `just simulate-stack sep 091-U19-unichain` | Did not produce hashes — consistent with the PR description's note that 091 cannot simulate standalone (Unichain's L1PAO cannot call `upgradeSuperchain` on the shared SuperchainConfig). Full stacked sim required before signing. |

### Hash Verification — Task 090

| Safe | Hash Type | VALIDATION.md | Simulation Output | Match |
|------|-----------|---------------|-------------------|-------|
| Security Council `0xf64b…1977` | Domain | `0xbe081970e9fc104bd1ea27e375cd21ec7bb1eec56bfe43347c3e36c5d27b8533` | identical | YES |
| Security Council | Message | `0xd21f74fd50e762864b58d044c112b53af322d3a7d407e877396db1e343a573bc` | identical | YES |
| Foundation `0xDEe5…3C8B` | Domain | `0x37e1f5dd3b92a004a23589b741196c8a214629d4ea3a690ec8e41ae45c689cbb` | identical | YES |
| Foundation | Message | `0x3c18c03d2f3aabde84e25a6e1e43b9a0c03136f7d3d4e9768609914f11e039f6` | identical | YES |

### Blocking Findings (must address before READY TO SIGN)

- [ ] **[VALIDATION.md, both tasks]** No "State Changes" section. Current VALIDATION.md contains only domain/message hashes + calldata. Per the skill's Phase 4, signers need explicit documentation of: per-chain state changes (DGF `gameImpls`, ASR `respectedGameType`, SystemConfig storage slots packed `superchainConfig`/`minBaseFee`, AddressManager rewrites, EthLockboxProxy writes, etc.), storage-key decoding, and a per-Safe state diff. PR description acknowledges this is incomplete.
- [ ] **Placeholder prestates** — both tasks pin `cannonPrestate = 0xdead…`. All VALIDATION.md hashes will change once real prestates land. Re-simulate everything afterward.

### Warnings

- [ ] **[090/README.md:3]** Empty markdown link in status line: `Status: [DRAFT, NOT READY TO SIGN]()`. Either drop the brackets or point at the PR.
- [ ] **[090 directory name]** `090-U19-op-ink-mmz-soneium` omits `zora` even though the config includes Zora Sepolia (chainId 999999999). The abbreviation `mmz` is opaque (Metal/Mode/Zora?). Config.toml header lists 6 chains; dirname implies 4–5. Consider renaming (e.g. `090-U19-sepolia-shared-pao`) or include `zora` in it.
- [ ] **[091/VALIDATION.md:6]** Placeholder warning is well placed; add an item to the PR description to remove it once real prestates land.
- [ ] **[090/.env, 091/.env]** `TENDERLY_GAS=15000000` sits at the Fusaka 15M ceiling. README warns to "confirm the Tenderly gas output stays under the 16,777,216 per-transaction gas cap" — tighten this env value to the actual simulated gas + a buffer rather than the ceiling, so an unintentional regression in gas cost is caught by the cap check.
- [ ] **[090/config.toml]** Each `OVERRIDES-CHALLENGER` and `OVERRIDES-L1PAOMULTISIG` in `expectedValidationErrors` needs an explicit justification in VALIDATION.md (which non-standard challenger/L1PAO, why it's allowed). Today only the config.toml comment notes this.
- [ ] **[090 sim warnings]** Sim emits `[WARN] Contract: EthLockboxProxy not found for chain: {Metal,Mode,Zora}`. These chains' Portals predate v5.2.0 or have lockbox = 0; U19 OPCM is what wires them. Tenderly state diff review should confirm `OptimismPortal.ethLockbox()` gets populated for the 3 currently-missing chains.

### Passed Checks

- [x] Simulation completes for 090 on both nested safes
- [x] All 4 verified Safe nonces match on-chain state (50 / 64 / 70 / 43)
- [x] OPCM at config address reports `version() = "7.1.17"` on-chain
- [x] Domain + Message hashes match VALIDATION.md for both nested safes (090)
- [x] cannonKonaPrestate matches superchain-registry kona-client v1.6.0-rc.1 entry
- [x] StandardValidator read dynamically from OPCM (not hardcoded in config)
- [x] All 7 chain IDs are valid Sepolia chains
- [x] Struct field order in TOML config matches `OPCMUpgrade` struct alphabetical fields (verified by `check-struct-order.sh`)
- [x] No hardcoded validator address in either config.toml

### User Verification Required

- [ ] **Tenderly state diff** for 090 on both safes: all 6 chains × U19 changes are present; no unexpected contracts modified.
- [ ] **Prestate verification**: once real prestates land, verify cannonKonaPrestate against the Notion Prestate Reviews page (`https://www.notion.so/oplabs/1bdf153ee16280aeadc0c101118ea969`).
- [ ] **Re-simulate everything** after prestate values replace `0xdead…`; all VALIDATION.md hashes will change.
- [ ] **091 full stacked sim**: simulate 090 → 091 sequentially once 090 is final, then capture Unichain's domain/message hash and update `091/VALIDATION.md`.

### Audit Trail

| Step | Action | Result |
|------|--------|--------|
| 1 | `cast call OPCM "version()"` (Sepolia) | `"7.1.17"` |
| 2 | `cast call <each safe> "nonce()"` × 4 | 50 / 64 / 70 / 43 — all match overrides |
| 3 | superchain-registry `standard-prestates.toml` lookup | kona-client v1.6.0-rc.1 cannon64-kona matches |
| 4 | superchain-registry `standard-versions-sepolia.toml` lookup | v7.1.17 row absent (registry pending) |
| 5 | `just simulate-stack sep 090-U19-op-ink-mmz-soneium` | PASS, Foundation hashes match VALIDATION.md |
| 6 | `just simulate-stack sep 090-U19-op-ink-mmz-soneium council` | PASS, Security Council hashes match VALIDATION.md |
| 7 | `just simulate-stack sep 091-U19-unichain` | Standalone sim does not produce hashes (expected per PR description) |

Wazabie:
## SuperchainAddressRegistry change — what & why

**Diff**: 8 lines added at `src/SuperchainAddressRegistry.sol` (one new interface method, one `try/catch` block) plus a new test at `test/registry/SuperchainAddressRegistry.t.sol`.

### What it does

After saving `OptimismPortalProxy` for a chain, the registry now reads `ethLockbox()` off the Portal and, if non-zero, registers the returned address as `EthLockboxProxy` for that chain:

```solidity
// SuperchainAddressRegistry.sol — _saveProxyEntries()

// EthLockboxProxy is exposed by OptimismPortal via the `ethLockbox()` getter
// starting at Portal v5.2.0 (op-contracts v7.x). Older Portals revert; chains
// that haven't wired a lockbox return address(0). Both cases skip registration.
try IFetcher(optimismPortalProxy).ethLockbox() returns (address ethLockbox) {
    if (ethLockbox != address(0)) saveAddress("EthLockboxProxy", chain, ethLockbox);
} catch {}
```

The `IFetcher` interface gains a single new line:
```solidity
function ethLockbox() external view returns (address);
```

### Why it's needed

`OPCMUpgradeV700` (U19) declares `EthLockboxProxy` in its `_taskStorageWrites()` allowlist (`src/template/OPCMUpgradeV700.sol:131`). At template setup time, `L2TaskBase._setAllowedStorageAccesses` resolves each entry against the registry — so the registry must be able to produce the `EthLockboxProxy` address for any chain whose Portal exposes one.

Three cases the `try/catch` + non-zero check handles cleanly:

| Portal state | `ethLockbox()` behavior | Registry behavior |
|---|---|---|
| Pre-v5.2.0 (no getter) | reverts | `catch {}` swallows, skip registration |
| v5.2.0+ but lockbox not yet wired | returns `address(0)` | non-zero guard skips registration |
| v5.2.0+ and wired | returns the lockbox address | registered under `EthLockboxProxy` |

The "skip" branch is what lets U19 run against chains whose Portals don't yet have a lockbox installed — for them the U19 upgrade itself is what wires it. The 086 stacked sim emits the expected `[WARN] Contract: EthLockboxProxy not found` for Metal, Mode, and Zora Sepolia, which are exactly the chains whose Portals haven't been wired yet.

### Test coverage

`test/registry/SuperchainAddressRegistry.t.sol` adds `testEthLockboxDiscoveredViaPortal`, which asserts a registry lookup of `EthLockboxProxy` for an OP chain returns a non-zero address. Passes on both Mainnet and Sepolia fork suites (`forge test --match-test testEthLockboxDiscoveredViaPortal` → 2 passed).

### Compatibility

- Backward compatible: chains whose Portal reverts on `ethLockbox()` continue to work; the lookup just no-ops.
- No removal of existing registry keys; pure addition.
- No impact on tasks that don't reference `EthLockboxProxy`.

Wazabie:
## MultisigTask change — reverted out of this PR

> **Update:** The two `src/tasks/MultisigTask.sol` fixes described below have been **reverted** and are no longer part of this PR. This PR now contains only the Sepolia U19 tasks — renumbered to `090-U19-op-ink-mmz-soneium` and `091-U19-unichain` — plus the `SuperchainAddressRegistry` change. The MultisigTask fixes will be handled in a separate PR.

<details>
<summary>Original rationale (historical — these changes are no longer in this PR)</summary>

## MultisigTask change — what & why

**Diff**: 2 small fixes in `src/tasks/MultisigTask.sol`.

### Fix 1: Storage-slot "value is an address" check now respects 160-bit upper bound

**Before**: when `_checkStateDiff` saw a storage write whose value did not contain code at `address(uint160(value))`, it asserted that *something* must have code there.

**After**: the assertion only runs when the slot value fits in 160 bits.

```solidity
// MultisigTask.sol — _checkStateDiff()
- } else {
+ } else if (value <= type(uint160).max) {
    // Only assert "no code" when the value could actually be an address.
    // A value above uint160.max is a packed storage slot (e.g. SystemConfig
    // slot 0x6c packs the `superchainConfig` address with `minBaseFee`), so its
    // low 160 bits are not an address — interpreting them as one is a false
    // positive. Such slots are skipped here.
    require(...);
}
```

**Why**: U19's SystemConfig packs an address into a slot that already contains other fields. The new code comment names the concrete case — SystemConfig slot `0x6c` packs the `superchainConfig` address (low 160 bits) with `minBaseFee` (upper bits). Once `minBaseFee` is non-zero the full slot value exceeds `uint160.max`. Taking the low 160 bits and asserting they point at code is meaningless — the slot doesn't store a single address.

Without this fix, U19 simulations against post-upgrade SystemConfig state would assert `address(uint160(packedValue)).code.length != 0` against whatever happens to be encoded in the low 160 bits — a false positive that has nothing to do with the actual address slot. The bound check correctly skips packed slots that aren't pure addresses.

### Fix 2: `gasRefunded` no longer rejected when negative

**Before**:
```solidity
require(gasInfo.gasRefunded >= 0, "MultisigTask: negative gas refund is invalid");
uint256 gasRefunded = uint256(uint64(gasInfo.gasRefunded));
```

**After**:
```solidity
uint256 gasRefunded = gasInfo.gasRefunded > 0 ? uint256(uint64(gasInfo.gasRefunded)) : 0;
```

**Why**: `vm.lastCallGas().gasRefunded` is an `int64` whose value forge exposes as a signed *net* refund delta, not the EVM's refund counter. It can legitimately be negative for some calls. The previous `require(... >= 0)` was misinformed about the field's semantics and would abort the post-execution gas check on perfectly valid simulations. Clamping to zero preserves the gas-cap check's correctness: if the net refund is negative, treating it as zero gives a conservative upper bound for `gasConsumedBeforeRefund`, which is what the cap check (`<= MAX_GAS_LIMIT`) wants anyway.

This is needed because the U19 090 task executes 6 chains worth of upgrade calls in a single bundled transaction, and the `vm.lastCallGas()` for that combined run can expose a negative refund delta that trips the old require.

### Net effect on this PR

Both fixes unlock U19 task simulation. Neither relaxes a real safety check:
- **Fix 1** narrows the address-shape check to slots that could plausibly hold a single address. Slots with values > `uint160.max` are unambiguously packed and are not address-like.
- **Fix 2** keeps the gas-cap check intact while removing a false assertion about `gasRefunded`'s sign. The cap (`gasConsumedBeforeRefund <= MAX_GAS_LIMIT`) is unchanged; only the refund-decoding step now tolerates the documented int64 contract.

Both changes pass `forge build --deny-warnings`, `forge fmt --check`, and `./src/script/check-struct-order.sh`. No regressions in the registry test suite.

</details>

Wazabie (commented):
## Task Review — addendum: 091 stacked-sim hashes verified

Correction to the earlier task-review comment. `just simulate-stack sep 091-U19-unichain` does complete (the "stack" recipe auto-runs 090 → 091), and the Unichain hashes are extractable. Reposting the hash-verification line for 091:

| Safe | Hash Type | VALIDATION.md | Simulation Output | Match |
|------|-----------|---------------|-------------------|-------|
| Unichain L1PAO `0xd363…9dbf` | Domain | `0x2fedecce87979400ff00d5cec4c77da942d43ab3b9db4a5ffc51bb2ef498f30b` | identical | YES |
| Unichain L1PAO | Message | `0x77a14c150de82e4f815c68e69d1dcb21a21f387af094c86202856d9ac5c5b199` | identical | YES |

So both tasks' current VALIDATION.md domain/message hashes match the simulator output. They will still all change once the real `cannonPrestate` replaces `0xdead…`, so the "re-simulate before signing" item remains the binding blocker.

Wazabie:
## Re-added: `MultisigTask._checkStateDiff` packed-slot guard (commit b6a7905d)

Reinstating the "Fix 1" from my [earlier comment](https://github.com/ethereum-optimism/superchain-ops/pull/1440#issuecomment-4595287977) — the stacked Sepolia sim won't pass without it.

### Why it's needed

`stacked_simulation_sep` is currently failing with:

```
Error: script failed: Likely address in storage has unexpected code
  account: 0xaeE94b9aB7752D3F7704bDE212c0C6A0b701571D   (SystemConfig proxy)
  slot:    0x000000000000000000000000000000000000000000000000000000000000006c
  value:   0x00000000000000000007a120c2be75506d5724086deb7245bd260cc9753911be
```

Slot `0x6c` on `SystemConfig` is the well-known **packed** slot:

| bits      | field             | value in this run                            |
|-----------|-------------------|----------------------------------------------|
| `[0:160)` | `superchainConfig`| `0xc2be75506d5724086deb7245bd260cc9753911be` |
| `[160:)`  | `minBaseFee`      | `0x7a120` (= **500000**)                     |

Because `minBaseFee != 0`, the full slot value exceeds `type(uint160).max`. Control flow in `_checkStateDiff`:

1. `Utils.isLikelyAddressThatShouldHaveCode(value, ...)` correctly returns **false** (non-zero upper bits → not address-shaped).
2. The previous code unconditionally fell into `else` and asserted `address(uint160(value)).code.length == 0` — i.e. it truncated to the low 160 bits and required *no* code there.
3. The low 160 bits **are** a real address (`superchainConfig`) with code → assertion trips.

The U19 `OPCMUpgradeV700` template writes a non-zero `minBaseFee` into this slot, so every Sepolia chain in the stacked run hits this on the SystemConfig state diff.

### The fix

Guard the `else` branch with an upper-bound check so packed slots are skipped:

```diff
-                } else {
+                } else if (value <= type(uint160).max) {
                     // Only run the "no unexpected code" check when the slot value
                     // could actually be an address. Values above uint160.max are
                     // packed storage slots (e.g. SystemConfig slot 0x6c packs
                     // `superchainConfig` in the low 160 bits with `minBaseFee`
                     // in the upper bits), so reinterpreting the low 160 bits as
                     // an address is a false positive — skip such slots.
                     string memory err = string.concat("Likely address in storage has unexpected code\n", ...);
                     require(address(uint160(value)).code.length == 0, err);
                 }
```

### Why this doesn't weaken safety

The check's intent is "if a storage write looks like an address, make sure it's not the **wrong** kind of address (code vs. no code)." A value above `uint160.max` is unambiguously not a single address — its low 160 bits are not an address either, they're a co-tenant of an unrelated field. There is nothing meaningful to assert about the code-ness of those bits, so skipping is the correct semantics.

The companion check on the *other* branch (`isLikelyAddressThatShouldHaveCode` → require code != 0) is unchanged, and the same packed-slot bound is implicit there because `isLikelyAddressThatShouldHaveCode` requires `value <= type(uint160).max` to even return true.

### Verification

- `forge build --deny-warnings` ✅
- `forge fmt --check` ✅
- Stacked sim re-run pending in CI on this push.

If the broader MultisigTask fix should still live in a separate PR (per my earlier "reverted out" note), happy to extract this single hunk — but leaving it here means #1440's CI can actually go green.

Wazabie:
## Aligned permissioned vs permissionless wiring per Adrian's U19 clarification (commit 5aba4d24)

Picked up the rule from the Sepolia U19 thread:

> Permissioned chain should keep respected game type as 1. Permissioned chains must NOT set an implementation for game type 8. They can use a placeholder (0xdead) for the permissioned game prestate.
> Permissionless chains have to set respected game type 8 but keep an implementation set for game type 1 as a fallback. The prestate for game type 1 can be a placeholder (0xdead).

### Current Sepolia on-chain state (queried via Portal `respectedGameType()` + DGF `gameImpls`)

| Chain          | resp. game type | gameImpls[0] | gameImpls[1] | gameImpls[8] | classification |
|----------------|-----------------|--------------|--------------|--------------|----------------|
| OP             | 0               | set          | set          | set          | permissionless |
| Ink            | 0               | set          | set          | set          | permissionless |
| Metal          | **1**           | 0x0          | set          | **0x0**      | **permissioned** |
| Mode           | **1**           | 0x0          | set          | **0x0**      | **permissioned** |
| Zora           | **1**           | 0x0          | set          | **0x0**      | **permissioned** |
| Soneium Minato | 0               | set          | set          | set          | permissionless |
| Unichain       | 0               | set          | set          | set          | permissionless |

Before this commit, **every chain** in 090 had `startingRespectedGameType = 8` and a real Kona prestate. That would have:
- Rotated Metal/Mode/Zora's respectedGameType from `1` → `8` ❌
- Installed a `CANNON_KONA` impl on Metal/Mode/Zora ❌

And config-only changes weren't sufficient — the template's `_isEnabled` was chain-agnostic and forced CANNON_KONA enabled for every chain regardless of prestate.

### Changes

**`src/template/OPCMUpgradeV700.sol`**
- New `_isPermissionedChain(chainId)` helper, derived from `startingRespectedGameType == PERMISSIONED_CANNON`. Inferred rather than configured as a separate TOML field so the two values cannot drift apart.
- Refactored `_isEnabled(IDisputeGameFactory, uint32 gt)` → `_isEnabled(uint32 gt, bool permissioned)`. `CANNON_KONA` is now `enabled = !permissioned`. `PERMISSIONED_CANNON` stays always-enabled.
- Threaded `permissioned` through `_gameConfig` and `_gameConfigs(chainId)`.
- Tightened `_extraInstructions` to require `startingRespectedGameType ∈ {PERMISSIONED_CANNON, CANNON_KONA}` (was: just `!= CANNON`) so a config typo can't reach the OPCM.

**`src/tasks/sep/090-U19-op-ink-mmz-soneium/config.toml`**
- Metal (1740), Mode (919), Zora (999999999): `startingRespectedGameType = 1`, `cannonKonaPrestate = 0xdead000...` (placeholder — never reaches `setImplementation` because CANNON_KONA is disabled for these chains).
- OP/Ink/Soneium-Minato: unchanged (rotate to 8 with the real Kona prestate).
- Header expanded to document the split.

**`src/tasks/sep/091-U19-unichain/config.toml`**: unchanged — Unichain is permissionless on chain.

### Why this is safe

- For Metal/Mode/Zora, `gameImpls[8]` is already `0x0` on chain. With `_isEnabled(CANNON_KONA, true) = false`, OPCMv2 emits `setImplementation(8, 0, "")` — a no-op write. The slot's post-state matches its pre-state.
- For permissionless chains (OP/Ink/Soneium-Minato/Unichain), behavior is bit-for-bit identical to before: `permissioned=false` → `_isEnabled(CANNON_KONA, false) = true` → full Kona wiring.
- Struct ABI is unchanged (no new TOML field), so the EXECUTED historical configs (076/079/086/089) would still decode if ever re-played.
- Confirmed `testRegressionCallDataMatches_OPCMUpgradeV700_U19_Betanet` still passes — the U19 betanet example task's calldata is bit-identical under the refactored template.

### Verification

- `forge build --deny-warnings` ✅
- `forge fmt --check` ✅
- `testRegressionCallDataMatches_OPCMUpgradeV700_U19_Betanet` ✅ (1 passed, 0 failed)
- CircleCI `stacked_simulation_sep` re-runs on this push.

Spot-check post-sim on Tenderly for one permissioned chain (e.g. Metal):
- `AnchorStateRegistry.respectedGameType` should stay `1`
- `DisputeGameFactory.gameImpls[8]` should stay `0x0`
- `DisputeGameFactory.gameImpls[1]` should rewrite to the v7.1.17 `PermissionedDisputeGame` impl

src/tasks/MultisigTask.sol
```diff
@@ -475,8 +475,13 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager, TaskManage
...
                     require(address(uint160(value)).code.length != 0, err);
-                } else {
-                    // Log account, slot, and value if there is code.
+                } else if (value <= type(uint160).max) {
+                    // Only run the "no unexpected code" check when the slot value
+                    // could actually be an address. Values above uint160.max are
+                    // packed storage slots (e.g. SystemConfig slot 0x6c packs
+                    // `superchainConfig` in the low 160 bits with `minBaseFee`
+                    // in the upper bits), so reinterpreting the low 160 bits as
+                    // an address is a false positive — skip such slots.
```
Ethnical:
```suggestion
```

src/tasks/MultisigTask.sol
Ethnical, line 479:
```suggestion
                uint256 value = uint160(storageAccess.newValue); // we do this since we know that solidity write the address on the last 20 bytes of a slot and we don't wants pollute the value with the highest bytes that would make this address invalid and higher than uint160. 
```

src/tasks/sep/099-U19-op-ink-mmz-soneium/config.toml
```diff
@@ -0,0 +1,144 @@
...
+cannonPrestate = "0xdead000000000000000000000000000000000000000000000000000000000000"
+# Permissioned chain: `_isPermissionedChain` is true (startingRespectedGameType == 1), so
+# CANNON_KONA is `enabled=false` and never wired. The kona prestate MUST be the zero address
+# (not a 0xdead placeholder) — enforced by OPCMUpgradeV700._templateSetup.
+cannonKonaPrestate = "0x0000000000000000000000000000000000000000000000000000000000000000"
+# PLDG-10: CANNON slot zeroed by V700. CKDG-NOSHAPE + CKDG-10: CANNON_KONA (game type 8) is not
+# wired on this permissioned chain (kona prestate = 0), so the standard validator reports it has
+# no shape / zero impl. Captured live from the v7.1.17 OPCM validator.
+expectedValidationErrors = "CKDG-NOSHAPE,PLDG-10,CKDG-10"
+initBond = 0
```
Ethnical, line 90: I don't remember but just to check back the initbond has to be zero when this is the permissioned? 
I can't remember if with cannon permissioned it was zero also?

Ethnical:
Okay, I did the simulation and found out that the tenderly link is broken. I had to add the override manually, and the gas is way higher, which is strange. The hard cap of 15M wasn't working; I don't understand why since only 12M gas is used here.
https://dashboard.tenderly.co/oplabs/op-sepolia/simulator/new?network=11155111&from=0x5aAdFB43eF8dAF45DD80F4676345b7676f1D70e3&contractAddress=0xcA11bde05977b3631167028862bE2a173976CA11&sharedSimulation=true
EDIT: Actually was a Tenderly bug, this work well with 15M gas, 
https://dashboard.tenderly.co/oplabs/op-sepolia/simulator/a5cb124f-f1be-4327-9265-a9c013f1c9b5

Wazabie:
Renumbered the two tasks to free up 097/098 for the new Osaki Sepolia L1/L2 PAO transfer PR (#1455), and bumped the three shared Sepolia Safe nonces in `099`'s `stateOverrides` by +2 to account for the nonce bumps those Osaki tasks consume on the L1PAO / FoundationUpgradeSafe / SecurityCouncil before this one is signed.

- `097-U19-op-ink-mmz-soneium` → `099-U19-op-ink-mmz-soneium` (live + 2: L1PAO 50→52, Foundation 70→72, Council 64→66)
- `098-U19-unichain` → `100-U19-unichain` (no nonce changes — Unichain's PAO is a separate Safe)
- Cross-references in READMEs and config comments updated.
- `099/VALIDATION.md` hashes refreshed and verified via `simulate-stack sep 099-U19-op-ink-mmz-soneium <foundation|council>` (framework hash check passes).

`100`'s hashes were not regenerated — its config carries the same `[!CAUTION]` placeholder-prestates note from before and still needs its own re-simulation when the real prestates land.

Happy to swap the order if the Osaki PR doesn't land first; let me know.

src/tasks/sep/099-U19-op-ink-mmz-soneium/config.toml
Wazabie: Permissioned bond 0, permissionless and kona are set to 0.08eth

src/tasks/MultisigTask.sol
sbvegan: this is being addressed differently with you recent PR and the one solomon is having reviewed: https://github.com/ethereum-optimism/superchain-ops/pull/1452

donoso-eth (approved):

PR #1445, opened 2026-06-02
U19 betanet update prestate + set gameargs template

Wazabie:
PR adds task to update U19 betanets prestates and adds a template to set the gameargs on the disputegamefactory


donoso-eth (dismissed):

src/tasks/sep/090-U19-betanet-kona-prestate-update/config.toml
```diff
@@ -0,0 +1,67 @@
+# Kona absolute-prestate update for the U19 OP Labs Betanet (karst-u19-beta).
+#
+# Both betanet chains — karst-u19-beta-0 (420110023) and karst-u19-beta-1
+# (420110024) — are already on U19 (op-contracts/v7.1.17). This task does NOT
+# re-run the U19 upgrade. It only re-points the CANNON_KONA (game type 8) slot in
+# each chain's DisputeGameFactory at a new absolute prestate.
+#
+# Mechanism: the CANNON_KONA game uses the blueprint pattern — the absolute
```
Inphi: nit: it uses a factory pattern not creator pattern. The blueprint pattern was used up until U16.
With the creator pattern; dispute game type implementations are fully deployed. When a dispute game is created, an implementation is cloned and suffixed with additional arguments that are specific to the game type.

src/template/SetDisputeGameArgs.sol
```diff
@@ -0,0 +1,319 @@
...
+            require(factory.gameImpls(r.gameType) == r.impl, "SetDisputeGameArgs: impl mismatch");
+            require(
+                keccak256(factory.gameArgs(r.gameType)) == keccak256(r.gameArgs),
+                "SetDisputeGameArgs: gameArgs mismatch"
+            );
+
+            if (r.impl != address(0)) {
+                // A live game must carry a non-zero init bond, otherwise games are free to create and
+                // the dispute economics break.
+                require(factory.initBonds(r.gameType) != 0, "SetDisputeGameArgs: zero init bond on live game");
```
Inphi: nit: it's OK to permit zero init bonds for the permissioned dispute game. (We've gone back and forth on this soft policy but we've now relaxed this :-)

src/template/SetDisputeGameArgs.sol
```diff
@@ -0,0 +1,319 @@
...
+    {
+        string memory key = string.concat(base, ".", field);
+        return toml.keyExists(key) ? toml.readAddress(key) : dflt;
+    }
+
+    /// @notice Decode an on-chain gameArgs blob into its fields. Reverts on an unexpected length.
+    function _decodeArgs(bytes memory args) internal pure returns (GameArgsFields memory f) {
+        if (args.length == 0) return f; // new or disabled slot — nothing to keep.
+        require(args.length == 124 || args.length == 164, "SetDisputeGameArgs: unexpected on-chain gameArgs length");
+        f.permissioned = args.length == 164;
```
Inphi: For safety, the permissioned flag should be based on the gameType (`gameType == 1` for PERMISSIONED_CANNON) rather than assume based on the args.length.

src/template/SetDisputeGameArgs.sol
```diff
@@ -0,0 +1,319 @@
...
+    ///         validate, and push the resolved action. Kept as its own frame to bound stack depth.
+    function _resolveRow(string memory toml, string memory base) internal {
+        // chainId + gameType are the only required fields — they identify the slot.
+        require(toml.keyExists(string.concat(base, ".chainId")), "SetDisputeGameArgs: row missing chainId");
+        require(toml.keyExists(string.concat(base, ".gameType")), "SetDisputeGameArgs: row missing gameType");
+        uint256 chainId = toml.readUint(string.concat(base, ".chainId"));
+        uint256 gtRaw = toml.readUint(string.concat(base, ".gameType"));
+        require(chainId != 0, "SetDisputeGameArgs: chainId zero");
+        require(gtRaw <= type(uint32).max, "SetDisputeGameArgs: gameType out of range");
+        uint32 gameType = uint32(gtRaw);
```
Inphi, line 96: let's prevent this template from being used for unsupported game types. Revert if the gameType neither 0, 1, nor 8, so we don't accidentally use this for others. Otherwise, we could invalidly set bad args for, `SUPER_CANNON_KONA`, for example.

src/template/SetDisputeGameArgs.sol
```diff
@@ -0,0 +1,319 @@
...
+                "SetDisputeGameArgs: duplicate (chainId, gameType) row"
+            );
+        }
+
+        IDisputeGameFactory factory =
+            IDisputeGameFactory(superchainAddrRegistry.getAddress("DisputeGameFactoryProxy", chainId));
+        require(address(factory).code.length > 0, "SetDisputeGameArgs: DisputeGameFactory has no code");
+
+        // Start from the live gameArgs, then override only the fields present in the TOML.
+        GameArgsFields memory m = _decodeArgs(factory.gameArgs(gameType));
```
Inphi: if we're setting CANNON_KONA, for example, for the first time, `m` will be empty. We should then assert that the config.toml will fully hydrates `m` with non-zero values. So that we don't then set the gameArgs to 0x000...

Otherwise, the _validate function ~~wouldn't~~ would still pass and not catch the case where CANNON_KONA is introduced for the first time, and the config.toml set all of its fields to zero.

Inphi (dismissed):
The task looks good. Just have some comments on making the template more robust.

Inphi (dismissed):

jelias2 (approved):

Inphi (approved):

test/tasks/example/sep/038-set-dispute-game-args/config.toml
```diff
@@ -0,0 +1,44 @@
...
+    {name = "karst-u19-beta-0", chainId = 420110023}
+]
+
+# Permissionless layout (124 bytes):
+#   prestate(32) | vm(20) | anchorStateRegistry(20) | delayedWETH(20) | chainId(32)
+[[gameConfig]]
+chainId = 420110023
+gameType = 8                                                                   # CANNON_KONA (permissionless)
+impl = "0x6dDBa09bc4cCB0D6Ca9Fc5350580f74165707499"                            # changed (real FDG v2.2.0 impl)
+prestate = "0x03000000000000000000000000000000000000000000000000000000deadbeef" # changed (synthetic)
```
Ethnical, line 26:
```suggestion
prestate = "0x03000000000000000000000000000000000000000000000000000000deadbeef" # changed (synthetic) Worth noting, this is an invalid prestate, but since example tasks just simulate them and don't verify the prestate value, any value is possible here.
```

test/tasks/example/sep/038-set-dispute-game-args/config.toml
```diff
@@ -0,0 +1,44 @@
...
+vm = "0x6463dEE3828677F6270d83d45408044fc5eDB908"                              # changed (op-program MIPS)
```
Ethnical, line 27: why the VM here is the MIPS one if we have the gametype set to 8 (which is kona above?)
@Inphi can you tell me if you have more information about this?

Ethnical: since here we just wants to simulate the template we don't really care if the value are really correct.

Ethnical:
In the future would be great to have the betanet folder itself.

Ethnical (approved):
approved since already executed we prove that the disputegame template updated correctly.

PR #1447, opened 2026-06-02
feat: U19 permissionless flip for karst-u19-alpha + beta-0 correction (CANNON_KONA)

Wazabie:
Split out of #1440. Contains the **karst-u19-alpha** permissionless rollout, the **karst-u19-beta-0** permissioned correction, and the `SetDisputeGameImpl` template extension they need. The ink/uni standard-chain U19 tasks and the OPCMUpgradeV700 / SuperchainAddressRegistry fixes stay in #1440.

## Tasks (Sepolia, shared PAO Safe `0x8E851F7d8bAeaD95F592847a020cAC7A062dafd9`)

| # | Chain | Action | Template |
|---|---|---|---|
| 092 | karst-u19-alpha-1 | wire CANNON_KONA (8) reusing `0x2DDA` + alpha kona prestate `0x03ff5c52…`; permissioned (1) prestate → `0xdead`; game 0 stays `0x0` | SetDisputeGameImpl |
| 093 | karst-u19-alpha-1 | `respectedGameType` → 8 | SetRespectedGameTypeTemplate |
| 094 | karst-u19-alpha-0 | permissioned (1) prestate → `0xdead` (stays permissioned; no game 8) | SetDisputeGameImpl |
| 095 | karst-u19-beta-0 | `respectedGameType` 8 → 1 (correction) | SetRespectedGameTypeTemplate |
| 096 | karst-u19-beta-0 | zero CANNON_KONA (8) impl (correction) | SetDisputeGameImpl |

Resulting end-states match the U19 clarification (permissioned: respected 1 + no game 8; permissionless: respected 8 + kona, with game 1 fallback at `0xdead`):
- **alpha-1** permissionless: g0=`0x0`, g1=`0xe1dF`/`0xdead`, g8=`0x2DDA`/kona, respected **8**
- **alpha-0** permissioned: g0=`0x0`, g1=`0xe1dF`/`0xdead`, g8=`0x0`, respected **1**
- **beta-0** permissioned (corrected from an earlier respected=8 / kona-wired state): g8=`0x0`, respected **1**

## Template change
`SetDisputeGameImpl` gains an optional `[[konaGameImplConfig]]` block (game type 8) — backward-compatible; existing FDG/PDG tasks unaffected. Supports wire (impl + 124-byte gameArgs) and disable (impl `0x0`).

## Prestate
[...]

PR #1448, opened 2026-06-03
feat: U19 templates — OPCMUpgradeV700 permissioned + SetDisputeGameArgs/Impl

Wazabie:
## Summary

Pulls together the reusable template / src / test changes (no Sepolia task directories) from three in-flight U19 branches into a single PR:

**From #1440**
- `src/template/OPCMUpgradeV700.sol` — permissioned upgrade path
- `src/SuperchainAddressRegistry.sol` — EthLockbox discovery via Portal
- `src/tasks/MultisigTask.sol`
- `test/registry/SuperchainAddressRegistry.t.sol` — registry test
- `test/tasks/example/sep/038-opcm-upgrade-v700-permissioned/` — example task
- permissioned regression test in `test/tasks/Regression.t.sol`

**From #1447**
- `src/template/SetDisputeGameImpl.sol` — template update

**From #1445**
- `src/template/SetDisputeGameArgs.sol` — new template
- `test/tasks/example/sep/038-set-dispute-game-args/` — example task
- regression test in `test/tasks/Regression.t.sol`

The actual `src/tasks/sep/090…096` task directories from those branches are intentionally **not** included here.

## Testing

- `forge build` — clean
- `testRegressionCallDataMatches_OPCMUpgradeV700_U19_Permissioned` — PASS
- `testRegressionCallDataMatches_SetDisputeGameArgs` — PASS
- `testEthLockboxDiscoveredViaPortal` — PASS

(The pre-existing `test_allowOverwrite_*` registry tests fail only on RPC historical-state availability, unrelated to this change.)


test/tasks/example/sep/039-opcm-upgrade-v700-permissioned/config.toml
```diff
@@ -0,0 +1,41 @@
+# Upgrade 19 example task — PERMISSIONED-chain path (Sepolia L1).
```
Ethnical, line 1: I don't see the 037,038 in the ceremony tracker for sep. 
is that intentional.

src/tasks/MultisigTask.sol
```diff
@@ -475,8 +475,13 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager, TaskManage
                     string memory err = string.concat("Likely address in storage has no code\n", "  account: ", vm.toString(account), "\n  slot:    ", vm.toString(storageAccess.slot), "\n  value:   ", vm.toString(bytes32(value)));
                     // forgefmt: disable-end
                     require(address(uint160(value)).code.length != 0, err);
-                } else {
-                    // Log account, slot, and value if there is code.
+                } else if (value <= type(uint160).max) {
```
Ethnical: Interesting I like the idea, there is no way that the address is on the slot but also there is some upper by for another reason, making this assumption not working? 
For example a contract that has upper value as setting and lower value as an address that would make if failing?

Ethnical: Okay after review with @Wazabie, I feel this is unsafe since this extremely common to have the first 12 bytes fill by NON-ZERO values. 
the schema is like this: 
```
  ASCII — the whole 32 bytes of slot 0x6c

                           SystemConfig storage slot 108  (0x6c)
                           32 bytes, big-endian (MSB left → LSB right)

    byte:  31  30  29  28 │ 27  26  25  24  23  22  21  20 │ 19 18 ... 02 01 00
          ┌───┬───┬───┬───┼───┬───┬───┬───┬───┬───┬───┬───┼──────────────────────┐
          │ 00  00  00  00 │            minBaseFee         │   superchainConfig   │
          │   (unused)     │           uint64 (8 B)        │     address (20 B)   │
          └───┴───┴───┴───┴───────────────────────────────┴──────────────────────┘
           offset 31..28        offset 27 .......... 20      offset 19 ........ 0
            4 bytes zero            8 bytes                      20 bytes
          └──────────────┘ └───────────────────────────┘ └──────────────────────┘
              padding            minBaseFee (wei)            SuperchainConfig addr
                MSB ◄─────────────────────────────────────────────────► LSB


    Example word (minBaseFee = 1 gwei = 0x3B9ACA00, scconf = 0xC2Be75...0d5c):

      0x  00000000  000000003B9ACA00  C2Be75506d5724086DEB7245bd260Cc9753911Be
          └─pad──┘  └──minBaseFee───┘  └─────────superchainConfig─────────────┘
```

what we should do is ONLY testing if the last 20 bytes are an `uint160` IMO.

src/tasks/MultisigTask.sol
Ethnical, line 479:
```suggestion
                uint256 value = uint160(storageAccess.newValue); // we do this since we know that solidity write the address on the last 20 bytes of a slot and we don't wants pollute the value with the highest bytes that would make this address invalid and higher than uint160. 
```

Ethnical: removed and fixed in the PR https://github.com/ethereum-optimism/superchain-ops/pull/1450

src/template/OPCMUpgradeV700.sol
```diff
@@ -216,31 +247,39 @@ contract OPCMUpgradeV700 is OPCMTaskBase {
 
     /// @notice U19-specific enabled-flag policy for each of the 7 OPCMv2 game-type slots.
     /// @dev
-    ///   - `CANNON_KONA`: always enabled. This is what U19 introduces — `kona-client`
-    ///     becomes the primary FPVM and the new respected game type.
+    ///   - `CANNON_KONA`: enabled iff the chain is **permissionless** post-upgrade. On
```
Ethnical:
```suggestion
    ///   - `CANNON_KONA`: enabled if the chain is **permissionless** post-upgrade. On
```

src/template/OPCMUpgradeV700.sol
```diff
@@ -216,31 +247,39 @@ contract OPCMUpgradeV700 is OPCMTaskBase {
...
+    ///     post-upgrade — this is the explicit ask per the U19 spec. Existing
     ///     CANNON game instances are unaffected (their impl is bytecode-bound at game
     ///     creation time) so they can still resolve to completion; only new CANNON games
     ///     are blocked from being created.
     ///   - `SUPER_CANNON`, `SUPER_PERMISSIONED_CANNON`, `SUPER_CANNON_KONA`,
     ///     `ZK_DISPUTE_GAME`: always disabled. The v7.1.17 OPCM container ships
     ///     `address(0)` for these impls; they belong to a later release, not U19.
-    function _isEnabled(IDisputeGameFactory, uint32 gt) internal pure returns (bool) {
-        if (gt == CANNON_KONA) return true;
+    function _isEnabled(uint32 gt, bool permissioned) internal pure returns (bool) {
```
Ethnical, line 279: it's strange to base the output return based on the input for permissioned that can be controlled before. 
We should likely use an hardcoded for permissioned that should be `is_permissioned` hardcoded globally that can be used.

Ethnical: actually line 345 solve this.

src/template/OPCMUpgradeV700.sol
```diff
@@ -314,24 +355,35 @@ contract OPCMUpgradeV700 is OPCMTaskBase {
...
+        // two values that line up with this template's enabled slots are
+        // PERMISSIONED_CANNON (1) and CANNON_KONA (8). The OPCM would already revert on
+        // any other value (the slot would be disabled), but rotating the respected game
+        // type to anything else is an irreversible footgun on the AnchorStateRegistry —
+        // refuse it at the override construction site so a config typo can never reach
+        // the OPCM.
         uint32 startingRespectedGameType = upgrades[chainId].startingRespectedGameType;
-        require(startingRespectedGameType != CANNON, "OPCMUpgradeV700: startingRespectedGameType cannot be CANNON");
+        require(
+            startingRespectedGameType == PERMISSIONED_CANNON || startingRespectedGameType == CANNON_KONA,
```
Ethnical, line 387: `startingRespectedGameType` is actually the ending respected game? So the GameType that would be final after the operation. 
Since this is extremely confusing we should: 
1. try to change the name variable but @Wazabie tried and had some issues? 
2. Add clear comments everywhere to avoid confusion.

src/template/SetDisputeGameImpl.sol
```diff
@@ -94,6 +125,40 @@ contract SetDisputeGameImpl is L2TaskBase {
...
+            if (address(factory.gameImpls(CANNON_KONA)) != address(0) || factory.gameArgs(CANNON_KONA).length != 0) {
+                factory.setImplementation(CANNON_KONA, address(0), "");
+            }
+            return;
+        }
+
+        bool slotMatches = address(factory.gameImpls(CANNON_KONA)) == k.impl
+            && keccak256(factory.gameArgs(CANNON_KONA)) == keccak256(k.gameArgs);
+        if (!slotMatches) {
+            _validateGameArgsFormat(k.gameArgs, chainId, false); // permissionless 124-byte layout
```
Ethnical, line 156: what is the 124 byte layout permissioneless?

Ethnical: if doesn't match the slot then this is implementing Kona permissionless can we add a comment to explain why? I don't understand for now why?

Ethnical: This is interesting because seems that we need to call this function multiples times. 
For example, 
You set the a game with Kona Permissionless for the first time then, you will end up in 
```
if (!slotMatches) {
            _validateGameArgsFormat(k.gameArgs, chainId, false); // perm
...
```
and never set the bond of the game?

Ethnical: my bad I am tired the if is outside and indeed if the slot doesn't match we do both set & init.

src/template/SetDisputeGameImpl.sol
```diff
@@ -134,6 +199,20 @@ contract SetDisputeGameImpl is L2TaskBase {
...
+    function _validateKona(IDisputeGameFactory factory, uint256 chainId) internal view {
+        if (!konaConfigured[chainId]) return;
+        KonaGameImplConfig storage k = konaCfg[chainId];
+        require(address(factory.gameImpls(CANNON_KONA)) == k.impl, "KONA implementation mismatch");
+        if (k.impl == address(0)) {
+            require(factory.gameArgs(CANNON_KONA).length == 0, "KONA gameArgs not cleared");
+            return;
+        }
+        require(keccak256(factory.gameArgs(CANNON_KONA)) == keccak256(k.gameArgs), "KONA game args mismatch");
+        if (k.bond != 0) require(factory.initBonds(CANNON_KONA) == k.bond, "KONA init bond mismatch");
```
Ethnical: if the bond is `0` we absolutely wants to revert no? 
This will indicates that we forgot to update the bond and this catatrosphic on L1 the next block since game might be free to open.
This is pseudo code (might be needed to write something better but roughly: 
```suggestion
        if (k.bond > 0) require(factory.initBonds(CANNON_KONA) == k.bond, "KONA init bond mismatch");
        else {
        revert ("KONA bond is ZERO");
        }
```

src/SuperchainAddressRegistry.sol
```diff
@@ -341,6 +342,13 @@ contract SuperchainAddressRegistry is StdChains {
 
         optimismPortalProxy = getOptimismPortalProxy(l1CrossDomainMessengerProxy);
         saveAddress("OptimismPortalProxy", chain, optimismPortalProxy);
+
+        // EthLockboxProxy is exposed by OptimismPortal via the `ethLockbox()` getter
+        // starting at Portal v5.2.0 (op-contracts v7.x). Older Portals revert; chains
+        // that haven't wired a lockbox return address(0). Both cases skip registration.
```
Ethnical, line 348: might interesting to indicates that this would not revert if this is zero and return `address(0)`

Ethnical: My bad this is clearly stipulated 

```
that haven't wired a lockbox return address(0).
```

test/tasks/example/sep/039-opcm-upgrade-v700-permissioned/config.toml
```diff
@@ -0,0 +1,41 @@
...
+cannonPrestate = "0xdead000000000000000000000000000000000000000000000000000000000000"
+# Permissioned chain: CANNON_KONA is not wired, so the kona prestate MUST be the zero address.
+cannonKonaPrestate = "0x0000000000000000000000000000000000000000000000000000000000000000"
+# Live validator output captured during simulation against the dev v7.1.17 OPCM at block
+# 10808500. A permissioned chain with kona prestate = 0 additionally reports CKDG-NOSHAPE and
+# CKDG-10 (CANNON_KONA game type 8 has no shape / zero impl) on top of the betanet overrides.
+expectedValidationErrors = "OVERRIDES-L1PAOMULTISIG,OVERRIDES-CHALLENGER,SYSCON-130,CKDG-NOSHAPE,PLDG-10,CKDG-10"
+initBond = 0
+# Permissioned: respected game type stays at PERMISSIONED_CANNON (1).
+startingRespectedGameType = 1
```
Ethnical, line 34:
```suggestion
startingRespectedGameType = 1 # `startingRespectedGameType` means we update TO this GameType so PERMISSIONED_CANNON.
```

src/template/OPCMUpgradeV700.sol
Wazabie: Comments added to avoid confusion. The naming is aligned to what we have in [OPCM here](https://github.com/ethereum-optimism/optimism/blob/86d125865f06eacf7dd0a044400dbe310d45ac07/packages/contracts-bedrock/src/L1/opcm/OPContractsManagerV2.sol#L90) if it makes sense.

Ethnical: ty!

Ethnical (approved):
LGTM let's wait for the green CI!

PR #1450, opened 2026-06-04
feat(MultisigTask): config-driven per-slot storage-code whitelist

Ethnical:
## Summary

Adds an opt-in, config-driven per-`(account, slot)` storage-code whitelist to `MultisigTask`, so a task can exempt a specific storage slot from the `isLikelyAddressThatShouldHaveCode` check in `_checkStateDiff`.

This PR is **framework-only** — just the mechanism in `MultisigTask.sol`. The U19 tasks that motivated it live in their own PR.

## The problem

`_checkStateDiff` treats the low 160 bits of every written slot as an address that must (or must not) have code. This is wrong for **packed slots** whose low bits are not an independent address.

Concrete case: `SystemConfig` slot `0x6c` packs `superchainConfig` (an address **with code**, low 160 bits) with `minBaseFee` (`uint64`, upper bits). Once `minBaseFee` is non-zero the full slot value exceeds `type(uint160).max`, the check takes its `else` branch, and `require(low160.code.length == 0)` reverts on the real `SuperchainConfig`:

```
Likely address in storage has unexpected code
  account: 0x...SystemConfigProxy
  slot:    0x...006c
  value:   0x0000000000000000000<minBaseFee><superchainConfig has code>
```

There was no escape hatch: `_getCodeExceptions()` is only consulted *inside* `isLikelyAddressThatShouldHaveCode`, which already returns `false` for packed slots and never reaches the exception list.

## The change

A new optional config section, parsed into `mapping(address => mapping(bytes32 => bool))` and checked in `_checkStateDiff`:

```toml
[storageCodeExceptions]
# SystemConfig slot 0x6c packs superchainConfig (has code) + minBaseFee
0xSystemConfigAddr = ["0x000000000000000000000000000000000000000000000000000000000000006c"]
```

```solidity
if (!_storageCodeExceptions[account][storageAccess.slot]) {
    if (isLikelyAddressThatShouldHaveCode(value, codeExceptions)) { require(code != 0) }
    else { require(code == 0) }
}
```

- Parser `_setStorageCodeExceptionsFromConfig` is called from `_overrideState`, alongside the existing `_setStateOverridesFromConfig`.
- Fully inert when a task has no `[storageCodeExceptions]` section — no behavior change for existing tasks.

## Why per-slot rather than skipping all packed slots

| | Skip all `value > uint160.max` | This whitelist |
|---|---|---|
| Scope | Every packed slot, every task, globally | Only the exact `(contract, slot)` listed |
| Safety | A bad address packed into upper bits slips through everywhere | All other packed slots still checked; opt out one slot at a time |
| Audit trail | None | Exception + rationale live in the task `config.toml`, reviewed with the task |

## Verification

- `forge build` passes.
- Exercised end-to-end on the U19 Unichain Sepolia task: without the exception the simulation reverts on `SystemConfig` slot `0x6c`; with a `[storageCodeExceptions]` entry it simulates cleanly. (That task is not part of this PR.)

### DIFF
--- a/src/tasks/MultisigTask.sol
+++ b/src/tasks/MultisigTask.sol
@@ -7,6 +7,7 @@ import {Script} from "forge-std/Script.sol";
 import {VmSafe} from "forge-std/Vm.sol";
 import {Test} from "forge-std/Test.sol";
 import {StdStyle} from "forge-std/StdStyle.sol";
+import {stdToml} from "forge-std/StdToml.sol";
 import {IMulticall3} from "forge-std/interfaces/IMulticall3.sol";
 
 import {Signatures} from "@base-contracts/script/universal/Signatures.sol";
@@ -29,6 +30,7 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager, TaskManage
     using AccountAccessParser for VmSafe.AccountAccess[];
     using AccountAccessParser for VmSafe.AccountAccess;
     using StdStyle for string;
+    using stdToml for string;
 
     /// @notice Maximum gas limit for transactions, 15M gas is ~90% of the limit
     uint256 internal constant MAX_GAS_LIMIT = 15_000_000;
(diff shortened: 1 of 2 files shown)


Wazabie (dismissed):

src/tasks/MultisigTask.sol
```diff
@@ -469,18 +478,23 @@ abstract contract MultisigTask is Test, Script, StateOverrideManager, TaskManage
...
-                } else {
-                    // Log account, slot, and value if there is code.
-                    // forgefmt: disable-start
-                    string memory err = string.concat("Likely address in storage has unexpected code\n", "  account: ", vm.toString(account), "\n  slot:    ", vm.toString(storageAccess.slot), "\n  value:   ", vm.toString(bytes32(value)));
-                    // forgefmt: disable-end
-                    require(address(uint160(value)).code.length == 0, err);
+                // Skip the code check for slots explicitly whitelisted via the task's
+                // `[storageCodeExceptions]` config (e.g. packed slots whose low 160 bits
+                // are not an independent address).
+                if (!_storageCodeExceptions[account][storageAccess.slot]) {
```
mds1, line 484: Could we add a few tests for this new functionality?

Ethnical: Ty added in 2bbbf503053a727e43356160e0ad3bd0a765ca2a

mds1 (approved):

PR #1453, opened 2026-06-05
fix(SetDisputeGameImpl): zero initBonds(8) on CANNON_KONA disable path

Wazabie:
## Summary

The kona disable branch in `SetDisputeGameImpl._setKonaImplementation` (`impl == 0`) cleared `gameImpls[CANNON_KONA]` and `gameArgs[CANNON_KONA]` but left `initBonds[CANNON_KONA]` untouched. Two compounding causes:

1. The disable branch **early-returns** before the bond-setting block (`return;` at the old line 150).
2. The bond-setting block was guarded by `k.bond != 0` — i.e. `bond = 0` in TOML is the "skip" sentinel used across this template (see [091/config.toml:52](https://github.com/ethereum-optimism/superchain-ops/blob/main/src/tasks/sep/091-alpha-set-dispute-game-impl-permissionless/config.toml#L52) for the same convention on PDG). So even without the early-return, you could not zero a bond from the TOML.

The post-disable validator (`_validateKona`) also only asserted that `gameArgs` was empty, so a residual `initBonds[CANNON_KONA]` passed simulation cleanly.

## Live evidence

Observed today on **karst-u19-beta-0** (Sepolia, DGF `0x313bcDa9f2D76Bab3554436a434A41779dAC74d0`) after [task 095](https://github.com/ethereum-optimism/superchain-ops/tree/main/src/tasks/sep/095-beta-0-zero-cannon-kona-game-type) ran:

| field | observed | expected |
|---|---|---|
| `gameImpls(8)` | `0x0` | `0x0` ✅ |
| `gameArgs(8)` | `0x` | `0x` ✅ |
| `initBonds(8)` | **`80000000000000000` (0.08 ETH)** | **`0`** ❌ |

The slot is inert (no one can call `create(8, …)` with `gameImpls(8) == 0`), so this is *functionally* harmless. But it's dead state that diverges from the disable path's intent and from what the task config thought it was producing (the task set `bond = 0` in the kona section assuming that meant \"set to zero\" — it meant \"skip\").

## The fix

- `_setKonaImplementation` disable branch: require `k.bond == 0` in the supplied config (consistency with the existing `k.gameArgs.length == 0` requirement), and call `factory.setInitBond(CANNON_KONA, 0)` when the on-chain bond is non-zero (idempotent otherwise).
- `_validateKona` disable case: also require `factory.initBonds(CANNON_KONA) == 0`.

Both changes are scoped to the disable path; the wire/update path is unchanged.

## Out of scope (worth follow-ups)

- The same \"`bond = 0` means skip\" sentinel applies to `_setInitBonds` for FDG and PDG (`SetDisputeGameImpl.sol:188-195`). If anyone ever wants to zero an FDG/PDG bond via this template, the same gap exists there. I left it untouched because there's no live evidence of it biting, and changing the semantics of `fdgBond/pdgBond = 0` would be a behaviour break for existing in-flight tasks (e.g. 091, 095, 096 all use it as a deliberate \"don't touch\" sentinel).
- The sibling [`SetDisputeGameArgs`](https://github.com/ethereum-optimism/superchain-ops/blob/main/src/template/SetDisputeGameArgs.sol#L56) template has a proper `bool setBond` flag that disambiguates \"skip\" from \"set to 0\". If/when we migrate kona-disable tasks to `SetDisputeGameArgs`, that's the cleaner long-term home.

## Cleanup for beta-0 itself

This PR doesn't retroactively zero the residual bond on beta-0; that needs a one-shot tx (e.g. via `SetDisputeGameArgs` with `{chainId=420110023, gameType=8, impl=0x0, bond=0}` once that template is taught to call `setInitBond(_, 0)` on disable, or a direct `factory.setInitBond(8, 0)` PAO Safe tx). Filing separately.

## Test plan

- [x] `forge build src/template/SetDisputeGameImpl.sol --skip test` — compiles clean
- [ ] Existing regression tests (`testRegressionCallDataMatches_SetDisputeGameImpl{,_Kona}`) — they exercise only the wire/update path (impl != 0), so behaviour is unchanged. (Couldn't run them locally — pinned forks need archive RPC; CI will.)
- [ ] Consider adding a new example task under `test/tasks/example/sep/` that exercises the kona disable path so the regression suite picks it up going forward.


Ethnical (approved):
Nice catch!

PR #1458, opened 2026-06-11
feat: support Tenderly's new ?draft= simulation link format

sbvegan:
## Context

Tenderly shipped a breaking change to how the Simulator hydrates a pre-filled form from a URL ([Slack thread](https://oplabs-pbc.slack.com/archives/C053NFWLG9K/p1781002848689189)). The old query-param format we emitted (`?network=…&contractAddress=…&from=…&stateOverrides=…&rawFunctionInput=…`) **no longer applies state overrides or gas** — signers had to re-enter everything by hand, defeating the purpose of the printed link.

The [new format](https://tenderlydev.notion.site/Simulator-draft-links-draft-37b84be6a2e680bb9874f18a833c1997) encodes the whole simulation as a single JSON payload, base64url-encoded (no padding), in a `?draft=` query parameter:

```
https://dashboard.tenderly.co/<org>/<project>/simulator/new?draft=<base64url(json)>
```

```json
{
  "v": 1,
  "network": { "id": "<chainid>" },
  "row": {
    "contractAddress": "<to>",
    "from": "<from>",
    "inputDataType": "raw",
    "rawFunctionInput": "<calldata>",
    "gas": "<TENDERLY_GAS>",
    "stateOverrides": [ { "contractAddress": "0x…", "balance": "", "storage": [ { "key": "0x…", "value": "0x…" } ] } ]
  }
}
```

## What changed

- **`src/libraries/MultisigTaskPrinter.sol`** — new `getTenderlySimulationLink(...)` builds the draft JSON `row`/payload and encodes it with the already-imported solady `Base64.encode(data, /*fileSafe*/ true, /*noPadding*/ true)`. `printTenderlySimulationData` now emits the draft link instead of the old query-param URL. (`Simulation.logSimulationLink` in the `base-contracts` submodule is left untouched; its only remaining callers are archived rehearsal scripts.)
- **Length fallback** — Tenderly's new format caps usable links at ~2000 chars (the old cap was 7980). Since real tasks wrap a Safe `execTransaction` (large calldata), the link is built **without** `rawFunctionInput` and the calldata is printed separately under `Insert the following hex into the 'Raw input data' field:` for manual paste. This mirrors the prior long-URL behavior and is the common path for production tasks.
- **`test/libraries/TenderlySimulationLink.t.sol`** — unit tests pinning the exact draft payload for the inline-calldata, no-raw-input (fallback), and no-gas paths, plus a guard that the legacy `network=`/`stateOverrides=%5B` params are gone.
- **Docs** — `runbooks/key-handover.md` and `docs/simulate-l2-ownership-transfer.md` updated to the new format.

## How it was tested

**Unit tests** (deterministic; env-sensitive assertions kept in one function to avoid the shared-`TENDERLY_GAS` race):
```
forge test --match-path "test/libraries/TenderlySimulationLink.t.sol" -vv
```

**Live Sepolia simulations** (against a real RPC):
- **Single task** — `086-U19-op-betanet-permissioned`, forked at block `10971228` (one before its on-chain execution tx, since 086 is already EXECUTED). Printed a valid draft link + fallback hex.
- **Stacked tasks** — the PR #1440 stack `[099-U19-op-ink-mmz-soneium, 100-U19-unichain]` via `just simulate-stack sep 100-U19-unichain`. Both tasks printed valid draft links.

Each produced link was base64url-decoded and confirmed to be valid JSON with the correct `v`/`network.id`/`contractAddress`/`from`/`gas`/`stateOverrides`.

> Note on stacked links: a downstream task's link can't reproduce the *chain-state* mutations made by earlier tasks in the stack (only Safe threshold/owner/nonce travel in the overrides). This is the pre-existing "stacked Tenderly links may not work" limitation the `simulate-stack` recipe already warns about — not introduced by this change.

## How a reviewer can validate

1. **Run the unit tests:** `forge test --match-path "test/libraries/TenderlySimulationLink.t.sol" -vv` (4 cases, should pass).
2. **Simulate a task** and inspect the printed `Simulation link:` — confirm it is a `…/simulator/new?draft=<b64url>` URL (no legacy query params).
3. **Decode the draft** to confirm the payload is correct:
   ```bash
   python3 -c 'import base64,json,sys; d=sys.argv[1]; print(json.dumps(json.loads(base64.urlsafe_b64decode(d+"="*(-len(d)%4))),indent=2))' "<paste the value after ?draft=>"
   ```
   Expect `v:1`, `network.id` = the chain id, and the right `contractAddress`/`from`/`gas`/`stateOverrides`.
4. **Open the link in Tenderly** (replacing `TENDERLY_USERNAME`/`TENDERLY_PROJECT` with your own org/project, or set the `TENDERLY_*` env vars before simulating) and confirm the form pre-fills network, contract, from, gas, and state overrides. For large-calldata tasks, paste the separately-printed hex into "Raw input data".


docs/simulate-l2-ownership-transfer.md
```diff
@@ -19,7 +19,7 @@ This template executes an L1 transaction, which is later forwarded to the L2 by
         }
       ```
 
-4. Using the data in the `TransactionDeposited` event, create a new Tenderly simulation via this [link](https://dashboard.tenderly.co/TENDERLY_USERNAME/TENDERLY_PROJECT/simulator/new?block=&blockIndex=0&from=0x0000000000000000000000000000000000000000&gas=8000000&gasPrice=0&value=0&contractAddress=&headerBlockNumber=&headerTimestamp=).
+4. Using the data in the `TransactionDeposited` event, create a new Tenderly simulation by opening the [Simulator](https://dashboard.tenderly.co/TENDERLY_USERNAME/TENDERLY_PROJECT/simulator/new) and filling in the form using the fields below. Set `from` to `0x0000000000000000000000000000000000000000`, `gas` to `8000000`, `gas price` and `value` to `0`.
```
Ethnical: This line looks weird to have the to set to addr(0) first review I didn't checked more for now.  
```
to 0x0000000000000000000000000000000000000000
```

src/libraries/MultisigTaskPrinter.sol
```diff
@@ -97,7 +101,97 @@ library MultisigTaskPrinter {
...
         printTitle("TENDERLY SIMULATION DATA");
         console.log("\nSimulation link:");
-        Simulation.logSimulationLink({_to: targetAddress, _data: finalExec, _from: sender, _overrides: overrides});
+
+        // First attempt: a single draft link with the full calldata embedded.
+        string memory link = getTenderlySimulationLink(targetAddress, finalExec, sender, overrides, true);
+        if (bytes(link).length <= TENDERLY_MAX_URL_LENGTH) {
+            console.log(link);
+            return;
+        }
```
Ethnical, line 110: I think we should have an else > 2000 (even unlikely if it's happening we should return something to say that the URL is longer than 2000. 
That would be confusing to have the output but link at all.

docs/simulate-l2-ownership-transfer.md
mds1: How come now you have to manually set the from, gas, etc? The tenderly URLs no longer let you auto-populate those?

src/libraries/MultisigTaskPrinter.sol
```diff
@@ -97,7 +101,97 @@ library MultisigTaskPrinter {
...
+
+        // The draft format caps usable links at ~2000 characters and the calldata is too large to
```
mds1, line 112: What does "draft" mean? I see this is what tenderly calls it but I don't know why. Also nit but without the notion doc in the PR description no one will know what "draft" means, so we should link to the notion doc and explain that in the code comments somewhere

docs/simulate-l2-ownership-transfer.md
Ethnical: My bad, this is seems to be only for the simulate-l2-ownership-transfer task (which is weird because the gas + from is correctly populated actually!). 

```wrap
{"v":1,"network":{"id":"1"},"row":{"contractAddress":"0xcA11bde05977b3631167028862bE2a173976CA11","from":"0xf13D09eD3cbdD1C930d4de74808de1f33B6b3D4f","inputDataType":"raw","gas":"10000000","stateOverrides":[{"contractAddress":"0x5a0Aae59D09fccBdDb6C6CcEB07B7279367C3d2A","balance":"","storage":[{"key":"0x0000000000000000000000000000000000000000000000000000000000000004","value":"0x0000000000000000000000000000000000000000000000000000000000000001"},{"key":"0x0000000000000000000000000000000000000000000000000000000000000005","value":"0x0000000000000000000000000000000000000000000000000000000000000023"}]},{"contractAddress":"0x847B5c174615B1B7fDF770882256e2D3E95b9D92","balance":"","storage":[{"key":"0x0000000000000000000000000000000000000000000000000000000000000004","value":"0x0000000000000000000000000000000000000000000000000000000000000001"},{"key":"0x0000000000000000000000000000000000000000000000000000000000000003","value":"0x0000000000000000000000000000000000000000000000000000000000000001"},{"key":"0xe90b7bceb6e7df5418fb78d8ee546e97c83a08bbccc01a0644d599ccd2a7c2e0","value":"0x000000000000000000000000ca11bde05977b3631167028862be2a173976ca11"},{"key":"0x316a0aac0d94f5824f0b66f5bbe94a8c360a17699a1d3a233aafcf7146e9f11c","value":"0x0000000000000000000000000000000000000000000000000000000000000001"},{"key":"0x0000000000000000000000000000000000000000000000000000000000000005","value":"0x0000000000000000000000000000000000000000000000000000000000000039"}]},{"contractAddress":"0xc2819DC788505Aac350142A7A707BF9D03E3Bd03","balance":"","storage":[{"key":"0x0000000000000000000000000000000000000000000000000000000000000005","value":"0x000000000000000000000000000000000000000000000000000000000000003a"}]}]}
```

Ethnical (approved):
LGTM! 

Image

validated on eth/53!

PR #1469, opened 2026-06-16
fix(FeeVaultUpgradeTemplate): configure fee vaults via setters, not initialize

Wazabie:
## Problem

`FeeVaultUpgradeTemplate` upgrades each fee-vault predeploy with `ProxyAdmin.upgradeAndCall(proxy, impl, initialize(recipient, minWithdrawal, network))`.

On any chain whose fee vaults are **already initialized in L2 genesis** (the normal case for live OP Stack chains), this **silently no-ops on L2 while the L1 task tx succeeds**:

- The fee-vault `initialize` uses OpenZeppelin v5's `initializer` modifier (version 1, not a bumped `reinitializer`). The proxies already have `_initialized == 1`, so `initialize` reverts with `InvalidInitialization()` (`0xf92ee8a9`).
- The OP `Proxy.upgradeToAndCall` `require`s the post-upgrade delegatecall to succeed, masking the inner error as `"Proxy: delegatecall to new implementation contract failed"`.
- `upgradeAndCall` reverts → the impl-slot write rolls back → the recipient never changes.
- The L1 portal deposits are still accepted, so the L1 task tx succeeds — the failure is invisible from L1.

The template's pre-flight only checks L2 ProxyAdmin ownership, not the already-initialized case, so this is a confusing land-on-L1 / silently-fail-on-L2 mode.

### Reproduced on-chain (Sepolia devnet `migration-src-0`, chainId 420120140)
- New impls deploy fine; L2 ProxyAdmin is owned by the aliased PAO ✓ (so it is *not* the ownership pre-flight issue)
- Replaying `upgradeAndCall` on L2 reverts with `Proxy: delegatecall to new implementation contract failed`
- Inner error decodes to `InvalidInitialization()`; all four vault proxies read `_initialized = 1` at the OZ-v5 namespaced slot

## Fix

Deposit `ProxyAdmin.upgrade(proxy, impl)` (no initializer call), then configure via the v1.6.0 storage setters `setRecipient` / `setMinWithdrawalAmount` / `setWithdrawalNetwork`. Each setter is gated by the L2 ProxyAdmin owner — the aliased rootSafe these portal deposits already execute as — so it works whether or not the proxy was previously initialized. Same approach as the existing `RevShareCommon.configureVaultsForRevShare` helper.

Per vault: **1 `upgrade` + 3 setters** (was 1 `upgradeAndCall`); `_validate`'s expected action count updated accordingly.

## Verification
- `forge build` ✓
- Regression fixtures in `testRegressionCallDataMatches_FeeVaultUpgradeTemplate` regenerated for the new deposit pattern: `expectedCallData` is now 14 deposits (2 CREATE2 deploys + 3 `upgrade` + 9 setters), plus the nested-multisig `expectedDataToSign` for Foundation + Council. The regenerated data-to-sign **domain hashes match the previous pinned values exactly** (they are override-driven / block-independent), and the calldata was verified selector-by-selector (`upgradeAndCall` gone, `upgrade`×3, 14×`depositTransaction`). CI's archive-fork run at the pinned block is the authoritative regression check.

## Follow-ups (out of scope here)
Tasks that consume this template must be re-simulated and their VALIDATION hashes regenerated, since the generated calldata changes:
- `src/tasks/sep/085-migrations-sop-1-fee-vault-update`
- the in-flight `migration-src-0` fee-vault task ([#1463](https://github.com/ethereum-optimism/superchain-ops/pull/1463))


Wazabie:
Closing as the template has been already updated for new fee vaults with setters

PR #1471, opened 2026-06-17
fix(MultisigTask): let bare `just execute` work on 1/1 safes (broadcast nonce-bump guardrail)

Wazabie:
**Optional guard**, not required fix. The normal way to run tasks `just sign` → `just execute SIGNATURES=...` already works and is unaffected. This just makes a bare `just execute` (no signatures) also work on a **1/1 safe** where today it silently hangs.

It's a safety net, if whoever runs the task forgets to pass `SIGNATURES=...` (easy to do, since nested tasks are run with a bare `just execute`) it just executes correctly instead of silently hanging forever.

## Issue

When you run `just execute` with no signatures, `MultisigTask` does a **simulation only** and it bumps each owner's nonce by 1 to mimic an approval that would happen on-chain. On a **1/1 safe the only owner is the same EOA that's broadcasting the transaction**, so that bump lands on the broadcaster's own nonce. Forge then sends the real transaction at `n+1` and leaves a gap at `n`. Nothing fills `n` so the tx sits **pending forever**.

This only affects a 1/1 EOA safe run with a bare `just execute`:
- **Signed runs** (`SIGNATURES=...`) take a different code path that does do the bump.
- **Nested safes** also run a bare `just execute` (the approvals happen earlier, in `just approve`) but there the bump lands on the child *Safe's* storage, not on the broadcasting EOA, so it's harmless.

## Guard

Skip the simulation-only bump when we're actually broadcasting:

```solidity
if (!_isBroadcastContext()) {
    _incrementOwnerNonce(owners[i]);
}
```

In a real broadcast the bump is faking something that never happens on-chain, so skipping it just lets forge use the true nonce. The transaction still goes through (the signature is valid because the owner *is* the sender). **Simulations are unchanged** they still do the bump, which are needed.

## Testing

- Two regression tests, both fail without the guard: one covers the contract-owner case (`vm.store`), one covers the EOA-owner case (`vm.setNonce` — the one behind the hang).
- Reproduced on a Sepolia fork with the real `migration-src-0` task on the 1/1 ProxyAdminOwner (`0xe934Dc97E347C6aCef74364B50125bb8689c40ff`):

  | | broadcasts at | result |
  |---|---|---|
  | **without guard** | n+1 (4567) | gap at n → hangs forever |
  | **with guard** | n (4566) | mines (block 11073801) → rotation applied → done |


Wazabie:
Closing in favor of #1477.

Per review feedback, we'd rather have one way to run an execution than add a Solidity fallback that makes a bare `just execute` work. #1477 takes the justfile-guard approach instead: `just execute` now fails fast when `SIGNATURES` is unset/empty, so the silent-hang footgun is surfaced as an error rather than worked around. The explicit pre-approved path (`SIGNATURES=0x just execute` after `just approve`) is preserved for nested safes.

Superseded by #1477.

PR #1474, opened 2026-06-17
Add SystemConfigGasLimit template (gas-limit-only)

sbvegan:
## Summary

Adds **`src/template/SystemConfigGasLimit.sol`** — a minimal `L2TaskBase` template that calls only `SystemConfig.setGasLimit` (no EIP-1559 params). It's the gas-limit-only counterpart to the existing `SystemConfigGasParams`.

## Why

Re-issuing `setGasLimit` re-emits the gas `ConfigUpdate` without touching EIP-1559 params — useful whenever only the gas limit needs to change. The immediate motivation is the **Karst (U19) gas-limit reset**, where re-emitting the `ConfigUpdate` (to clear the post-activation +55 MGas op-node/kona bug) is the *entire* mitigation; the EIP-1559 params are unrelated.

## Details

- Reads `gasParams.gasLimit` from `config.toml` and applies it to each chain in `l2chains`.
- Validates `SystemConfig.gasLimit()` post-execution.
- The gas limit is authorized by the SystemConfig owner, so `safeAddressString` defaults to `"SystemConfigOwner"` (resolved per-chain from `SystemConfig.owner()`); override in `config.toml` if a different root safe is required.
- Does not modify the shared `SystemConfigGasParams` template — fully additive.

## Status / follow-ups

Split out from the U19 gas-reset tasks PR (#1470) so the template can be reviewed independently. No example task / unit test is included yet — happy to add one mirroring `test/tasks/example/eth/006-system-config-gas-params` + the `SystemConfigGasParams` entries in `TaskManager.t.sol` / `Regression.t.sol` if desired before merge.


smartcontracts (approved):

smartcontracts:
@OptimismBot please review

OptimismBot (commented):
**Review: LGTM (non-blocking).** Clean, fully-additive gas-limit-only template that correctly mirrors `SystemConfigGasParams` minus the EIP-1559 calls.

Verified:
- `SystemConfigGasLimit.sol` compiles, and `_build`/`_validate` correctly issue and check only `setGasLimit` — calldata in the new regression test decodes to a single `setGasLimit(40_000_000)` (selector `0xb40a817c`, arg `0x2625a00`), no `setEIP1559Params`.
- `safeAddressString() = "SystemConfigOwner"` is a registered registry key (`SuperchainAddressRegistry.sol:323`) and is already used as a root-safe string elsewhere (task 082). Resolving per-chain via `SystemConfig.owner()` is more general than hardcoding `FoundationUpgradeSafe`; for OP Mainnet both resolve to `0x847B…9D92`, consistent with the test.
- No central registry needs updating — `TaskManager` resolves templates by compiled-artifact name (`out/<templateName>.sol/...`), so the template is wired just by existing.

Notes (non-blocking):
- As you called out, there's no example task / unit-test coverage beyond the regression case (no `TaskManager.t.sol` entry, no VALIDATION.md). The added `Regression.t.sol` calldata + data-to-sign assertion is the meaningful safety check here, so this is fine to land as-is; adding the example mirror would be a nice follow-up.
- I could not execute the mainnet fork regression test in this environment (no archive RPC / full submodules), so the `expectedCallData`/`expectedDataToSign` values are trusted to CI rather than independently reproduced.

PR #1482, opened 2026-06-23
Karst gas-limit reset — Unichain Mainnet (060)

sbvegan:
## Context

Karst (U19) has an op-node/kona-node bug that leaves the effective L2 block gas limit ~55 MGas above the value stored in the SystemConfig after the activation block (see Notion *"P/PS: Karst Gas Limit Bug"*). The approved mitigation is a per-chain **SystemConfig transaction** that re-emits the gas `ConfigUpdate`, resetting the effective gas limit back to the on-chain value.

Mainnet Karst activates **2026-07-08**. This task runs *after* activation.

## What this adds

Reference task for **Unichain Mainnet** re-affirming the chain's current on-chain gas limit via `SystemConfig.setGasLimit`. Uses the `SystemConfigGasLimit` template (setGasLimit only — no EIP-1559 params).

| Task | Chain | gasLimit |
|---|---|---|
| `060-gas-limit-unichain` | Unichain | 60M |

Unichain's SystemConfig is owned by the **operator Safe** (`0x9245d5D10AA8a842B31530De71EA86c0760Ca1b1`), not the Foundation Upgrade Safe. This is a reference task for the operator to execute independently of the FUS. Nonce override is **9** (verify live nonce before signing).

## Validation

Simulates cleanly via `just simulate-stack eth 060-gas-limit-unichain`. The recorded hashes in `VALIDATION.md` are confirmed against simulation output.

## Notes

- Only `setGasLimit` is called (no `setEIP1559Params`) — the `SystemConfigGasLimit` template is used.
- Unichain already has `gasLimit = 60000000` on chain; the call re-emits the `ConfigUpdate` event without changing storage.


src/tasks/eth/060-gas-limit-unichain/README.md
```diff
@@ -0,0 +1,40 @@
...
+> `cast call 0x9245d5D10AA8a842B31530De71EA86c0760Ca1b1 "nonce()(uint256)" --rpc-url mainnet`
+
+## Simulation & Signing
+
+```bash
+# Simulate
+just simulate-stack eth 060-gas-limit-unichain
+
+# Sign
+USE_KEYSTORE=1 just sign-stack eth 060-gas-limit-unichain
```
sbvegan:
```suggestion
just sign-stack eth 060-gas-limit-unichain
```

PR #1462, opened 2026-06-12
feat: Ink SystemConfig owner transfers to FoundationOperationsSafe (eth/055, sep/101)

Wazabie:
## Description

Transfers ownership of the Ink `SystemConfigProxy` on both networks from the current Ink-controlled owner Safe to the Optimism **FoundationOperationsSafe**, using the existing `TransferSystemConfigOwnership` template (one chain per task):

| Task | Chain | SystemConfigProxy | Current owner (signer) | New owner (FOS) |
|---|---|---|---|---|
| `eth/055-ink-transfer-system-config-owner` | Ink (57073) | [`0x62C0a111929fA32ceC2F76aDba54C16aFb6E8364`](https://etherscan.io/address/0x62C0a111929fA32ceC2F76aDba54C16aFb6E8364) | `0xBeA2Bc852a160B8547273660E22F4F08C2fa9Bbb` | [`0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A`](https://etherscan.io/address/0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A) |
| `sep/101-ink-transfer-system-config-owner` | Ink Sepolia (763373) | [`0x05C993e60179f28bF649a2Bb5b00b5F4283bD525`](https://sepolia.etherscan.io/address/0x05C993e60179f28bF649a2Bb5b00b5F4283bD525) | `0xBeA2Bc852a160B8547273660E22F4F08C2fa9Bbb` | [`0x837DE453AD5F21E89771e3c06239d8236c0EFd5E`](https://sepolia.etherscan.io/address/0x837DE453AD5F21E89771e3c06239d8236c0EFd5E) |

### Input verification

- Current owners verified on-chain: `cast call <SystemConfigProxy> "owner()(address)"` returns `0xBeA2Bc852a160B8547273660E22F4F08C2fa9Bbb` on both networks (the same 3-of-5 Safe, v1.3.0, identical owner set on both networks).
- `SystemConfigProxy` addresses verified against the superchain-registry ([mainnet/ink.toml](https://github.com/ethereum-optimism/superchain-registry/blob/main/superchain/configs/mainnet/ink.toml), [sepolia/ink.toml](https://github.com/ethereum-optimism/superchain-registry/blob/main/superchain/configs/sepolia/ink.toml)).
- New owners match `FoundationOperationsSafe` in `src/addresses.toml` (`[eth]` / `[sep]`).
[...]

PR #1490, opened 2026-07-07
Ink Mainnet Migration

Wazabie:
## Ink Mainnet Migration (Gelato → OP Enterprise) — on-chain cutover tasks

Adds the two L1 tasks for the Ink mainnet (chainId 57073) migration. Mainnet counterparts of the executed Ink Sepolia
rehearsal tasks `sep/101` / `sep/102`.

- **`eth/060-ink-set-batcher-unsafe-signer`** — cutover steps 2 & 3.
  `SetBatcherAndOrSigner`: registers the OPE batcher + unsafe block signer on
  `SystemConfigProxy` in one Multicall3. Signed by the OPE Safe (new SystemConfig
  owner after Gelato's off-repo `transferOwnership`, W25/W26).
- **`eth/061-ink-proposer-rotation`** — cutover step 4.
  `SetDisputeGameArgs`: field-diffs **only the proposer** on the
  PermissionedDisputeGame (gameType 1, dormant fallback). Challenger left unchanged
  (already OP governance). Nested-signed by the L1PAO mainnet 2-of-2 (FUS + SC).

No fee-vault task included as it's not planned for the migration


src/tasks/eth/060-ink-set-batcher-unsafe-signer/README.md
```diff
@@ -0,0 +1,60 @@
...
+| `batcherHash()` | `0x000000000000000000000000500d7ea63cf2e501dadaa5feec1fc19fe2aa72ac` | `0x0000000000000000000000006db6161fc5662450e801398bad62dd9921216b98` |
+| `unsafeBlockSigner()` | `0x7D056B99AA2021864c42E25B4F8cE3BdEAc9463C` | `0x7b322282DF45E537E5de76D60E1432Db3cF3F8E1` |
+
+Plus the FoundationOperationsSafe nonce increments `118` → `119`. (The `SystemConfig.owner()` slot-`0x33` change is a **simulation-only override**, not a state change produced by this task — see the note above.)
+
+## Simulation & Signing
+
+```bash
+cd src/tasks/eth/060-ink-set-batcher-unsafe-signer
+SIMULATE_WITHOUT_LEDGER=1 just --dotenv-path $(pwd)/.env --justfile ../../../justfile simulate
```
mds1, line 31: Can we add the two `.env` files before marking these tasks `READY TO SIGN`? Both READMEs invoke `--dotenv-path $(pwd)/.env`, and the task scaffold/review workflow requires `TENDERLY_GAS`. Without it, the generated Tenderly payload omits an explicit gas limit, so we cannot complete the task-level gas check. The corresponding rehearsal tasks use `TENDERLY_GAS=15000000`.

src/tasks/eth/060-ink-set-batcher-unsafe-signer/README.md
```diff
@@ -0,0 +1,60 @@
...
+## State Changes
+
+Writes to `SystemConfigProxy` ([`0x62C0a111…E8364`](https://etherscan.io/address/0x62C0a111929fA32ceC2F76aDba54C16aFb6E8364#readProxyContract)):
+
+| Field | Current (on-chain) | New |
+|-------|--------------------|-----|
+| `batcherHash()` | `0x000000000000000000000000500d7ea63cf2e501dadaa5feec1fc19fe2aa72ac` | `0x0000000000000000000000006db6161fc5662450e801398bad62dd9921216b98` |
+| `unsafeBlockSigner()` | `0x7D056B99AA2021864c42E25B4F8cE3BdEAc9463C` | `0x7b322282DF45E537E5de76D60E1432Db3cF3F8E1` |
+
+Plus the FoundationOperationsSafe nonce increments `118` → `119`. (The `SystemConfig.owner()` slot-`0x33` change is a **simulation-only override**, not a state change produced by this task — see the note above.)
```
mds1: Can we update both nonce references to the stacked value `120 → 121`? The config, simulation, and `VALIDATION.md` all use FOS nonce `120`, derived from live nonce `118` plus tasks `058` and `059`. The README still says `118 → 119`, which describes a different Safe hash.

src/tasks/eth/061-ink-proposer-rotation/README.md
```diff
@@ -0,0 +1,77 @@
+# 061-ink-proposer-rotation
+
+Status: READY TO SIGN
+
+Simulated against live state (signer nonces L1PAO=36 / FUS=60 / SC=60; live `initBonds(1)` = 0.08 ETH confirmed). The proposer-only `gameArgs(1)` diff is confirmed (Gelato proposer fully replaced, challenger preserved); the nested signer domain/message/safe hashes in [VALIDATION.md](./VALIDATION.md) match `just simulate council` / `just simulate foundation` and only move if a signer-safe nonce advances.
```
mds1: Can we change `FUS=60` to the stacked value `FUS=62`? The recorded Foundation hash uses nonce `62`, derived from live nonce `60` plus tasks `056` and `057`. `VALIDATION.md` and `config.toml` already say `62`.

src/tasks/eth/060-ink-set-batcher-unsafe-signer/config.toml
```diff
@@ -0,0 +1,50 @@
...
+# The new SystemConfig owner (FoundationOperationsSafe). A custom identifier is used instead
+# of the registry-resolved "SystemConfigOwner", which still reads the live Gelato owner
+# (0xBeA2Bc…9Bbb) until the ownership transfer executes on-chain.
+safeAddressString = "OPESystemConfigOwnerSafe"
+
+l2chains = [
+    {name = "Ink", chainId = 57073}
+]
+
+# OPE receiving infrastructure (Ink Mainnet Migration Plan §3.3 Addresses).
```
mds1: Can we link the exact source files here? I independently verified all three values against `k8s-netchef-prod@7115c120`: [batcher config](https://github.com/ethereum-optimism/k8s-netchef-prod/blob/7115c12018a3cd0e71d08ffa0cc8c3969dc437ec/opc/oplabs-prod-ent-mainnet-eu/mn-ink-mainnet-0-op-batcher/values.yaml), [op-signer key mapping](https://github.com/ethereum-optimism/k8s-netchef-prod/blob/7115c12018a3cd0e71d08ffa0cc8c3969dc437ec/kustomize/op-signer/oplabs-prod-ent-mainnet-eu/mn-ink-mainnet-0-op-signer/configmap.yaml), and [proposer config](https://github.com/ethereum-optimism/k8s-netchef-prod/blob/7115c12018a3cd0e71d08ffa0cc8c3969dc437ec/opc/oplabs-prod-ent-mainnet-eu/mn-ink-mainnet-0-op-proposer/values.yaml). The generic `Migration Plan §3.3` reference does not let reviewers reproduce the input verification.

src/tasks/eth/060-ink-set-batcher-unsafe-signer/README.md
```diff
@@ -0,0 +1,60 @@
...
+| Field | Current (Gelato) | New (OPE) |
+|-------|------------------|-----------|
+| `batcherHash()` | `0x00…500d7Ea63CF2E501dadaA5feeC1FC19FE2Aa72Ac` | `0x00…6db6161fC5662450E801398Bad62dD9921216B98` |
+| `unsafeBlockSigner()` | `0x7D056B99AA2021864c42E25B4F8cE3BdEAc9463C` | `0x7b322282DF45E537E5de76D60E1432Db3cF3F8E1` |
+
+- **Target**: `SystemConfigProxy` `0x62C0a111929fA32ceC2F76aDba54C16aFb6E8364` (resolved from the superchain-registry).
+- **Signer**: `FoundationOperationsSafe` `0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A` — the SystemConfig owner after the Gelato → FOS ownership transfer.
+
+> [!IMPORTANT]
+> This task assumes the `SystemConfig` owner is the **FoundationOperationsSafe**. On Ink mainnet the owner is currently the **Gelato Safe** (`0xBeA2Bc852a160B8547273660E22F4F08C2fa9Bbb`, verified on-chain); the Gelato → FOS `transferOwnership` is performed **outside this repo** (task [`eth/055`](../055-ink-transfer-system-config-owner) / PR #1462, Migration Plan steps **W25/W26**). For simulation, [config.toml](./config.toml) overrides `SystemConfig.owner()` (slot `0x33`) to the FOS. **Remove that override once the ownership transfer is executed on-chain.**
```
mds1: Can we link directly to PR #1462 until the prerequisite task lands? `../055-ink-transfer-system-config-owner` does not exist at this PR head—`eth/055` is currently `055-gas-limit-op`. PR #1462 is still open and its transfer task says `DRAFT, NOT READY TO SIGN`.

src/tasks/eth/060-ink-set-batcher-unsafe-signer/config.toml
```diff
@@ -0,0 +1,50 @@
...
+#
+# Signed by the NEW SystemConfig owner — the FoundationOperationsSafe (FOS). The Gelato →
+# FOS SystemConfig.transferOwnership is executed OUTSIDE this repo (see eth/055 / PR #1462,
+# warm-phase step W25/W26); this task assumes that transfer has landed. For standalone
+# simulation the [stateOverrides] below set SystemConfig.owner() (slot 0x33) to the FOS so
+# the setters authorize. Remove that owner override once the transfer is executed on-chain.
+
+templateName = "SetBatcherAndOrSigner"
+
+# The new SystemConfig owner (FoundationOperationsSafe). A custom identifier is used instead
```
mds1, line 15: Can you remind me where/why we chose the FOS here and not the FUS? We should have that decision documented somewhere for posterity

src/tasks/eth/060-ink-set-batcher-unsafe-signer/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+> [!CAUTION]
+> Per plan step **W4**, the OPE batcher and unsafe-block-signer addresses were independently verified by ≥3 OP Labs engineers. **Mainnet keys are NEW — not reused from the Ink Sepolia rehearsal.**
```
mds1: Who were the 3 engineers / where is their verification attested to?

src/tasks/eth/060-ink-set-batcher-unsafe-signer/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+## State Changes
```
mds1: Have you reviewed the state diff in tenderly to verify this is the full set of changes?

src/tasks/eth/060-ink-set-batcher-unsafe-signer/README.md
mds1: An even better approach to prevent mistakes like this and the below comment is to simply not write what the nonce is in so many different places, and just let the config file speak to what the current nonce is and why. For example here you can just say "the nonce increments from the value in the config file"

src/tasks/eth/060-ink-set-batcher-unsafe-signer/README.md
mds1: Nit, but these state changes sections are redundant against from whats in the validation file so we probably should just delete them?

src/tasks/eth/060-ink-set-batcher-unsafe-signer/README.md
sbvegan: added!

src/tasks/eth/060-ink-set-batcher-unsafe-signer/config.toml
sbvegan: I'm pretty sure because that lines up with OP Mainnet

src/tasks/eth/060-ink-set-batcher-unsafe-signer/README.md
sbvegan: This is from our SOP and I don't think we've actually documented the attestations. But I know that I have, Zach, and JP have all pointed at the same addresses

src/tasks/eth/060-ink-set-batcher-unsafe-signer/README.md
sbvegan: I have, outside of the safes it looks like it just changes the batcher hash, unsafe signer, and the proposer (packed in the game args) are the only changes

src/tasks/eth/061-ink-proposer-rotation/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+Cutover step 4. Signatures collected in the warm phase (plan W21/W22). **Non-blocking** — does not hold up the sequencer switch (fallback game is dormant; active game is permissionless type 8), but must be confirmed before cutover step 10 (OPE proposer start). Independent of the SystemConfig owner transfer (signed by the already-OP-governed L1PAO).
+
+## Simulation & Signing
+
+```bash
+cd src/tasks/eth/061-ink-proposer-rotation
+SIMULATE_WITHOUT_LEDGER=1 just --dotenv-path $(pwd)/.env --justfile ../../../justfile simulate
+# nested signing flow — see docs/NESTED.md:
+just --dotenv-path $(pwd)/.env --justfile ../../../justfile sign
```
mds1: These commands target the root `ProxyAdminOwner` Safe, not either nested signer. I ran the command as written against this head: it produced root Safe hash `0xff21287d…f196bc`, while `VALIDATION.md` records Security Council `0x06a7c8c2…c5937f` and FUS `0x7c40d5c6…d21b27`. The bare `sign` path also checks the Ledger EOA as a direct owner of the root Safe, whose direct owners are the child Safes. Can we list the four explicit commands here: `simulate council`, `simulate foundation`, `sign council`, and `sign foundation`?

src/tasks/eth/061-ink-proposer-rotation/config.toml
```diff
@@ -0,0 +1,61 @@
...
+#   https://github.com/ethereum-optimism/k8s-netchef-prod/blob/7115c12018a3cd0e71d08ffa0cc8c3969dc437ec/opc/oplabs-prod-ent-mainnet-eu/mn-ink-mainnet-0-op-proposer/values.yaml
+# Cross-referenced with Ink Mainnet Migration Plan §3.3 Addresses; independently verified by ≥3 OP Labs engineers (plan step W4).
+#
+# Single row: PDG (game type 1), proposer only. Every omitted field is read from the live
+# gameArgs(1) and preserved.
+[[gameConfig]]
+chainId  = 57073
+gameType = 1                                                  # PERMISSIONED_CANNON
+proposer = "0x3832bfbeF03173E4C49a00ec0DD178817A02D177"       # OPE proposer (verified W4). Live pre-migration Gelato proposer (for rollback): 0x65436DDcBc026F34118954f229F7f132b696B3b4.
+bond     = 80000000000000000                                  # 0.08 ETH — pins the CURRENT PDG init bond (confirmed live `initBonds(1)` on mainnet DGF, matches the Sepolia rehearsal). No-op that lets _validate assert it. RE-READ live before signing; a value of 0 here would actively RESET the bond to zero.
```
mds1: Why pin `bond` instead of omitting it? `SetDisputeGameArgs` treats a present `bond` as `setBond=true` and calls `setInitBond` if the live value changes. At the current `0.08 ETH` value the Foundry diff is proposer-only, but a bond change before cutover makes this task reset it to `0.08 ETH`, contrary to the comments that every non-proposer field is read live and preserved. Can we omit `bond` here? If pinning is intentional, let's document bond drift as a blocker and stop describing the task as unconditionally proposer-only.

src/tasks/eth/061-ink-proposer-rotation/README.md
```diff
@@ -0,0 +1,54 @@
+# 061-ink-proposer-rotation
+
+Status: READY TO SIGN
+
+Simulated against the stacked signer nonces recorded in [config.toml](./config.toml) (L1PAO=36 / FUS=62 / SC=60; live `initBonds(1)` = 0.08 ETH confirmed). The proposer-only `gameArgs(1)` diff is confirmed (Gelato proposer fully replaced, challenger preserved); the nested signer domain/message/safe hashes in [VALIDATION.md](./VALIDATION.md) match `just simulate council` / `just simulate foundation` and only move if a signer-safe nonce advances.
```
mds1: `only move if a signer-safe nonce advances` is too narrow. A child Safe signs approval of the root L1PAO transaction hash, so its hash also changes if the L1PAO nonce or any live value used to rebuild `gameArgs(1)` changes. A bond drift can also add `setInitBond` because `bond` is pinned in the config. Can we say to regenerate hashes whenever the root nonce, child nonce, or live task payload changes? `VALIDATION.md:31` needs the same update.

src/tasks/eth/061-ink-proposer-rotation/config.toml
mds1: (I don't think this should matter because we aren't going to change the bond amount, but it's a good find. I'm ok just doing whatever we did for sepolia)

src/tasks/eth/061-ink-proposer-rotation/README.md
mds1: (similar to my above comment, I don't think we need to worry about bond drift and should do what we did for sepolia)

src/tasks/eth/060-ink-set-batcher-unsafe-signer/config.toml
sbvegan: We checked and it does not map to OPM, we'll need to document this somewhere and align on how we want to handle it

src/tasks/eth/061-ink-proposer-rotation/config.toml
sbvegan: I don't know if the template requires all the args or not. My guess is that it doesn't because the absolute prestate isn't covered here. But this is what was done on Ink sepolia: https://github.com/ethereum-optimism/superchain-ops/blob/main/src/tasks/sep/102-ink-sepolia-proposer-rotation/config.toml

src/tasks/eth/061-ink-proposer-rotation/README.md
sbvegan: added

src/tasks/eth/061-ink-proposer-rotation/README.md
sbvegan: added

mds1 (approved):

PR #1493, opened 2026-07-08
[CONTINGENCY] Ink mainnet migration rollback task (eth/061)

Wazabie:
## Ink Mainnet Migration — ROLLBACK CONTINGENCY (`eth/061`)

A single reverse task to roll back the Ink mainnet (chainId 57073) Gelato → OP Enterprise **sequencer & batcher** cutover if a problem is detected after it lands. Direct companion to the forward task [`eth/060-ink-set-batcher-unsafe-signer`](https://github.com/ethereum-optimism/superchain-ops/pull/1490) (PR #1490, branch `af/ink-mainnet-migration`).

| Task | Reverses | Template | Signer | Restores |
|---|---|---|---|---|
| `061-ink-revert-batcher-unsafe-signer` | `eth/060` | `SetBatcherAndOrSigner` | FoundationOperationsSafe (single) | batcher `0x500d7Ea6…72Ac`, unsafe signer `0x7D05…463C` |

### What changed from the original version of this PR
- **Removed `062-ink-revert-system-config-owner`** — SystemConfig ownership stays with the FoundationOperationsSafe through this rollback; we are **not** handing ownership back to Gelato as part of the contingency.
- **Removed `064-ink-revert-proposer-rotation`** — out of scope for this rollback.
- **Renumbered `063 → 061`** so the contingency sits directly after — and aligned 1:1 with — the forward [`eth/060`](https://github.com/ethereum-optimism/superchain-ops/pull/1490) sequencer & batcher rotation it reverses.

The restored batcher / unsafe-block-signer values are the exact pre-migration Gelato values recorded in `eth/060`'s config as the rollback reference; the modelling overrides mirror `eth/060`'s post-cutover state so a standalone simulation shows a genuine OPE → Gelato diff.

## ⚠️ This is an ARMED break-glass task — status `CANCELLED` on purpose

A rollback can fire at **any** stage of the migration, so the signer nonce cannot be sequenced ahead of time. Pinning a nonce would break CI every time the `eth` stack shifts and would be wrong at activation regardless. Instead:

- **Status `CANCELLED`** → `src/script/fetch-tasks.sh` excludes `EXECUTED`/`CANCELLED`/`APPROVED` from the active `eth` stacked simulation, so CI does not enforce a pinned nonce/hash. This is deliberate — it is an armed contingency, **not** abandoned work. (`DRAFT` / `READY TO SIGN` would keep it in-stack and CI would enforce its hashes — see `TaskManager.sol`.)
- **No nonce override** → at activation the framework reads the safe's **live** on-chain nonce (`StateOverrideManager._getNonceOrOverride` → `IGnosisSafe.nonce()`).
- **Post-migration modelling overrides kept** (owner→FOS, batcherHash/unsafeBlockSigner→OPE) so a pre-activation simulation shows a genuine OPE→Gelato diff; removed at activation once that state is live on-chain.
- **VALIDATION.md** keeps the nonce-independent Domain hash; Message/Safe hashes are generated at activation.

---

## How to execute (activation runbook)

> Run only if the `eth/060` sequencer & batcher rotation must be rolled back. Prereqs: `just --justfile src/justfile install` once; a connected+unlocked Ledger with the Ethereum app (or `SIMULATE_WITHOUT_LEDGER=1` for a dry run). Commands are run from the task directory.

### Step 0 — arm the task
1. Edit `README.md`: change `Status: CANCELLED` → `Status: READY TO SIGN`.
2. If the forward `eth/060` already executed on-chain, edit `config.toml` and **remove the three post-migration modelling overrides** (they only exist so a pre-activation sim shows a diff; the real state is now live): `SystemConfig.owner()` slot `0x33`, `batcherHash` slot `0x67`, and `unsafeBlockSigner` slot `0x65a7ed54…1c08`.
3. Leave the `[stateOverrides]` **nonce absent** — it is read live automatically.

### Execute — single-safe (FoundationOperationsSafe)
```bash
cd src/tasks/eth/061-ink-revert-batcher-unsafe-signer

# 1. Simulate & validate (dry-run: prepend SIMULATE_WITHOUT_LEDGER=1). Reads the live FOS nonce.
just --dotenv-path $(pwd)/.env simulate
#    Copy the printed Domain/Message/Safe hashes into VALIDATION.md and verify on-ledger.

# 2. Each FOS owner signs (Ledger). HD_PATH=<n> to pick a derivation path.
just --dotenv-path $(pwd)/.env sign

# 3. [Facilitator] execute once threshold signatures are collected:
export SIGNATURES="0x<sig1><sig2>..."
just --dotenv-path $(pwd)/.env execute
```

### Post-execution verification
```bash
# batcher + unsafe signer back to Gelato
cast call 0x62C0a111929fA32ceC2F76aDba54C16aFb6E8364 "batcherHash()(bytes32)" --rpc-url mainnet
#   → 0x000000000000000000000000500d7ea63cf2e501dadaa5feec1fc19fe2aa72ac
cast call 0x62C0a111929fA32ceC2F76aDba54C16aFb6E8364 "unsafeBlockSigner()(address)" --rpc-url mainnet
#   → 0x7D056B99AA2021864c42E25B4F8cE3BdEAc9463C
```

See the task's `README.md` / `VALIDATION.md` for full details. Full signing walkthrough: `docs/SINGLE.md`.


src/tasks/eth/062-ink-revert-system-config-owner/README.md
```diff
@@ -0,0 +1,60 @@
+# 062-ink-revert-system-config-owner: ROLLBACK — revert Ink SystemConfig ownership to the Gelato Safe
+
+Status: CANCELLED
```
mds1: CONTINGENCY TASK, SIGN AS NEEDED is a more appropriate status here

src/tasks/eth/062-ink-revert-system-config-owner/.env
mds1, line 1: These files are empty and missing TENDERLY_GAS

src/tasks/eth/064-ink-revert-proposer-rotation/README.md
```diff
@@ -0,0 +1,68 @@
...
+| challenger | 144–164 | `0x9ba6e03d…006b3a` | unchanged |
+| all other fields | — | live | unchanged (read live) |
+
+Plus the ProxyAdminOwner nonce `36`→`37` and the FUS/SC signer nonces (60) increment by 1.
+
+## Simulation & Signing
+
+```bash
+cd src/tasks/eth/064-ink-revert-proposer-rotation
+SIMULATE_WITHOUT_LEDGER=1 just --dotenv-path $(pwd)/.env --justfile ../../../justfile simulate
```
mds1: These commands seem wrong cause this one is a nested safe so you need to specify which safe to sign as, hashes etc may be wrong as a result

sbvegan:
### Scope reduction: rollback is now a single task (`eth/061`)

I've trimmed this contingency PR down to the one rollback we actually intend to arm, and renumbered it to line up with the forward migration in #1490.

**What changed and why:**

- **Removed `062-ink-revert-system-config-owner`.** The rollback plan does not hand SystemConfig ownership back to Gelato. Ownership stays with the FoundationOperationsSafe; if we abort, we only need to point the batcher/sequencer back at Gelato — the FOS keeps control of SystemConfig either way, so reverting ownership is unnecessary (and would be a larger, L1PAO-scale action we don't want armed as break-glass).
- **Removed `064-ink-revert-proposer-rotation`.** Out of scope for this contingency.
- **Renumbered `063 → 061`.** The remaining task (revert batcher + unsafe block signer) is the exact reverse of the forward [`eth/060`](https://github.com/ethereum-optimism/superchain-ops/pull/1490) sequencer & batcher rotation, so it now sits at `061` directly after it. It's aligned 1:1 with `eth/060`: same `SetBatcherAndOrSigner` template, same FOS signer, same SystemConfigProxy target, and it restores the exact pre-migration Gelato values that `eth/060` records as its rollback reference. Also cleaned up the `README`/`config` cross-references to the deleted `062`/`064` and reframed the docs (ownership is not reverted here).

**The trigger:** if we detect a problem after the sequencer & batcher rotation lands, we arm this task (flip `Status: CANCELLED → READY TO SIGN`, drop the modelling overrides) and roll the batcher + unsafe block signer back to Gelato.

⚠️ **Heads-up on numbering:** #1490 currently also adds an `eth/061-ink-proposer-rotation` (forward). If that task stays as-is, merging both PRs would land two `061-*` directories on `main`. Whichever way we resolve it — dropping/renumbering the forward proposer rotation, or renumbering this rollback — the two PRs need to be reconciled before merge so there's a single `061`.

src/tasks/eth/064-ink-revert-proposer-rotation/README.md
Wazabie: Wouldn't bare `just simulate`/`just sign` default to FOS here? As we only need to sign with it and it's a 5 of 7 safe with EOA owners?

PR #1498, opened 2026-07-14
fix: preserve Safe owners in simulation overrides

mds1:
## Summary

Preserve the Safe's real owner links when adding the synthetic simulation owner. The override now appends that owner and increments `ownerCount` instead of replacing the real owner list, which fixes Tenderly simulations that rotate the first owner.

This only changes simulation authentication state. The simulated list has one extra owner, but remains internally consistent; task calldata, Safe transaction hashes, signatures, and onchain state are unchanged. Tasks that depend on the exact owner count or enumerate owners still require a different simulation strategy.

## Validation

- `forge test --match-contract StateOverrideManagerUnitTest --match-test test_getStateOverrides_appendsSimulationOwnerWithoutReplacingOwnerList -vv`

### DIFF
--- a/src/tasks/StateOverrideManager.sol
+++ b/src/tasks/StateOverrideManager.sol
@@ -115,9 +115,9 @@ abstract contract StateOverrideManager is CommonBase {
         defaultOverride.contractAddress = rootSafe;
         defaultOverride = Simulation.addThresholdOverride(defaultOverride.contractAddress, defaultOverride);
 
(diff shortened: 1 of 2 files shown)


Ethnical (approved):

mds1:
Superseded by #1499, which uses an existing Safe owner for Tenderly simulation and leaves the root owner list unchanged.

PR #1499, opened 2026-07-14
fix: simulate Safe tasks from an existing owner

mds1:
## Summary

Use an existing pre-execution Safe owner as the Tenderly simulation sender. This leaves the root Safe owner list and `ownerCount` unchanged, fixing first-owner rotations such as task 057.

Stacked simulations previously used an internal contract caller and made it a synthetic owner. Child Safe simulation overrides are unchanged.

## Validation

- `mise exec -- forge fmt --check`
- `mise exec -- forge build`


Ethnical (approved):
Lgtm!

PR #1502, opened 2026-07-15
feat: re-add OPCM V800 templates with simplified super-permissioned args

### LINKED ISSUE #1466: OPCMUpgradeV800: Fix for the simplified SUPER_PERMISSIONED game
The V800 template does not support the simplified SUPER_PERMISSIONED_GAME type because that game args structure has changed. Since SUPER_PERMISSIONED_GAME hasn't shipped yet, we should just replace the old game args with the simplified one.

The abi encoded game args structure for the simplified super permissioned game can be found here - https://github.com/ethereum-optimism/optimism/blob/423d93e6374a65f3a83c1ed5e29c5998c972a5da/packages/contracts-bedrock/interfaces/L1/opcm/IOPContractsManagerUtils.sol#L39C1-L41C6.

OptimismBot:
Re-adds the **full** OPCM V800 template set that was removed in #1501, now finalized against the released [op-contracts/v8.0.0-rc.2](https://github.com/ethereum-optimism/optimism/releases/tag/op-contracts%2Fv8.0.0-rc.2) OPCM deployed on Sepolia at [`0xda83150fd745d35bb347a7afaf2d7af1d1e8b2b6`](https://sepolia.etherscan.io/address/0xda83150fd745d35bb347a7afaf2d7af1d1e8b2b6) (version `8.0.0`, StandardValidator 2.12.0 at `0xB7a5710a35b49aCFd705312DCba16aCf7591D7f5`, per the superchain-registry `standard-versions-sepolia.toml`).

## What this adds
- `src/template/OPCMUpgradeV800.sol` + `src/template/OPCMMigrateV800.sol`
- `test/integration/SuperRootUpgrade.t.sol` + `test/integration/SuperRootMigrate.t.sol`
- `test/tasks/example/sep/035-opcm-upgrade-v800/*` + `test/tasks/example/sep/036-opcm-migrate-v800/*`
- `src/ci.just` (only the `036` skip remains — see below)

## OPCMUpgradeV800: finalized against the v8.0.0-rc.2 ABI
- **OPCM version check**: requires `8.0.x` (was `7.1.x` against the dev OPCM). `skipOPCMVersionCheck` escape hatch retained for devnets.
- **Exactly six game configs**, in the order the v8 `_assertValidFullConfig` requires: `[CANNON, PERMISSIONED_CANNON, CANNON_KONA, SUPER_PERMISSIONED, SUPER_CANNON_KONA, ZK_DISPUTE_GAME]`. `SUPER_CANNON` (4) is retired — the v8 OPCM rejects a config for it and clears its DGF registration itself.
- **`SUPER_PERMISSIONED` (gt=5)**: proposer-only game args (`abi.encode(proposer)`) and **forced `initBond = 0`** — both enforced by the v8 OPCM.
- **Extra instructions**: only whitelisted keys survive in v8 (`upgrade()` reverts on anything else). `PermittedProxyDeployment`/`DelayedWETH` is dropped (removed from the upgrade path in ethereum-optimism/optimism#22096). The template now also emits **`overrides.cfg.startingAnchorRoot`**: rotating an output-root chain onto super-root games must re-anchor the AnchorStateRegistry to an honest super root (the pre-upgrade anchor is an output root at an L2 block number, meaningless to super-root games). The anchor is a required per-chain config field and `_validate` asserts it landed.
- **`startingRespectedGameType`** restricted to the super game types (5 or 9) — the only values that can satisfy the OPCM's enabled-respected-game check given this template's config builder. Chains with a registered `ZK_DISPUTE_GAME` are rejected up front (dev-feature gated in v8; the template doesn't carry `ZKDisputeGameConfig` inputs).
- **`ProtocolVersions` removed** from the storage-write allowlist — the contract and its OPCM integration were removed in v8.0.0 (ethereum-optimism/optimism#20527).
- **Dev-OPCM `vm.etch` hacks replaced** with a production `_getCodeExceptions()` (V700 pattern): reads `owner()`/`unsafeBlockSigner()`/`batcherHash()` post-upgrade. `batchInbox()` is intentionally absent — SystemConfig 4.0.0 removed the getter and zeroes the legacy slot during reinitialization.

## Example task 035: real OPCM, sepolia-devnet-2
`035-opcm-upgrade-v800` now simulates the production U20 flow: **sepolia-devnet-2** (chainId 420130015, [in the superchain registry](https://github.com/ethereum-optimism/superchain-registry/blob/main/superchain/configs/sepolia-devnet-2/sepolia-devnet-2.toml), on v7: SystemConfig 3.14.2, respected game type 8, `lastUsedOPCM` reports 7.1.17 → the v8 upgrade-sequence gate permits it) upgraded by the deployed rc.2 OPCM at fork block 11470000, rotating to `SUPER_CANNON_KONA` (9). Because the chain is registry-listed, the fixture uses the normal registry discovery flow (no fallback addresses.json), and its missing ETHLockbox also covers the v8 lockbox-feature-disabled upgrade path. Expected validator output is pinned from a live run: `OVERRIDES-L1PAOMULTISIG,OVERRIDES-CHALLENGER,SYSCON-130` (devnet role deviations and its non-standard 60M gas limit).

The previous sdg-v1 devnet fixture chains could not be kept: they are already interop-migrated (shared DGF/ASR/lockbox), and the 2.12.0 validator hard-reverts validating them because the shared DelayedWETH's proxy admin doesn't match the first chain's ProxyAdmin.

`test_upgrade_sepolia()` is restored in `SuperRootUpgrade.t.sol` and passes against the live OPCM; the `035` regression skip is removed from `src/ci.just`.

## OPCMMigrateV800 stays on the dev OPCM
The production rc.2 OPCM ships `migrate()`, but it requires the `OPTIMISM_PORTAL_INTEROP` dev feature, which is **disabled** on the production deployment (`devFeatureBitmap == 0`) — `migrate()` reverts with `OPContractsManagerMigrator_InteropNotEnabled`. The migrate template, the 036 fixture, and its CI skip therefore keep targeting the interop-enabled dev OPCM until a production OPCM enables interop.

Closes #1466 (once merged)

Prompted by: @mds1


mds1:
Pushed a73258b7 to this branch: `OPCMUpgradeV800` (and the duplicated etching loop in `SuperRootMigrate.t.sol`) now read `batchInbox()` through a tolerant staticcall.

ethereum-optimism/optimism#21641 removes `batchInbox()` from the SystemConfig at op-contracts v8.0.0 and clears the legacy slot during reinitialization, so the direct `sysCfg.batchInbox()` call here would revert once a chain is on SystemConfig 4.0.0. The staticcall resolves to `address(0)` when the selector is missing, and still surfaces any revert that carries data rather than silently treating it as the post-removal case.

This carries forward the fix from #1492, which targeted the V800 templates before they were removed from main in #1501. Closing that one as obsolete.

mds1:
Reverted a73258b7 in dcfeea24 — please disregard, my change was wrong.

The etching loop is in `_templateSetup`, which runs before the upgrade executes, so the SystemConfig is still on the pre-V800 version there and `batchInbox()` exists. There is no post-upgrade read of it in the template either. So the tolerant staticcall guarded a state that cannot occur at that point, and its `address(0)` fallback would have masked a genuine failure (wrong registry address, or a chain unexpectedly already upgraded) that the plain call surfaces loudly.

This branch is back to where it was. ethereum-optimism/optimism#21641 needs no superchain-ops change after all.

mds1:
Correction to the SHA in my last comment: the revert landed as ae7ae169, not dcfeea24. My first push raced with the `main` merge (8de8c781) and was rejected, so I re-applied the revert on top of that. Branch tip is ae7ae169 and the templates are back to a plain `sysCfg.batchInbox()`.

Wazabie:
Pushed b31a53d0: finalizes `OPCMUpgradeV800` against the released v8.0.0-rc.2 OPCM (`0xda83…b2b6` on Sepolia) — 6-slot game configs, proposer-only/zero-bond `SUPER_PERMISSIONED`, whitelisted-only extra instructions (drops `PermittedProxyDeployment`, adds a required `overrides.cfg.startingAnchorRoot` re-anchor), `ProtocolVersions` allowlist removal, and `_getCodeExceptions` instead of the dev-OPCM etch hacks. Example task 035 now simulates u19-beta-0 through the real OPCM (skip removed from `src/ci.just`), and `test_upgrade_sepolia()` is restored and passing. The migrate template stays on the dev OPCM — production rc.2 has `OPTIMISM_PORTAL_INTEROP` off, so `migrate()` reverts. PR description updated with full details.

Verified locally against a Sepolia archive fork: `SuperRootUpgrade` + `SuperRootMigrate` suites green, full `simulate-task.sh` run of 035 green.

Wazabie:
Waiting for [Superchain Registry PR to add sepolia-devnet-3](https://github.com/ethereum-optimism/superchain-registry/pull/1311). After merging that the ci will pass.

src/template/OPCMMigrateV800.sol
mds1, line 1: Can we delete this since migrate is out of scope for U20?

src/ci.just
```diff
@@ -53,6 +53,15 @@ simulate-all-templates:
...
             task_path="$root_dir/test/tasks/example/$network/$task_name"
 
+            # OPCMMigrateV800 requires the target chain to already be upgraded to V800/Super Root.
+            # The raw example task simulation forks current chain state and cannot perform that
+            # prerequisite upgrade before invoking migrate. Coverage lives in SuperRootMigrateTest,
+            # which forks Sepolia, upgrades the chain, then runs migrate in the same forked state.
+            if [ "$task_name" = "036-opcm-migrate-v800" ]; then
+                echo "Skipping $task_name (requires pre-upgraded V800/Super Root chain state)"
+                continue
+            fi
```
mds1: We should be able to delete this if we're deleting the migration template

test/tasks/example/sep/036-opcm-migrate-v800/.env
mds1, line 1: Similarly, we could also delete this whole folder.

lib/superchain-registry
```diff
@@ -1 +1 @@
-Subproject commit 848b7c912f02fb97403ea78a1a152ae7e181e6cc
+Subproject commit 3a6192e2d83f845301a17d5e2e9ce1155c1fed3f
```
mds1: Can we bump this to at least `superchain-registry@52037bf`? This task targets rc.3 and says the OPCM comes from `standard-versions-sepolia.toml`, but that entry was added in `52037bf` and is missing from this pin. This pin is a real ancestor of `main`, and the two devnet chain configs match current `main`, but the checked-in source of truth does not contain the rc.3 release data used by the fixture. I think this blocks approval.

src/template/OPCMUpgradeV800.sol
```diff
@@ -0,0 +1,525 @@
...
+import {stdToml} from "forge-std/StdToml.sol";
+import {LibString} from "solady/utils/LibString.sol";
+import {EnumerableSet} from "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
+
+import {OPCMTaskBase} from "src/tasks/types/OPCMTaskBase.sol";
+import {SuperchainAddressRegistry} from "src/SuperchainAddressRegistry.sol";
+import {Action} from "src/libraries/MultisigTypes.sol";
+
+/// @notice A template contract for configuring OPCMTaskBase templates.
+/// Supports: op-contracts/v8.0.0-rc.2
```
mds1: nit, should this say rc.3 now? The checked-in fixture and deployed OPCM target rc.3. The game-type comment on line 46 also still names rc.2.

Wazabie: Right! Missed this

lib/superchain-registry
Wazabie: Bumped in a95345ac to upstream main 9dce5d25 which contains 52037bf

src/template/OPCMUpgradeV800.sol
Wazabie: Fixed, the game-type constants comment now say v8.0.0-rc.3

mds1 (approved):

PR #1505, opened 2026-07-20
eth/062: Ink Mainnet fee-vault recipient update (Gelato → OPE, fee vault recipient)

Wazabie:
## Ink Mainnet Migration — fee-vault fast follow (eth/062)

Adds `eth/062-ink-fee-vault-recipient-update`: transfers Ink Mainnet's (chainId 57073) two cost-covering fee vaults to the OP Enterprise **cost recipient**, as the post-cutover fast follow of the Gelato → OPE migration (plan step 15 / W33; not on the 2026-07-28 cutover critical path).

| Vault | Change |
|---|---|
| `L1FeeVault` (`0x…001a`, v1.6.1) | recipient `0xa6f0F94C…6c93` → `0x1eB630b2e7409597D462dd5f3D21E305FC56B8C9`; minWithdrawalAmount **2 ETH → 0** |
| `OperatorFeeVault` (`0x…001b`, v1.1.1) | recipient `0x…0019` (BaseFeeVault) → same cost recipient; withdrawalNetwork **L2 → L1** (min already 0) |

`SequencerFeeVault` / `BaseFeeVault` are untouched (chain-governor revenue). Both cost vaults end with **no minimum withdrawal**, so cost-recipient sweeps can trigger at any accrued balance (`withdraw()` is permissionless; no automated trigger planned). Exactly **4 portal deposits** (per-field skip-unchanged).

### Recipient provenance
`0x1eB630b2e7409597D462dd5f3D21E305FC56B8C9` is the Ink Mainnet **cost-recipient** L1 EOA (Google KMS key ring `ink-mainnet-0`, key `cost-recipient`), deployed 2026-07-17. Source of truth: `k8s-netchef-prod` → `manifests/ink-mainnet-0/mn-ink-mainnet-0-op-signer/mn-ink-mainnet-0-op-signer.yaml` (auth entry `mn-ink-mainnet-0-cost-recipient`, chainID 1). It funds Ink's L1 operating costs (batcher / proposer / challenger top-ups). Vault split per the ETH chain fee-flow design: L1FeeVault + OperatorFeeVault → Cost Recipient on L1; Base/Sequencer → customer fee receiver (unchanged).

### Status: READY TO SIGN — signing gates
1. **Stacked on #1504** (this PR's base branch): the task uses the new `SetFeeVaultConfig` template — post-Karst the vaults are on the setter design, so this is an in-place config update (no impl deploys).
2. **Governance:** signed by the L1PAO (nested 2-of-2 Foundation Upgrade Safe + Security Council) → requires the Ink chain-servicer-migration Maintenance Upgrade proposal to clear its veto window.
3. **Ordering / nonces:** pins account for both merged migration tasks — `eth/060` is FOS-signed (no effect on these safes); `eth/061` is L1PAO-signed (+1 each). Pins = live 2026-07-20 values (36/62/60) + 1 → L1PAO=37 / FUS=63 / SC=61. Re-simulate + regenerate hashes on any drift.
4. **Key-control proof** for the (never-yet-used) cost recipient before signing.

### Verification performed
- Live state (2026-07-20/21, `rpc-gel.inkonchain.com`): vault versions 1.6.1 / 1.1.1, current recipients/min/network as documented in config comments; L2 ProxyAdmin owner = aliased L1PAO (`0x6B1B…C4E3b`).
- `just simulate council` + `just simulate foundation`: template pre-flight (owner assertion, version gate, setter dry-run) passes; 4 deposits built, selectors/args verified (`setRecipient` `0x3bbed4a0` ×2, `setMinWithdrawalAmount` `0x85b5b14d`, `setWithdrawalNetwork` `0x307f2962`).
- `just simulate-stack eth 062-… council|foundation` (what CI runs): both green; domain/message hashes match VALIDATION.md; root L1PAO safe tx hash identical on both paths.
- Anvil fork replay of the four deposits on Ink Mainnet as the aliased L1PAO: exact expected end state, untouched vaults byte-identical, 40.6k–45.7k gas per call vs the 150k deposit limit, non-owner callers rejected (`ProxyAdminOwnedBase_NotProxyAdminOwner`). Recorded in VALIDATION.md ("Simulating the L2 effect before signing" + "L2 post-execution validation").
- `check-task-statuses.sh` / `check-task-names.sh` pass.


src/tasks/eth/062-ink-fee-vault-recipient-update/README.md
```diff
@@ -0,0 +1,80 @@
+# 062-ink-fee-vault-recipient-update
+
+Status: READY TO SIGN
```
mds1: `READY TO SIGN` does not match the repository status definition or this README's own gates. At Ethereum block 25,584,150 the live L1PAO/FUS/SC nonces remain `36/62/60`, while this task pins the future post-`eth/061` values `37/63/61`. `#1504` is unmerged, CI is red, and the governance/key-control gates are explicitly uncleared.

Can we set this back to `DRAFT, NOT READY TO SIGN` until all four gates clear? Also, before signing, let's make sure the current migration plan and govpost record the split recipients and zero minimum. The first draft attached to `solutions#819` still said Kraken, although the newer Slack decision supersedes it.

src/tasks/eth/062-ink-fee-vault-recipient-update/VALIDATION.md
```diff
@@ -0,0 +1,200 @@
...
+```
+
+## Simulating the L2 effect before signing
+
+Two layers exist here — one automatic, one you can run yourself:
+
+**Automatic (runs inside `just simulate`):** the template's mandatory pre-flight forks Ink via the
+config's `l2RpcUrls`, asserts the L2 ProxyAdmin owner is the aliased L1PAO, version-gates each
+vault, and **dry-runs every setter call as the aliased owner** — so if any of the three L2 writes
```
mds1: nit, both references to `three` here are stale after the `L1FeeVault` minimum became the fourth deposit. These should say `four L2 writes` and `four deposits`.

src/tasks/eth/062-ink-fee-vault-recipient-update/config.toml
```diff
@@ -0,0 +1,82 @@
...
+
+# Per-chain array shape: both vaults withdraw to L1 (0). L1FeeVault is unchanged (no deposit);
+# OperatorFeeVault changes L2 -> L1 so operator fees stop cascading into the BaseFeeVault
+# predeploy and instead withdraw to the L1 cost recipient.
+networks = [0]
+
+# Per-chain array shape: NO minimum withdrawal on either vault. L1FeeVault changes 2 ETH -> 0
+# (emits a deposit); OperatorFeeVault is already 0 (skipped). Cost-recipient sweeps can then be
+# triggered at any accrued balance (withdraw() is permissionless; no automated trigger planned).
+minWithdrawalAmounts = [0]
```
mds1: The Slack thread resolves this in favor of no minimum, so this is not blocking. Worth documenting one implication in the durable plan/govpost: [`withdraw()` is permissionless and passes at zero balance when the minimum is zero](https://github.com/ethereum-optimism/optimism/blob/0f21af947798a94aa50e797c61f576ca18b1334d/packages/contracts-bedrock/src/L2/FeeVault.sol#L145-L170), so a caller can generate zero-value L2-to-L1 withdrawal messages at their own gas cost. Who owns triggering withdrawals if no keeper is planned, and are we explicitly accepting this spam/cadence tradeoff?

mds1 (commented):
Reviewed exact head `f9698231` against current live state and the related Slack thread. Both `council` and `foundation` simulations pass with the pinned dependencies and reproduce the four-deposit calldata plus every hash in `VALIDATION.md`. The recipient matches the Ink KMS config, and the zero-minimum change matches the Slack decision.

I found one sign-readiness blocker and two documentation/operational points inline. The payload is close, but I would keep the task draft until its gates clear and rerun CI to green. The current `stacked_simulation_eth` failure is an RPC `block is out of range` error in `createSelectFork`; both exact-head standalone paths pass.

src/tasks/eth/062-ink-fee-vault-recipient-update/config.toml
Wazabie: I think we are accepting the spam/cadence tradeoff. Looping in @sbvegan here and @zhwrd, do we want to set a low minimum instead?

sbvegan: I don't hold a strong opinion here tbh, we found that 2 ether is about a year of operating costs, so maybe it's better to be set to 0.15 ether to be able to withdrawal on a monthly cadence. If that's something we want to do

src/tasks/eth/062-ink-fee-vault-recipient-update/README.md
Wazabie: Reverted to `DRAFT, NOT READY TO SIGN` in 1aa7ddec; the four signing gates (#1504 merge, governance window, eth/061 nonce ordering, key-control proof) stay explicit in the README. The cost-vs-governor vault split and zero minimum are now documented in the README Objective section (e373a861); migration plan / govpost will be updated to match before signing.

src/tasks/eth/062-ink-fee-vault-recipient-update/VALIDATION.md
Wazabie: Fixed in e373a861 — both references now say four L2 writes / four deposits.

src/tasks/eth/062-ink-fee-vault-recipient-update/config.toml
Ethnical: Same here, I don't think there is any impact, but having something slightly bigger than 0 should be considered, in my opinion.

Wazabie: Ok, I have updated the task to have 0.15 ETH as min withdrawal

Ethnical (approved):
LGTM!

PR #1519, opened 2026-08-11
template: multichain support for TransferOwners and TransferL2PAOFromL1

Wazabie:
`TransferOwners` and `TransferL2PAOFromL1` previously required one chain per task. This PR adds multi-chain support (all transferred to the same new owner) in one task/transaction.

Use case: the Metal/Mode/Zora/Dust handover to Conduit ([#1521](https://github.com/ethereum-optimism/superchain-ops/pull/1521)) can be done in two bundled tasks (L1 and L2 transfers) per network instead of 16 single-chain ones.

Examples and tests:

* eth/047 + eth/048 (mainnet) and sep/042 + sep/043 (sepolia); Dust is not in the superchain-registry, so its addresses come from a fallback `addresses.json`
* The L2PAO regression tests stack the tasks in production order: the L1 ownership task executes first and the L2PAO task is built on the resulting state. The fixtures pin `ProxyAdminOwner` in `[addresses]`, same as the production tasks: after the L1 task, the registry resolves the owner to the new Safe, but the deposits must still be sent by the previous L1PAO.
* 048/043 keep their `stateOverrides` only so `simulate-all-templates` can simulate them standalone; in the stacked tests, and in production, the L1 task has already executed and the overrides are a no-op.


test/tasks/example/eth/048-transfer-l2pao-multichain/config.toml
```diff
@@ -0,0 +1,35 @@
...
+fallbackAddressesJsonPath = "test/tasks/example/eth/048-transfer-l2pao-multichain/addresses.json"
+templateName = "TransferL2PAOFromL1"
+
+# The new owner address to alias: the operator's designated mainnet Safe.
+newOwnerToAlias = "0x4a4962275DF8C60a80d3a25faEc5AA7De116A746"
+
+[stateOverrides]
+# Set the current L1PAO of each chain to the new owner. The TransferL2PAOFromL1 template
+# assumes all L1 ownership transfers have already been completed.
+# Mode ProxyAdmin
```
mds1: Shouldn't the overrides not be needed since the tasks are stacked? You'll probably have to change the regression test so that it stacks the two tasks, but that seems preferable because that's exactly what we're going to do in production and makes the test better reflect reality

Wazabie:
Overrides removed. The fix declares `DEPENDS_ON` in `.env` and ci (simulate-all-templates and the regression tests) simulates 48 stacked on 047 (and 43 stacked on 42). `FETCH_TASKS_ONLY` is used as a new filter in `fetch-tasks.sh`

Ethnical (approved):
LGTM!

PR #1520, opened 2026-08-13
eth/066: Soneium Mainnet fee-vault recipient update (all four L2 vaults)

Wazabie:
Rotates the recipient of all four Soneium Mainnet fee vaults to the new Soneium fee recipient Safe `0x34ffF1A1CB3C054E9eD1BbD36883B14A66E6C260`, and lowers the minimum withdrawal amount from 10 ETH to 5 ETH on the Sequencer, Base and L1 fee vaults (operator approved; OperatorFeeVault stays at 0).

Same `SetFeeVaultConfig` flow as eth/062 on Ink: 7 deposits from the L1PAO through the portal (4x `setRecipient`, 3x `setMinWithdrawalAmount`). Withdrawal networks unchanged.

One template change: `minWithdrawalAmounts` is now read with forge's typed parser, so values above TOML's int64 limit (like 10 ETH) can be written as strings. No behavior change for existing tasks.


fiexer:
As representative of Sonieum Foundation, I confirm that the address `0x34ffF1A1CB3C054E9eD1BbD36883B14A66E6C260` is the correct address as new fee-vault recipient.

src/tasks/eth/066-soneium-fee-vault-recipient-update/.env
mds1, line 1: Do we really still need to add these one line .env files every time? I would love if we can fix that and stop needing these. At a minimum this value should be in config files instead

src/tasks/eth/066-soneium-fee-vault-recipient-update/config.toml
```diff
@@ -0,0 +1,83 @@
+# Soneium Mainnet (chainId 1868) — fee-vault recipient rotation + minimum-withdrawal reduction.
+#
+# Rotates the recipient of ALL FOUR L2 fee-vault predeploys to the new Soneium fee
+# recipient Safe, and lowers the minimum withdrawal amount 10 ETH -> 5 ETH on the three
+# vaults that carry it (value agreed with the chain operator; 5 ETH also fits TOML's
```
mds1: How can I verify the addresses and amounts were confirmed with the chain operator?

src/tasks/eth/066-soneium-fee-vault-recipient-update/config.toml
```diff
@@ -0,0 +1,83 @@
...
+# int64 integer range — 10 ETH does not — keeping this a task-only change with the
+# unmodified SetFeeVaultConfig template). Withdrawal networks are pinned to their live
```
mds1: It feels more robust to change the template to expect/read/validate a string for more flexibility if this is actually a problem. My sense is we are changing things to 5 ETH and just got lucky that this fits into the TOML constraint, and that this code comment doesn't really matter and is an LLM artifact?

src/tasks/eth/066-soneium-fee-vault-recipient-update/config.toml
```diff
@@ -0,0 +1,83 @@
...
+# values (all L2) and therefore produce no deposits:
```
mds1: What does "Withdrawal networks are pinned to their live values and (all L2) therefore produce no deposits"?

src/tasks/eth/066-soneium-fee-vault-recipient-update/config.toml
```diff
@@ -0,0 +1,83 @@
...
+#   - SequencerFeeVault (0x…0011, v1.6.1): recipient 0xF07b…1112 -> 0x34fF…C260; min 10 -> 5 ETH
+#   - BaseFeeVault      (0x…0019, v1.6.1): recipient 0xF07b…1112 -> 0x34fF…C260; min 10 -> 5 ETH
+#   - L1FeeVault        (0x…001a, v1.6.1): recipient 0xF07b…1112 -> 0x34fF…C260; min 10 -> 5 ETH
+#   - OperatorFeeVault  (0x…001b, v1.1.1): recipient 0x4200…0019 (BaseFeeVault cascade)
+#                                          -> 0x34fF…C260 (paid directly); min stays 0
```
mds1: I also don't know what these comments mean, like "BaseFeeVault cascade" and "paid directly"

src/tasks/eth/066-soneium-fee-vault-recipient-update/config.toml
```diff
@@ -0,0 +1,83 @@
...
+  "0x4200000000000000000000000000000000000011",  # SequencerFeeVault (v1.6.1)
+  "0x4200000000000000000000000000000000000019",  # BaseFeeVault      (v1.6.1)
+  "0x420000000000000000000000000000000000001a",  # L1FeeVault        (v1.6.1)
+  "0x420000000000000000000000000000000000001b",  # OperatorFeeVault  (v1.1.1)
+]
+
+# New Soneium fee recipient — a 3-of-5 Safe deployed at the SAME address on both
+# Ethereum L1 and Soneium L2 (verified live 2026-08-13: canonical Safe v1.4.1
+# singletons — SafeL2 0x29fcB43b…C762 on Soneium, Safe 0x41675C09…461a on L1 —
+# with identical owner sets). Withdrawals stay on L2 (network unchanged), so the
```
mds1: Any important verification information should be in the readme, which means we can delete code comments like this

src/tasks/eth/066-soneium-fee-vault-recipient-update/config.toml
```diff
@@ -0,0 +1,83 @@
...
+
+# REQUIRED by the template: L2 RPC for the pre-flight (L2 ProxyAdmin-owner assertion,
+# per-vault version gate, live-value caching, and a dry-run of every setter call).
+l2RpcUrls = ["https://rpc.soneium.org"]
+
+[addresses]
+# Pinned for robustness (also resolvable from the superchain-registry:
+# superchain/configs/mainnet/soneium.toml).
+ProxyAdminOwner     = "0x5a0Aae59D09fccBdDb6C6CcEB07B7279367C3d2A" # L1PAO mainnet (nested 2-of-2: Foundation Upgrade Safe + Security Council)
+OptimismPortalProxy = "0x88e529A6ccd302c948689Cd5156C83D4614FAE92"
```
mds1: Isn't pinning less robust because then we aren't resolving them on-chain at runtime? The motivation for contracts like the SuperchainAddressRegistry was to explicitly avoid having to hardcode addresses, because it means the reviewer has to verify these addresses, which is more work for the reviewers

src/tasks/eth/066-soneium-fee-vault-recipient-update/config.toml
```diff
@@ -0,0 +1,83 @@
...
+
+[stateOverrides]
+# Nested L1PAO signer-safe nonces (slot 5), pinned to the values EXPECTED AT SIGNING TIME.
+# Pins = live on-chain values (read 2026-08-13, re-verified 2026-08-14: L1PAO=40 / FUS=66 /
```
mds1: Another comment that seems odd, today is Aug 17 so am I supposed to re-verify as I review these?

src/tasks/eth/066-soneium-fee-vault-recipient-update/README.md
```diff
@@ -0,0 +1,85 @@
...
+| `BaseFeeVault` `0x4200000000000000000000000000000000000019` | v1.6.1 | recipient `0xF07b3169ffF67A8AECdBb18d9761AEeE34591112` → `0x34ffF1A1CB3C054E9eD1BbD36883B14A66E6C260`; minWithdrawalAmount **10 ETH → 5 ETH** |
+| `L1FeeVault` `0x420000000000000000000000000000000000001A` | v1.6.1 | recipient `0xF07b3169ffF67A8AECdBb18d9761AEeE34591112` → `0x34ffF1A1CB3C054E9eD1BbD36883B14A66E6C260`; minWithdrawalAmount **10 ETH → 5 ETH** |
+| `OperatorFeeVault` `0x420000000000000000000000000000000000001b` | v1.1.1 | recipient `0x4200000000000000000000000000000000000019` (BaseFeeVault) → `0x34ffF1A1CB3C054E9eD1BbD36883B14A66E6C260` (min stays 0, network stays L2) |
+
+**Why 5 ETH:** the value is agreed with the chain operator. It keeps `withdraw()` (permissionless, no automated trigger planned) spam-resistant while making sweeps to the new recipient roughly twice as frequent — and it fits TOML's int64 integer range (max ~9.22e18 wei ≈ 9.22 ETH), which the previous 10 ETH value did not, keeping this a task-only change against the unmodified `SetFeeVaultConfig` template.
```
mds1: What does the amount have to do with spam resistance? And did we actually make this change because of TOML?

src/tasks/eth/066-soneium-fee-vault-recipient-update/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Mechanism
+
+All four Soneium fee vaults are on the post-Karst (U19, [eth/053](../053-U19-op-ink-mmz-soneium/README.md)) mutable / setter design (Seq/Base/L1 v1.6.1, Operator v1.1.1 — verified live 2026-08-13). Config lives in proxy storage behind owner-gated setters, so this task is an **in-place config update** — no implementation deployments, no proxy upgrades. It is the same `SetFeeVaultConfig` mechanism proven on Ink Mainnet by [eth/062](../062-ink-fee-vault-recipient-update/README.md).
+
+`SetFeeVaultConfig` sends each changed field as an `OptimismPortal2.depositTransaction` from the L1PAO; the deposit's aliased sender (`0x6B1BAE59D09fCcbdDB6C6cceb07B7279367C4E3b`, the alias of the L1PAO — verified live as the Soneium L2 ProxyAdmin owner) is exactly the owner the setters authorize against. The template's mandatory pre-flight forks Soneium via `l2RpcUrls` to assert L2 ProxyAdmin ownership, enforce the per-vault version gate (≥ 1.6.0 / ≥ 1.1.0), and dry-run every setter before any signature is collected.
+
+Per-field skip-unchanged yields exactly **7 deposits** (each with a 150,000 L2 gas limit), in `vaultProxies` order with recipient before minimum per vault:
```
mds1: I don't know what "Per-field skip-unchanged" means. Can we take a pass at all the text and code comments and make things more concise and clear?

src/tasks/eth/066-soneium-fee-vault-recipient-update/VALIDATION.md
```diff
@@ -0,0 +1,232 @@
...
+## Simulating the L2 effect before signing
+
+Two layers exist here — one automatic, one you can run yourself:
+
+**Automatic (runs inside `just simulate`):** the template's mandatory pre-flight forks Soneium via
+the config's `l2RpcUrls`, asserts the L2 ProxyAdmin owner is the aliased L1PAO, version-gates each
+vault, and **dry-runs every setter call as the aliased owner** — so if any of the seven L2 writes
+could revert, the simulation itself fails loudly before any signature is collected.
+
+**Manual replay (recommended independent check):** reproduce exactly what the seven deposits will
```
mds1: Recommended, or required?

src/tasks/eth/066-soneium-fee-vault-recipient-update/VALIDATION.md
```diff
@@ -0,0 +1,232 @@
...
+         0x4200000000000000000000000000000000000019 \
+         0x420000000000000000000000000000000000001a; do
+  cast send $v "setRecipient(address)" $NEW --from $ALIASED --unlocked --rpc-url $L2RPC --gas-limit 150000
+  cast send $v "setMinWithdrawalAmount(uint256)" $MIN --from $ALIASED --unlocked --rpc-url $L2RPC --gas-limit 150000
+done
+cast send 0x420000000000000000000000000000000000001b "setRecipient(address)" $NEW \
+  --from $ALIASED --unlocked --rpc-url $L2RPC --gas-limit 150000
+
+# 4. Run the same read-backs as the post-execution section below against $L2RPC and
+#    confirm the changed/unchanged expectations, then kill anvil.
```
mds1: I thought for deposit transaction tasks, we give them steps on how to open that transaction simulation in Tenderly. Do we not do that anymore? This lets you see the full state diff, whereas the current approach would not expose a malicious state diff to you because you are only using getters.

src/tasks/eth/066-soneium-fee-vault-recipient-update/VALIDATION.md
```diff
@@ -0,0 +1,232 @@
...
+  - **Decoded Kind:** `struct ResourceMetering.ResourceParams`
+  - **Before:** ``
+  - **After:** ``
+  - **Summary:** params
+  - **Detail:** `ResourceMetering` bookkeeping (`prevBoughtGas` / `prevBlockNum`) updated by the
+    Soneium Mainnet `OptimismPortal` as a side effect of the seven `depositTransaction` calls.
+    The exact packed value depends on the block the transaction lands in; only this slot of the
+    portal should change.
+
+## L2 post-execution validation
```
mds1: Related to the above comment, how come we don't list the L2 state diff anymore? I think we should

src/tasks/eth/066-soneium-fee-vault-recipient-update/config.toml
Wazabie: I actually simulated the task keeping 10eth as minWithdrawal and it exceeded the toml int range, hence the comment. 

It was agreed with soneium to lower the minimum amount not only to circumvent this, but also because it's a high minimum that would make them withdraw less often. 

Regardless of the above, I agree it would be more robust to fix this so I'm adding a change to address it.

src/tasks/eth/066-soneium-fee-vault-recipient-update/.env
Wazabie: Right, I think we can actually remove this as we have the 15M cap in MultisigTask.sol

src/tasks/eth/066-soneium-fee-vault-recipient-update/config.toml
Wazabie: These were agreed with the chain operator in our shared channel with them. Addresses were also confirmed with a comment on this PR.

src/tasks/eth/066-soneium-fee-vault-recipient-update/README.md
Wazabie: We did not make this change uniquely because of the TOML, it was also agreed so that withdrawals could be done more frequently. The spam-resistance refers to the fact that having a 0 minimum would allow anyone to spam withdrawal calls.

src/tasks/eth/066-soneium-fee-vault-recipient-update/README.md
Wazabie: It refer to the fact that we wouldn't see a deposit for no ops (e.g. operatorFee minimum withdrawal remains 0, hence the missing of what would have been the 8th deposit tx - 4 addresses update + 3 minWithdrawal updates).

I'm currently streamlining all comment to make everything clearer.

src/tasks/eth/066-soneium-fee-vault-recipient-update/.env
Wazabie: Actually the env file is used to set the gas limit in the tenderly simulation. Removing it makes tenderly set the default 8M in the simulation.

src/tasks/eth/066-soneium-fee-vault-recipient-update/config.toml
Wazabie: Agreed, removing these so they fetch from SCR

src/tasks/eth/066-soneium-fee-vault-recipient-update/VALIDATION.md
Wazabie: Added

src/tasks/eth/066-soneium-fee-vault-recipient-update/VALIDATION.md
Wazabie: Added full l2 simulation steps for Tenderly

src/tasks/eth/066-soneium-fee-vault-recipient-update/config.toml
Wazabie: Removed

Wazabie:
> As representative of Sonieum Foundation, I confirm that the address `0x34ffF1A1CB3C054E9eD1BbD36883B14A66E6C260` is the correct address as new fee-vault recipient.

@fiexer for tracking, could you also please confirm that the minWithdrawal should be set to 5 ETH for the L1, the Base and Sequencer Fee Vaults?

src/template/SetFeeVaultConfig.sol
```diff
@@ -137,7 +137,11 @@ contract SetFeeVaultConfig is L2TaskBase {
 
         recipients = abi.decode(toml.parseRaw(".recipients"), (address[]));
         networks = abi.decode(toml.parseRaw(".networks"), (uint256[]));
-        minWithdrawalAmounts = abi.decode(toml.parseRaw(".minWithdrawalAmounts"), (uint256[]));
+        // Typed (coercing) reader instead of parseRaw: TOML integers are int64-bounded, so
+        // minimums above ~9.22 ETH (e.g. a live 10 ETH value) cannot be written as bare
+        // integers — encode them as decimal strings ("10000000000000000000"). The typed
+        // reader accepts both bare integers and string-encoded values.
+        minWithdrawalAmounts = toml.readUintArray(".minWithdrawalAmounts");
```
mds1: We should have tests verifying this claim. And what happens if you enter a bare integer larger than `int64`?

src/tasks/eth/066-soneium-fee-vault-recipient-update/config.toml
```diff
@@ -0,0 +1,45 @@
...
+]
+
+# New Soneium fee recipient
+recipients = ["0x34ffF1A1CB3C054E9eD1BbD36883B14A66E6C260"]
+
+# Unchanged, all four vaults already withdraw on L2 (Network 1)
+networks = [1]
+
+# Flat per-(chain, vault) shape matching vaultProxies order: Seq / Base / L1FeeVault
+# lower from 10 ETH to 5 ETH (operator-approved; keeps values int64-representable),
```
mds1: "keeps values int64-representable" — outdated comment?

src/tasks/eth/066-soneium-fee-vault-recipient-update/VALIDATION.md
```diff
@@ -0,0 +1,304 @@
...
+The output prints domain and message hashes (to check against the expected ones at top of this file) and a
+Tenderly simulation link. Open the link, paste the [task calldata](#task-calldata) into the
+**Raw input data** field and simulate: the state diff must match
+[L1 State Changes](#l1-state-changes) and the **Events** must show seven
+`TransactionDeposited` events emitted by the portal.
+
+### L2 Simulation
+
+**Manual replay (required independent check):** reproduce exactly what the seven deposits will
+do on Soneium and inspect the resulting state, on a local fork:
```
mds1: Can't we get rid of this and just ask signers to only do the tenderly step?

Wazabie: We can for this task. Other tasks for chains that have no Tenderly integration will have to do the anvil fork check instead.

src/template/SetFeeVaultConfig.sol
Wazabie: Added test/template/SetFeeVaultConfigToml.t.sol. The read is extracted into `_parseMinWithdrawalAmounts` so the tests pin the template's actual code path: bare integers, string-encoded values (10 ETH and uint256 max) and a mix of both in one array all parse.

A bare integer above int64.max is a TOML parse error: `vm.parseTomlUintArray` reverts at task setup with "number too large to fit in target type" so it fails loudly and can never be silently truncated. Also covered by the tests.

src/tasks/eth/066-soneium-fee-vault-recipient-update/config.toml
Wazabie: Yes, leftover from before the template switched to the typed reader. Removed.

src/tasks/eth/066-soneium-fee-vault-recipient-update/VALIDATION.md
Wazabie: Done, removed the anvil replay. The L2 Simulation section is now just the Tenderly Virtual TestNet steps.

mds1 (approved):

PR #1555, opened 2026-09-21
eth/073: Ink Mainnet SystemConfig gas limit 60M to 30M

Wazabie:
Adds `src/tasks/eth/073-gas-limit-ink`: 
- sets the Ink Mainnet (chainId 57073) `SystemConfig` gas limit from 60,000,000 back to 30,000,000 with a `SystemConfig.setGasLimit(uint64)` call on the `SystemConfigProxy` `0x62C0a111929fA32ceC2F76aDba54C16aFb6E8364` 
- executed directly by its owner, the Foundation Upgrade Safe `0x847B5c174615B1B7fDF770882256e2D3E95b9D92` (owner since eth/065). No other `SystemConfig` parameter changes.

The `SystemConfigGasLimit` template, has already been used for eth/055 (OP Mainnet) and sep/103 (Ink Sepolia).

This PR is stacked on #1556


src/tasks/eth/073-gas-limit-ink/VALIDATION.md
```diff
@@ -0,0 +1,152 @@
...
+
+While eth/071 and eth/072 are pending, the stacked simulation executes them first; the hashes
+and Tenderly link for this task are the last ones printed.
+
+Check:
+
+1. The domain and message hashes printed to the terminal match the ones at the top of this
+   file.
+2. In the Tenderly link printed by the simulation: paste the
+   [task calldata](#task-calldata) into the **Raw input data** field and simulate; the
```
mds1: Let's point signers to the outer `execTransaction` calldata printed beneath the Tenderly link here. The draft targets the Safe, but the linked task calldata starts with `0x174dea71` (`aggregate3Value`) and is only the inner payload.

Reproduced on a local fork with the draft's overrides: pasting the inner payload reverts and leaves the gas limit at 60M. Using the generated `0x6a761202` execution payload succeeds, sets 30M, and emits exactly one gas-limit `ConfigUpdate`. The inner calldata can stay below for decoding and verification.

src/tasks/eth/073-gas-limit-ink/config.toml
```diff
@@ -0,0 +1,28 @@
+templateName = "SystemConfigGasLimit"
+
+# Sets the Ink Mainnet SystemConfig gas limit from 60,000,000 back to 30,000,000 (the genesis value
+# in superchain-registry superchain/configs/mainnet/ink.toml) with one SystemConfig.setGasLimit call.
+# Single-safe task executed by the SystemConfig owner, the Foundation Upgrade Safe (owner since
+# eth/065). Sequenced after eth/071-U20-op-ink-soneium-uni and
+# eth/072-soneium-fee-vault-recipient-update, each of which consumes one FUS nonce as an L1PAO
+# child.
```
mds1: nit, would recommend not including comments like this in the config, and keep:
- Info about the task itself to the README only 
- Info about sequencing: out altogether, and just leave it implicitly in the task folder

This just makes maintenance simpler and ensures a single source of truth

mds1 (approved):

